PR #108, opened 2026-06-10
Refactor: Extract rail tiles and space actions into separate files

krazyjakee:
## Summary
Refactored the `accord_home_rail.dart` file by extracting large helper classes and functions into dedicated, focused modules. This improves code organization, maintainability, and follows the single-responsibility principle without changing any functionality.

## Changes
- **`accord_home_rail_tiles.dart` (new):** Extracted `_DraggableSpace`, `_DraggableSpaceState`, `_FolderTile`, and `_SpaceIcon` widget classes that handle the visual representation and interaction of space tiles and folders in the rail.

- **`accord_home_space_actions.dart` (new):** Extracted space action functions (`_serverActionEntries`, `_copyServerLink`, `_leaveSpace`, `_leaveAndDeleteSpace`) that handle muting, inviting, settings, and leaving spaces. These are shared between the standalone-space menu and folder-member menu.

- **`accord_home_reactions.dart` (new):** Extracted reaction-related widgets (`_ReactButton`, `_ReactionPill`, `_ReactorsDialog`) from `accord_home_message_row.dart` for cleaner separation of concerns.

- **`accord_home_mute_button.dart` (new):** Extracted the `_MuteButton` widget and `_NotifAction` enum from `accord_home_message_row.dart` to isolate channel notification management.

- **`accord_home.dart`:** Updated to include new `part` directives for the extracted files, maintaining the part-file pattern used throughout the codebase.

- **`accord_home_rail.dart`:** Removed ~750 lines of extracted code; file now focuses on the main `_SpaceRail` widget and rail layout logic.

- **`accord_home_message_row.dart`:** Removed ~400 lines of extracted reaction and mute button code; file now focuses on message rendering and actions.

## Implementation Details
- All extracted code uses `part of 'accord_home.dart'` to maintain access to the same import scope and private members.
- No logic changes; this is a pure refactoring for improved code organization.
- Follows the existing Bonfire/Daccord codebase pattern of using part files for feature-specific UI components.

https://claude.ai/code/session_01G5zyWf2nQTv7wJpVkhEvAx

PR #111, opened 2026-06-11
feat(messaging): emoji reactions with custom-emoji image resolution

krazyjakee:
## Summary
- Adds emoji reactions to message rows: reaction pills, a reaction picker button, and a "who reacted" dialog (`_ReactButton`, `_ReactionPill`, `_ReactorsDialog`).
- Resolves custom-emoji reaction images from the space's emoji catalog by id (using the authoritative `image_url`), falling back to the bare CDN path, so animated/custom emoji render correctly. Unicode reactions render without an image.
- Emoji management now prefers `emoji.imageUrl` and uses the `gif`/`png` format based on `emoji.animated`.
- Fixes channel-mute detection in the channel context menu to read `channel_id` from the structured mutes list rather than stringifying whole entries.

## Why
Reaction rendering previously used a bare CDN emoji URL that didn't account for custom/animated emoji or the space path segment, so custom reactions wouldn't display. This wires reactions through the space emoji catalog for correct images.

## Notes for reviewers
- `flutter analyze --no-fatal-infos` is clean on all four changed files.
- `_MuteButton` is defined in `accord_home_message_row.dart` but appears unused — flagging in case it should be wired up or removed in a follow-up.

## Test plan
- [ ] React to a message with a unicode emoji — pill shows the glyph and count.
- [ ] React with a custom/animated space emoji — pill shows the correct image.
- [ ] Long-press / right-click a reaction pill — reactors dialog lists who reacted.
- [ ] Mute a channel, reopen its context menu — mute state reflects correctly.


krazyjakee:
Pushed a follow-up commit (`0f75d518`) that makes this branch compile and adds two fixes:

**Build fix — duplicate definitions.** The reaction/mute widgets were defined both in `accord_home_message_row.dart` and in the dedicated part files (`accord_home_mute_button.dart` / `accord_home_reactions.dart`) that refactor #108 had already extracted, causing `duplicate_definition` errors (`_MuteButton`, `_MuteButtonState`, `_NotifAction`, `_ReactButton`, `_ReactionPill`, `_ReactorsDialog`). Removed the stale copies from `accord_home_message_row.dart` and ported the fixes into the canonical part files.

**Bug fixes ported into the canonical files:**
- *Mute state never reflected:* `listMutes` returns objects `{user_id, channel_id, created_at}`, but the client stringified each whole map (`e.toString()`) instead of reading `channel_id`, so the muted check was always false. Now parses `channel_id`.
- *Custom-emoji reactions/management showed broken images:* the client built `<cdn>/emojis/<id>.png`, but the server serves `<cdn>/emojis/<space_id>/<id>.<ext>`. Now uses the server-provided `image_url` (resolved via the space emoji catalog for reactions), matching the emoji picker / message-markup pattern.

`flutter analyze --no-fatal-infos` is clean and all 186 tests pass.

krazyjakee:
## Review findings and fixes (commit `298bd3b`)

Findings ranked by severity:

---

### 🔴 CRITICAL — Duplicate class definitions (compile error)

`_ReactButton`, `_ReactionPill`, and `_ReactorsDialog` were added to `accord_home_message_row.dart` by this PR, but identical/older definitions of all three already existed in `accord_home_reactions.dart`. Both files are `part of 'accord_home.dart'`, so the same class name in two `part` files is a **compile-time duplicate-definition error** — the app would not build.

**Fix:** Moved all reaction/notification widgets to `accord_home_reactions.dart` (their natural home) and removed the duplicate definitions from `accord_home_message_row.dart`. The improved `_ReactionPill` with `imageUrl` (from this PR) replaced the old `cdnUrl`-based version in the reactions file.

---

### 🟠 HIGH — Wrong docstring on `_MuteButton`

The doc-comment on `_MuteButton` began with:
```dart
/// Opens the full emoji picker to add a reaction to the message.
/// A bell toggle in the channel header …
```
The first line describes `_ReactButton`, not `_MuteButton` — a stale line accidentally left from a copy-paste. **Fixed** by removing it.

---

### 🟡 MEDIUM — `accord_home_message_row.dart` over 1000 lines

After the PR the file reached **1234 lines**. Resolved as a side-effect of fix #1 (moving the widgets to the reactions file brings it back to **837 lines**).

---

### 🟡 MEDIUM — Double `accordAvatarUrl()` call per reactor

In `_ReactorsDialogState.build()`, `accordAvatarUrl(user, cdnUrl)` was called twice per user — once for the null check and once to get the actual value. **Fixed** by extracting a `_ReactorTile` widget that stores the URL in a local variable.

---

### 🟢 LOW — Redundant `emojiUrl` alias in `_ReactionPill`

```dart
final emojiUrl = imageUrl;  // pointless rename
```
`emojiUrl` was used in exactly the same way `imageUrl` would be. **Fixed** by removing the alias and using `imageUrl!` directly.

---

### 🟢 LOW — Line too long in `_reactionEmojiUrl`

```dart
final id = rawId ?? parseEmojiToken(reaction.emoji['name']?.toString() ?? '').id;
```
Wrapped to two lines to stay within 80 columns.

---

### 🔵 TEST COVERAGE — New test file added

`test/features/messaging/emoji_utils_test.dart` — covers:
- `parseEmojiToken`: name/id splitting, colon-wrapped shortcodes, non-numeric suffixes, empty input
- `resolveEmojiGlyph`: known shortcode → glyph, passthrough unicode, colon-stripped lookup, unknown fallback
- Mute-list `channel_id` extraction logic (mirrors the parsing in `channel_context_menu.dart` and `_MuteButtonState._load`)


---

PR #115, opened 2026-06-11
feat: MCP permission tools + rail reorder, image cropping, faster member loading

krazyjakee:
## Summary

Bundles the current in-flight working-tree changes across the client and accordkit.

### MCP / developer tooling
- **Permission management tools** in the in-app MCP client (`mcp_tools.dart`):
  - Read: `list_roles`, `list_permissions` (local enum of valid permission ids + descriptions)
  - Manage (privileged, off by default like `moderate`): `create_role`, `update_role`, `delete_role`, `add_member_role`, `remove_member_role`, and per-channel overwrites `list_channel_permissions` / `set_channel_permission` / `delete_channel_permission`
  - Surfaces the new `manage` permission group in settings + the developer settings page
  - Includes follow-up review fixes (reject empty-name `update_role`) and `mcp_tools` tests

### Member roster loading
- `members.list` gains a `withUser` flag — the server embeds each member's `user` object in one batched query, so the controller skips the per-member fetch. Older servers ignore the flag and fall back to a **bounded-concurrency (8)** user backfill instead of one request per member.
- Message rendering only loads the roster when the message actually contains a mention.

### Space rail
- Drag-reorder with insertion gaps between items; reorder spaces within and across folders. `moveSpaceToFolder` gains a `before` anchor for precise placement.

### Image cropping
- New shared `showImageCropDialog` (`crop_your_image`) with a pan/zoom frame, wired into avatar (circular 1:1), space icon, and banner (16:9) editing.

### Admin
- Open a space directly in the rail via `/spaces?space=<id>` using the admin's own connection, rather than a public member-join.

### Misc
- Error reporting: navigation breadcrumbs driven off persisted space/channel selection (avoids a settings→reporting dependency cycle).
- Open-tab strip: keep a tab's captured label in sync on channel rename.

## Test plan
- `flutter analyze --no-fatal-infos` clean on touched files
- `dart test` in `packages/accordkit` (members `withUser` endpoint test added)
- `mcp_tools` tests added for the new permission tools

### DIFF
--- a/lib/features/admin/views/admin_spaces_tab.dart
+++ b/lib/features/admin/views/admin_spaces_tab.dart
@@ -1,9 +1,11 @@
 import 'package:accordkit/accordkit.dart';
 import 'package:bonfire/features/authentication/models/accord_auth.dart';
 import 'package:bonfire/features/authentication/repositories/accord_auth.dart';
+import 'package:bonfire/features/spaces/controllers/spaces.dart';
 import 'package:bonfire/theme/theme.dart';
 import 'package:flutter/material.dart';
 import 'package:flutter_riverpod/flutter_riverpod.dart';
+import 'package:go_router/go_router.dart';
 
 /// Instance-admin "Spaces" tab: lists every space on the instance, with search,
 /// create, delete and owner-transfer. Mirrors the reference client's
@@ -61,6 +63,20 @@ class _AdminSpacesTabState extends ConsumerState<AdminSpacesTab> {
       _busy = false;
       _spaces = data is List ? data.cast<AccordSpace>() : <AccordSpace>[];
     });
+  }
+
+  /// Opens [space] in the rail using the admin's own connection — no public
+  /// join. As an instance admin we already have access, so we just surface the
+  /// space and navigate into it rather than hitting the member-join endpoint.
+  void _open(AccordSpace space) {
+    final state = ref.read(accordAuthProvider);
+    final key = state is AccordAuthLoggedIn ? state.session.key : null;
+    // Flip the active server *before* upserting: `setActiveServer` reseeds the
+    // rail from the connection's cached space list, which would otherwise wipe
+    // out the space we surface here.
+    if (key != null) ref.read(accordAuthProvider.notifier).setActiveServer(key);
+    ref.read(spacesControllerProvider.notifier).upsertSpace(space);
+    context.go('/spaces?space=${Uri.encodeComponent(space.id)}');
   }
 
   Future<void> _create() async {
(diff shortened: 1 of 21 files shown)


krazyjakee:
## Code review — findings and fixes (commit `2b3900c`)

All findings below have been fixed and pushed to the PR branch.

---

### 🔴 HIGH — `_updateRole` allows empty `name` to reach the API

**File:** `lib/features/developer/services/mcp_tools.dart` · `_updateRole`

The original code:
```dart
if (args.containsKey('name')) data['name'] = (args['name'] ?? '').toString();
```
If an agent passes `name: ""` explicitly, the empty string is forwarded to `client.roles.update(...)`. The server would either accept a blank role name or return an opaque error. **Fixed** with an explicit guard:
```dart
if (args.containsKey('name')) {
  final name = (args['name'] ?? '').toString();
  if (name.isEmpty) return {'error': 'name cannot be empty'};
  data['name'] = name;
}
```

---

### 🟡 MEDIUM — `_listPermissions` unnecessarily marked `async`

**File:** `lib/features/developer/services/mcp_tools.dart` · `_listPermissions`

The handler never `await`s anything but was declared `async`, causing an unnecessary microtask allocation on every call. Replaced with an expression body using `Future.value(...)`, consistent with other pure handlers in the class (e.g. `_openDm`).

---

### 🟡 MEDIUM — No tests for new tools

**File:** `test/features/developer/mcp_tools_test.dart` (new)

No tests existed for the developer MCP layer. Added a test suite covering:
- Tool registration: all 10 new tools exist and are placed in the correct group (`read` vs `manage`)
- Schema validation: required-field arrays are correct (e.g. `set_channel_permission` requires `channel_id`, `target_id`, `target_type`)
- `list_permissions` handler: returns `ok: true`, includes all `AccordPermission.all()` entries with non-empty descriptions
- `update_role` validation: rejects `name: ""` and the no-fields-to-update path
- `set_channel_permission`: rejects invalid and empty `target_type`
- All network-bound manage tools: return `Not connected` when no `AccordAuthLoggedIn` state is present

---

### ℹ️ LOW — Observation: `_overwriteBrief` skips `AccordPermissionOverwrite` deserialization (intentional)

`channels.listOverwrites` returns raw JSON without calling `.deserializeArray(AccordPermissionOverwrite.fromJson)`. `_overwriteBrief` therefore reads the raw map directly. This is actually **correct** for round-trip consistency: `AccordPermissionOverwrite.fromJson` normalises the `type` field from `"member"` → `"user"` internally, but the MCP `set_channel_permission` tool accepts `"member"`. By using the raw server response, `list_channel_permissions` output can be fed directly back into `set_channel_permission`. No change made, but worth documenting.

---

### ℹ️ File size

`mcp_tools.dart` is now ~998 lines — just under the 1 000-line threshold. If future tool groups are added here, consider splitting the handlers into a separate `_mcp_manage_handlers.dart` mixin or part file.


---

Issue #81, opened 2026-06-09
Background / push notifications (only foreground local notifications today)


krazyjakee:
## Summary
Notifications only fire as local mention notifications while the app is running and connected to the gateway. There is no background or push delivery, so mentions/DMs are missed when the app is closed.

## Current state
- `lib/features/notifications/*` + `flutter_local_notifications`: `showMentionNotification` is fired from the live gateway handler's mention listener (step 7j). No-ops on web.
- Firebase push was removed in step 1 (intentional — no Discord/Firebase).
- `docs/PROGRESS.md` 7j covers only in-app/local notifications; product-spec §6.6 + open question note any push solution "must be Accord/Daccord-native and is a later concern."

## Work / decision needed
- Decide the Daccord-native delivery mechanism: server-sent push, background polling, or a background gateway service.
- Implement background/closed-app notification delivery on mobile + desktop.
- Web notification story.

_Source: product-spec.md §6.6 + §10; docs/PROGRESS.md §7j._

---
_Migrated from DaccordProject/daccord-app#5_

krazyjakee:
Deferred by maintainer decision (2026-08-05) — keeping open, not scheduled.

The issue's own "decision needed" is still the blocker: which Daccord-native delivery mechanism. Recording the options as they stand today so the next person doesn't re-derive them:

- **Extend the existing Android background service.** `backgroundConnectionControllerProvider` and a foreground service already exist (wired in `main.dart`, gated on the "Background connection" setting, no-op elsewhere and on Play builds). Extending that to keep the gateway alive and fire notifications when closed is fully client-side and shippable now — but covers Android only.
- **Server-sent push.** The end state the product spec implies (§6.6, §10), and the only route that covers iOS. accordserver has no push or device-registration story today, so the client half would sit inert — the same situation as #193.

Related: #137 deliberately left out the backgrounded missed-call notification because it depends on this.

krazyjakee:
Confirmed iOS reproduction for the background case:

1. Put Daccord in the background on the recipient's iOS device.
2. Send that user a direct message from another account/device.
3. No notification appears on the recipient device.

Device, iOS version, and Daccord version were not provided yet.

The same test also found that DMs do not notify while the app is in the foreground. That is tracked separately because the foreground local-notification path already exists: #326.

krazyjakee:
[...]

PR #117, opened 2026-06-11
feat(android): survive backgrounding — reconnect on resume + optional foreground service

krazyjakee:
The app lost its gateway connection whenever Android froze the process
(cached-app freezer / Doze): heartbeats stopped, sockets died silently,
the automatic reconnect budget burned out while frozen, and nothing
recovered on refocus. Notifications also never showed on Android 13+
because POST_NOTIFICATIONS was never declared or requested.

accordkit:
- GatewaySocket.ensureConnected(): revives a dead socket with a fresh
  backoff budget (resuming the session when one is held), and probes a
  nominally-connected socket with an out-of-band heartbeat, force-closing
  it into the reconnect path when the ACK never arrives. Forwarded on
  AccordClient; covered by six new gateway tests.

app:
- AppLifecycleListener in MainWindow calls AccordAuth.ensureConnectedAll()
  on resume, so every server connection is verified/revived on refocus.
- A re-READY (fresh identify gets no event replay) now refetches the
  history of every open message pane so panes catch up on what was missed.
- POST_NOTIFICATIONS declared + requested at startup (Android 13+).

background connection (new, opt-in setting):
- BackgroundConnectionService: a dataSync foreground service that keeps
  the process out of the cached-app freezer while backgrounded, so the
  gateway stays connected and mention notifications keep firing. Driven
  by a new 'Background connection' settings toggle (Android only) via a
  Riverpod controller and a small platform channel in MainActivity.

Partially addresses #81 (the app-alive half; closed-app push delivery
remains a server-side follow-up).

https://claude.ai/code/session_01DG54V8d6B9scQEvcU168RQ

### DIFF
--- a/android/app/src/main/AndroidManifest.xml
+++ b/android/app/src/main/AndroidManifest.xml
@@ -7,8 +7,19 @@
     <uses-permission
         android:name="android.permission.READ_EXTERNAL_STORAGE" />
     <uses-permission
-        android:name="android.permission.MANAGE_EXTERNAL_STORAGE" />. <uses-permission
+        android:name="android.permission.MANAGE_EXTERNAL_STORAGE" />
+    <uses-permission
         android:name="android.permission.VIBRATE" />
+    <!-- Android 13+: runtime permission to post notifications (mentions and
+         the background-connection service's status notification). -->
+    <uses-permission
+        android:name="android.permission.POST_NOTIFICATIONS" />
+    <!-- Background connection: a dataSync foreground service keeps the process
+         (and so the gateway socket) alive while the app is backgrounded. -->
+    <uses-permission
+        android:name="android.permission.FOREGROUND_SERVICE" />
+    <uses-permission
+        android:name="android.permission.FOREGROUND_SERVICE_DATA_SYNC" />
 
 
     <uses-feature
@@ -72,6 +83,13 @@
                 <data android:scheme="daccord" />
             </intent-filter>
         </activity>
+        <!-- Optional (settings-gated) foreground service that keeps the app
+             process out of the cached-app freezer while backgrounded, so the
+             gateway connection and mention notifications keep working. -->
+        <service
+            android:name=".BackgroundConnectionService"
+            android:exported="false"
+            android:foregroundServiceType="dataSync" />
         <!-- Don't delete the meta-data below.
              This is used by the Flutter tool to generate GeneratedPluginRegistrant.java -->
         <meta-data
--- a/android/app/src/main/kotlin/com/daccord_projects/daccord/BackgroundConnectionService.kt
+++ b/android/app/src/main/kotlin/com/daccord_projects/daccord/BackgroundConnectionService.kt
@@ -1,0 +1,93 @@
+package com.daccord_projects.daccord
+
+import android.app.NotificationChannel
+import android.app.NotificationManager
+import android.app.PendingIntent
+import android.app.Service
+import android.content.Context
+import android.content.Intent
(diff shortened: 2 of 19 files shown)


krazyjakee:
## Code Review — PR #117

Reviewed all 19 changed files. Four issues found and fixed in commit cb23c6d. Ranked by severity:

---

### 🔴 HIGH — `_probeLiveness()` race with heartbeat timer

**File:** `packages/accordkit/lib/src/gateway/gateway_socket.dart`

The probe reset `_heartbeatAckReceived = false` and then `await`-ed `probeTimeout` (5 s). If the heartbeat timer happened to fire in that window it would see `false`, conclude the connection was stale, and call `_conn?.close(4000, 'heartbeat timeout')` — killing a live connection before the probe ACK had a chance to arrive. With a typical heartbeat interval of 30 s this had a ~1-in-6 chance of triggering every time the user foregrounded the app.

**Fix:** `_stopHeartbeat()` at the top of `_probeLiveness()`, then `_startHeartbeat()` in the success branch (ACK received) to resume normal heartbeating.

---

### 🟠 MEDIUM — `ensureConnected()` could probe mid-handshake

**File:** `packages/accordkit/lib/src/gateway/gateway_socket.dart`

When `_state` was `connecting` or `resuming`, `ensureConnected()` fell through to `_probeLiveness()` and sent an out-of-band heartbeat opcode while the server was still waiting for IDENTIFY or RESUME. This is undefined behaviour against the Accord gateway.

**Fix:** Added guard `if (_state != GatewayState.connected) return;` — if a handshake is already in flight, let it complete.

---

### 🟠 MEDIUM — `BackgroundConnectionService` used `START_STICKY`

**File:** `android/app/src/main/kotlin/com/daccord_projects/daccord/BackgroundConnectionService.kt`

`START_STICKY` tells Android to auto-restart the service if the OS kills the process. But the service holds no logic of its own — its only job is to exempt the **running** Dart isolate from the cached-app freezer. If the process is killed, there is no Flutter engine and no gateway, so an auto-restarted service would hold a foreground notification slot without doing anything useful.

**Fix:** Changed to `START_NOT_STICKY`. The Riverpod `BackgroundConnectionController` re-starts the service when the user brings the app back.

---

### 🟡 LOW — `reload()` didn't reset `isLoadingOlder`

**File:** `lib/features/messaging/controllers/accord_messages.dart`

If `loadOlder()` was in flight when a re-READY triggered `reload()`, both futures would race to write `state`. `loadOlder` captures a snapshot of `current` before its REST call; if `reload` completed first (replacing `state` with fresh messages), `loadOlder` would then prepend its stale page on top.

**Fix:** `reload()` now sets `isLoadingOlder = false` before calling `_load`, signalling any in-flight `loadOlder` that its result is superseded.

---

### 🟡 LOW — `unawaited` future not explicit

**File:** `packages/accordkit/lib/src/gateway/gateway_socket.dart`

`ensureConnected()` discarded the `Future<void>` returned by `_probeLiveness()` without `unawaited()`. Fixed alongside the guard above.

---

### Tests added (2 new cases)

- **`is a no-op when the handshake is still in progress`** — calls `ensureConnected()` while `_state == connecting` and verifies no probe heartbeat is sent and no second connection is opened.
- **`heartbeat timer does not race with probe during ack window`** — starts a socket with a HELLO (so the heartbeat timer is active), runs a gated probe, delivers an ACK, and verifies the connection remains intact (i.e., the timer did not fire during the probe window and close the socket).

No structural issues, no files over 1000 lines, no 3D model changes.

---

PR #118, opened 2026-06-11
Inset top banners below the OS status bar on mobile

krazyjakee:
The update and web-update banners render in the top-level Column above
the body's SafeArea, so on Android they slid under the status bar and
their text/dismiss controls were partly unreachable. Wrap each in
SafeArea(bottom: false), matching the existing RolePreviewBanner.

### DIFF
--- a/lib/features/updates/views/update_banner.dart
+++ b/lib/features/updates/views/update_banner.dart
@@ -29,31 +29,34 @@ class UpdateBanner extends ConsumerWidget {
     final colors = BonfireThemeExtension.of(context);
     return Material(
       color: colors.primary,
-      child: InkWell(
-        onTap: () => showUpdatesSettings(context),
-        child: Padding(
-          padding: const EdgeInsets.symmetric(horizontal: 12, vertical: 6),
-          child: Row(
-            children: [
-              const Icon(Icons.system_update, size: 16, color: Colors.white),
-              const SizedBox(width: 8),
-              Expanded(
-                child: Text(
-                  'Update available — ${release.name}. Tap to view.',
-                  overflow: TextOverflow.ellipsis,
-                  style: const TextStyle(color: Colors.white, fontSize: 13),
+      child: SafeArea(
+        bottom: false,
+        child: InkWell(
+          onTap: () => showUpdatesSettings(context),
+          child: Padding(
+            padding: const EdgeInsets.symmetric(horizontal: 12, vertical: 6),
+            child: Row(
+              children: [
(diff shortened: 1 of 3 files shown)


krazyjakee:
## Review findings (ranked by severity)

### MEDIUM — Double `SafeArea` padding when two banners are simultaneously visible (pre-existing, not introduced here)

`accord_home.dart` stacks the banners in a `Column`:
```dart
const UpdateBanner(),       // SafeArea(bottom: false)  ← new in this PR
const WebUpdatePrompt(),    // SafeArea(bottom: false)  ← new in this PR
const RolePreviewBanner(),  // SafeArea(bottom: false)  ← pre-existing
Expanded(child: body),
```
Flutter's `SafeArea` reads `MediaQuery.padding.top` but does **not** consume it from the widget tree. So if `UpdateBanner` and `RolePreviewBanner` are both visible on mobile (update available + user is in role-preview mode), `RolePreviewBanner` independently re-applies the full status-bar inset (~24–32 dp), giving it a redundant top gap.

**Status:** Pre-existing — `RolePreviewBanner` already used `SafeArea(bottom: false)` before this PR. This PR is consistent with that prior choice. Tracking as a follow-up: the clean fix is to lift the single `SafeArea` up to the Column level in `accord_home.dart` and strip it from each individual banner (using `MediaQuery.removePadding` or a wrapping `SafeArea` around a `mainAxisSize: min` Column that excludes the body). Scope is intentionally left for a separate PR.

---

### LOW — `WebUpdatePrompt`'s `SafeArea` is a no-op on web (harmless)

`WebUpdatePrompt` only ever renders when `kIsWeb == true`, where `MediaQuery.padding.top == 0` (browsers have no overlapping OS status bar). The `SafeArea(bottom: false)` added here is harmless and keeps the widget consistent with its siblings. No change needed.

---

### LOW — No tests for banner show/hide logic or SafeArea wrapping (fixed)

Commit `6aa84ce` adds `test/features/updates/update_banner_test.dart` covering:
- Collapses to zero height when no update is available
- Renders banner text and `SafeArea(bottom: false)` when an update is available
- Collapses when `dismissedVersion` matches the release version
- Collapses when `skippedUpdateVersion` (settings) matches the release version
- `WebUpdatePrompt` renders nothing on non-web platforms

---

### Summary

The core change — wrapping both banners in `SafeArea(bottom: false)` to match the pre-existing `RolePreviewBanner` pattern — is correct and consistent. The only fix committed here is the missing test coverage. The double-SafeArea edge case is pre-existing and low-impact enough to track separately.

---

PR #119, opened 2026-06-12
docs: clarify http:// prefix needed for local/self-hosted servers

krazyjakee:
## Summary

Connecting the desktop client to a local or self-hosted Accord server using a bare address (e.g. `localhost:39099`) fails with a `ClientException: Broken pipe` error. The cause: when no scheme is entered, the client assumes `https://` (`AccordServer.normalizeBaseUrl`), then attempts a TLS handshake against a plain-HTTP port — the server reads the handshake as malformed HTTP and drops the socket.

This PR is **docs-only**. Per discussion, the client's scheme behavior is intentionally left unchanged (bare hostnames default to `https://`, keeping public servers secure; an explicit `http://` is honored). The docs are updated to steer users toward the `http://` prefix for local/LAN servers.

## Changes

- **`docs/self-hosting/desktop-app.md`** — connect steps now use `http://localhost:39099` / `http://192.168.1.50:39099`, with a callout explaining the broken-pipe cause.
- **`docs/getting-started/adding-a-server.md`** — URL-format section notes that local/self-hosted servers need the `http://` prefix.
- **`docs/troubleshooting/common-issues.md`** — added a "Broken pipe error on a local server" bullet under *Can't Connect to a Server*.
- **`docs/self-hosting/deploying-a-server.md`** — connect step distinguishes the HTTPS-reverse-proxy case from direct `http://your-ip:39099`.

## Test plan

- [x] Docs-only change; no code touched.
- [x] Verified the workaround: entering `http://127.0.0.1:39099` connects successfully where the bare address produced the broken-pipe error.

### DIFF
--- a/docs/getting-started/adding-a-server.md
+++ b/docs/getting-started/adding-a-server.md
@@ -22,7 +22,9 @@ The simplest URL is just a hostname like `chat.example.com`. You can also includ
 
 - A port number: `chat.example.com:8443`
 - A specific space: `chat.example.com#my-space` (defaults to "general" if omitted)
-- A protocol: `https://chat.example.com` (HTTPS is used by default)
+- A protocol: `https://chat.example.com` (HTTPS is used by default when no scheme is given)
+
+> **Connecting to a local or self-hosted server?** Add the `http://` prefix, e.g. `http://localhost:39099` or `http://192.168.1.50:39099`. Self-hosted servers serve plain HTTP on the local network, and without the prefix the client assumes `https://` and the connection fails with a "Broken pipe" error.
 - A pre-filled token: `chat.example.com?token=yourtoken` (logs you in automatically)
 - An invite code: `chat.example.com?invite=yourcode` (joins the space using the invite)
 
--- a/docs/self-hosting/deploying-a-server.md
+++ b/docs/self-hosting/deploying-a-server.md
@@ -133,7 +133,7 @@ Make sure the following ports are accessible:
 
 ## Connecting from daccord
 
-Once your server is running, open daccord and click the **+** button in the sidebar to add a new server. Enter your server's address (e.g., `chat.example.com` or `your-ip:39099`) and create an account.
+Once your server is running, open daccord and click the **+** button in the sidebar to add a new server. Enter your server's address and create an account. Use `chat.example.com` for a server behind an HTTPS reverse proxy, or `http://your-ip:39099` when connecting directly over the local network — the `http://` prefix is required there, since a bare address defaults to `https://` and a plain-HTTP server will fail with a "Broken pipe" error.
 
 ## Updating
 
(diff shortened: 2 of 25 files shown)


krazyjakee:
## Review findings and fixes

Fixes pushed in commit `29ca87f` on `claude/eager-curie-102mx2`.

---

### 🔴 Severity 1 — Crash bug: empty-folder `StateError` in the space rail

**Files:** `lib/features/spaces/views/accord_home_rail.dart`

`_buildUnits` can produce a `_RailUnit.folder` whose `spaces` list is empty — this happens when every space in a folder belongs to a server the client is currently not connected to (all space IDs are in `orderedIds` but none are present in `byId`). Two call-sites then call `.first` on that empty list and throw `StateError`:

1. `_unitAnchor` — `unit.spaces.first.id`
2. `onReorderFolderBefore` lambda — `unit.spaces.first.id`

**Fix applied:**
- `_buildUnits`: skip folders whose visible member list is empty (`if (members.isNotEmpty) units.add(...)`) — an all-unavailable folder has nothing useful to show.
- `_unitAnchor`: added a guard (`unit.spaces.isNotEmpty ? unit.spaces.first.id : ''`) so a stale unit can never crash even if one slips through.
- `onReorderFolderBefore` lambda: replaced the direct `.first.id` with `unit.spaces.isNotEmpty ? unit.spaces.first.id : null` (null maps to "append at end" in `_moveFolderBefore`).

---

### 🟡 Severity 2 — Misleading test name

**File:** `test/features/developer/mcp_tools_test.dart`

The test named `'returns not-connected when no fields to update'` actually asserts `result['error'] == 'No fields to update'` — a validation error that fires *before* any network call, never 'Not connected'. The name implies the opposite of what's tested.

**Fix applied:** Renamed to `'returns no-fields error without hitting client'`.

---

### 🟡 Severity 2 — Discord branding in source comment (project rule violation)

**File:** `lib/shared/components/image_crop_dialog.dart`

The doc-comment read:
> *"Mirrors the reposition+zoom flow Discord uses for server icons/banners and user avatars."*

CLAUDE.md explicitly prohibits Discord references in the codebase.

**Fix applied:** Removed the Discord attribution; the functional description of the crop flow is retained.

---

### 🟢 Severity 3 — Informational: `_createRole` validates `name` after the client check

**File:** `lib/features/developer/services/mcp_tools.dart` (not changed)

`name.isEmpty` is checked after `client == null`, so a caller who supplies an empty name while disconnected sees `'Not connected'` rather than `'name is required'`. This is intentional — commit `8282a2b` explicitly reordered the checks to match the existing tests — but it's slightly inconsistent with the `update_role` handler (which validates `name` before the client check). No action taken; flagging for awareness.

---

### 🟢 Severity 3 — Informational: mobile folder-menu access changed

**File:** `lib/features/spaces/views/accord_home_rail_tiles.dart` (not changed)

`_FolderTile` previously opened its management menu on `onLongPressStart`. It now uses `onDoubleTap` / `onSecondaryTapUp`, because long-press initiates a `LongPressDraggable` drag on touch platforms. Mobile users who are used to long-pressing the folder icon to open the menu will need to double-tap instead. This appears intentional but is worth UX validation on a physical device.

---

**No 3D/visual rendering changes** were present; the visual-review step was skipped.


---

PR #122, opened 2026-06-13
fix(error-reporting): point default DSN at GlitchTip project 2

krazyjakee:
## What

The baked-in default GlitchTip DSN targeted **project 1** (`https://…@crash.daccord.gg/1` → `https://crash.daccord.gg/api/1/store/`), so opt-in crash reports landed in the wrong project. This points it at the correct **project 2** (`cattrallgames`) dashboard.

```diff
-    'https://8ca39e8b380a42548f2d8e9468d47ac2@crash.daccord.gg/1';
+    'https://f7c1b4791bdc42daac0b6411d41c8c4f@crash.daccord.gg/2';
```

## Notes

- Reporting remains **opt-in** — nothing is sent until the user enables `errorReportingEnabled`.
- Tests use generic example DSNs and don't assert on the default constant, so no test changes are needed.
- ⚠️ **Action item outside this PR:** release builds use the `SENTRY_DSN` GitHub secret if set (`release.yml`). If that secret still holds the old project-1 DSN, it will override this default. Update or clear it so release builds also report to project 2.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,74 +1,210 @@
-# Daccord (Flutter)
+# Daccord — Flutter Client
 
-A free, open-source chat app for communities. Connect to one or more [accordserver](https://github.com/DaccordProject/accordserver) instances and chat in real time with your friends, teammates, or community.
+### Communication without compromise.
 
-This is the **Flutter client** — a fork of [Bonfire](https://github.com/OpenBonfire/bonfire) that reuses its mature Flutter UI and **replaces the Discord networking layer with the Accord protocol** via [`accordkit-dart`](https://github.com/DaccordProject/accordkit-dart). It aims for feature parity with the Godot-based reference client, [`daccord`](https://github.com/DaccordProject/daccord).
+**Daccord** is the free, open-source alternative to Discord — chat, voice, and video with native apps you control. This repository is the **Flutter client**: a fast, beautiful, cross-platform app that talks to *your* server, not someone else's cloud.
 
-> **Status:** early development. The Accord networking foundation, auth, read path, and a first write path are in place.
+No ads. No tracking. No corporate lock-in. **Your server, your data, your rules.**
 
-## Features
+<p align="left">
+  <img alt="License: GPLv3" src="https://img.shields.io/badge/license-GPLv3-blue.svg">
+  <img alt="Built with Flutter" src="https://img.shields.io/badge/built%20with-Flutter-027DFD.svg">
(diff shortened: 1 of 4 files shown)


krazyjakee:
## Code Review — findings & fixes (commit `624bd9a`)

The core one-line DSN change is correct and focused. Two improvements were pushed on top of it, ranked by severity:

---

### 🟡 Medium — Missing regression test for the DSN constant

**Finding:** The existing test `'the dev DSN is used when no dart-define override is set'` only asserts `resolveGlitchTipDsn() == kDefaultGlitchTipDsn` — a tautology that passes regardless of the constant's value. If the DSN were silently reverted or typo'd to point at the wrong project, no test would catch it.

**Fix applied:** Added a new test that parses `kDefaultGlitchTipDsn` through `GlitchTipClient.init()` and asserts the derived store URL contains `/api/2/store/`:

```dart
test('kDefaultGlitchTipDsn is parseable and targets project 2', () {
  final c = GlitchTipClient();
  expect(c.init(kDefaultGlitchTipDsn), isTrue, reason: '...');
  expect(c.storeUrl, contains('/api/2/store/'), reason: '...');
});
```

This test would have caught the project-1 → project-2 bug this PR fixes.

---

### 🟢 Low — Silent HTTP 4xx/5xx failures in `GlitchTipClient._sendEvent`

**Finding:** `_sendEvent` discarded the HTTP response entirely. If the DSN key or project ID were wrong, every event would silently return a 401/404 with zero developer feedback — exactly the failure mode that motivated this PR.

**Fix applied:** The response is now captured; in debug mode, any 4xx/5xx logs to `debugPrint` with the store URL. The "never throw" invariant is preserved — the log is inside the existing `try` block, which swallows exceptions from `post` itself.

```dart
final response = await _http.post(...);
if (kDebugMode && response.statusCode >= 400) {
  debugPrint('[GlitchTip] event rejected: HTTP ${response.statusCode} (store URL: $_storeUrl)');
}
```

---

### ✅ No issues found

- **File size:** both files are well under 1000 lines (controller: ~185 lines, client: ~210 lines, test: ~345 lines).
- **Code quality:** `scrubPiiText`, `_truncateId`, breadcrumb ring-buffer, `_hooksInstalled` guard, and `ref.onDispose` teardown are all correct.
- **Opt-in invariant:** confirmed — nothing is sent until `errorReportingEnabled` is true; the consent flags survive JSON round-trip (covered by existing tests).
- **Action item from PR description:** the `SENTRY_DSN` GitHub secret should still be updated to the project-2 key if set — this is outside the scope of this code change.


---

PR #125, opened 2026-06-13
Fix mobile left-edge swipe popping home to sign-in

krazyjakee:
On narrow layouts the home screen (/spaces) is a child of the login
route (/). Android's left-edge back gesture shares the same edge as the
swipe-to-open-sidebar gesture, so it would pop the home route down to
the sign-in screen instead of opening the channel drawer.

Wrap the mobile Scaffold in a PopScope (canPop: false) that intercepts
the system back gesture and steps through / opens the drawers instead:
close the member end-drawer, else close the channel drawer, else open
the channel drawer. The left swipe now opens the left sidebar as
expected.

### DIFF
--- a/lib/features/spaces/views/accord_home.dart
+++ b/lib/features/spaces/views/accord_home.dart
@@ -578,73 +578,95 @@ class _AccordHomeScreenState extends ConsumerState<AccordHomeScreen> {
         // Narrow: rail + channel list move into a drawer, members into an
         // end-drawer, so the message pane keeps full width instead of
         // overflowing.
-        return Scaffold(
-          key: _scaffoldKey,
-          backgroundColor: colors.background,
-          drawerEdgeDragWidth: 48,
-          drawer: Drawer(
-            width: 292,
+        //
+        // Wrap in [PopScope] so Android's left-edge back swipe (which shares the
+        // same edge as the swipe-to-open-sidebar gesture) steps through / opens
+        // the drawers instead of popping the route. The home (`/spaces`) is a
+        // child of the login route (`/`), so an un-intercepted system back would
+        // pop down to the sign-in screen — the cause of the "swiping left
+        // re-prompts for sign in" bug.
+        return PopScope(
+          canPop: false,
+          onPopInvokedWithResult: (didPop, _) {
+            if (didPop) return;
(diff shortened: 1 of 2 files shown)


krazyjakee:
## Code Review — findings & fixes

Reviewed `lib/features/spaces/views/accord_home.dart` (the only changed file). The fix is correct: `PopScope(canPop: false)` is the right tool here, the callback priority order (end-drawer → drawer → open drawer) matches the described behavior, and no Discord/Firebase code was re-introduced.

### Findings ranked by severity

---

#### 🔴 High — Missing test coverage (fixed in this review)

The new `PopScope` behavior had no tests. The three back-gesture cases and the core regression (back must not pop to sign-in) are now covered in a new file:

**`test/features/spaces/mobile_pop_scope_test.dart`** (commit `96d2aac`)

Tests added:
1. Back when no drawer is open → opens the channel drawer
2. Back when the channel drawer is open → closes it
3. Back when the end drawer (members) is open → closes it
4. Back never pops the route back to the sign-in screen (regression guard)
5. Works correctly when there is no end drawer (no space selected)

The `AccordHomeScreen` widget has too many Riverpod/Hive dependencies to drive end-to-end in a unit test; the tests use a minimal `StatefulWidget` that replicates the same `PopScope + Scaffold + GlobalKey` structure so the drawer state machine can be exercised directly.

---

#### 🟡 Medium — `_MessagePane` construction is duplicated (pre-existing, not introduced here)

The `messageArea` variable (used in the wide layout) builds `_MessagePane` with identical parameters to the inline construction in the narrow layout body. Any future change to those parameters must be applied in both places. This is a pre-existing issue outside the scope of this PR — noting it here for a follow-up refactor.

---

#### 🟢 Low — Comment verbosity

The 7-line comment block before the `PopScope` is slightly longer than the project style recommends ("only add a comment when the WHY is non-obvious"). The WHY here (Android left-edge swipe == system back gesture) genuinely is non-obvious, so the comment is justified — but could be condensed to 3–4 lines. Not blocking.

---

### Summary

One commit pushed to this branch: adds 5 widget tests covering all new behavior paths. No source-code changes were needed — the implementation is correct as written.

---

PR #128, opened 2026-06-13
fix(messaging): make message header times date-aware

krazyjakee:
## Problem

The message header timestamp always rendered as bare `HH:MM`, so a message sent days ago looked identical to one sent today (e.g. `18:43`).

## Change

`_time` in `accord_home_message_row.dart` now shows progressively more context the older a message is:

| When | Shown |
|------|-------|
| Today | `18:43` |
| Yesterday | `Yesterday at 18:43` |
| 2–6 days ago | `Monday at 18:43` |
| Older | `12/06/2026 18:43` |

The 6-day bound keeps the weekday unambiguous (a 7-day-old message would repeat today's weekday name), so it falls through to the calendar date.

The grouped-message hover gutter is a fixed narrow slot (`avatarRadius * 2`) where the longer strings would overflow, so it uses a new bare-clock `_clock` getter. Grouped rows are within minutes of their header anyway, so the clock alone is correct there. The full hover tooltip (`Monday, 5 June 2026 at 14:30`) is unchanged.

No new dependencies — still hand-formatted (intl isn't a dep).

### DIFF
--- a/lib/features/spaces/utils/message_time.dart
+++ b/lib/features/spaces/utils/message_time.dart
@@ -1,0 +1,39 @@
+const _weekdays = [
+  'Monday', 'Tuesday', 'Wednesday', 'Thursday',
+  'Friday', 'Saturday', 'Sunday',
+];
+
+/// Bare `HH:MM` string for a local [DateTime].
+String messageClockString(DateTime local) {
+  final hh = local.hour.toString().padLeft(2, '0');
+  final mm = local.minute.toString().padLeft(2, '0');
+  return '$hh:$mm';
+}
+
+/// Date-aware label for a message timestamp shown in the message header.
+///
+/// | Age      | Result                  |
+/// |----------|-------------------------|
+/// | Today    | `18:43`                 |
+/// | Yesterday| `Yesterday at 18:43`    |
+/// | 2–6 days | `Monday at 18:43`       |
+/// | Older    | `12/06/2026 18:43`      |
+///
+/// The 6-day ceiling keeps weekday names unambiguous (a 7-day-old message
+/// would repeat today's weekday). [now] is injected for testability; callers
+/// omit it and the current wall-clock time is used.
+String messageTimeString(DateTime local, {DateTime? now}) {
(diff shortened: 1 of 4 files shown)


krazyjakee:
## Review findings and fixes (commit 98d2f9f)

All three issues below have been fixed in the follow-up commit on this branch.

---

### 🔴 High — No test coverage

The date-aware `_time` logic has five distinct rendering cases and non-trivial boundary conditions (today/yesterday cut, 6-day weekday ceiling), but shipped with zero tests. A regression here (wrong day label, off-by-one on the weekday window) would be invisible.

**Fix:** Extracted the core logic into `lib/features/spaces/utils/message_time.dart` as two pure functions (`messageClockString`, `messageTimeString`) and added `test/features/spaces/message_time_test.dart` with 9 cases covering all branches, clock padding, the full weekday cycle, and the future-timestamp edge case.

---

### 🟡 Medium — Double timestamp parse

`_time` called `_clock` to reuse the HH:MM string, but `_clock` re-ran `DateTime.tryParse(_message.timestamp)` independently. Every render of a non-grouped header triggered two parses of the same string.

**Fix:** Both getters now parse once and delegate to the extracted utility functions. `messageTimeString` computes the clock string from the already-converted `DateTime` rather than calling back into a getter.

---

### 🟡 Medium — Dead condition

```dart
if (daysAgo >= 2 && daysAgo <= 6) {  // the >= 2 half is unreachable
```

The `daysAgo == 1` branch above already handles 1, so `daysAgo` is always ≥ 2 when this line is reached.

**Fix:** Simplified to `if (daysAgo <= 6)`.

---

### ℹ️ Minor — Future-dated timestamps

The original fell through to the calendar-date format for messages with `timestamp` in the future (`daysAgo < 0`). The new code treats future timestamps the same as "today" (bare clock). This is arguably better UX and is covered by a dedicated test case.

---

**No new dependencies were added.** No visual changes beyond what the PR already describes. The file is well under 1 000 lines so no structural split was needed.

---

PR #129, opened 2026-06-13
fix(reactions): dedup unicode reaction so it shows once

krazyjakee:
## What

Adding a unicode emoji reaction from the emoji picker showed up **twice** (two pills). This fixes it to show a single pill at the correct count.

## Why

The emoji picker passes `toggleReaction` a unicode emoji's **shortcode** name (e.g. `hamburger`), so the optimistic pill was stored/keyed by that shortcode. The gateway's `reaction.add` echo arrives keyed by the **glyph** (`🍔`) — what the server actually stored. The dedup key (`_emojiName`) returned the shortcode for the optimistic pill and the glyph for the echo, so they never matched: the echo created a **second pill** instead of merging, and the `includesMe` double-count guard never fired.

Tapping an *existing* pill didn't double, because that path already passes the glyph — only the picker path passed a shortcode.

## How

Introduce a canonical `_emojiKey(name, id)` helper that resolves unicode emoji to their glyph (custom emoji still key on name, with the id carried separately). All three reaction touchpoints now route through it:

- `toggleReaction` — the "do I already have this reaction?" lookup
- `applyReaction` — the match-or-create index, and it now **stores** the canonical key as the reaction's name
- `_emojiName` — delegates to `_emojiKey`

So `hamburger` (optimistic) and `🍔` (gateway echo) both normalize to `🍔` and collapse to one pill.

## Notes for reviewers

- The pill renderer already resolves names to glyphs via `resolveEmojiGlyph`, so storing the glyph as the reaction name renders unchanged.
- I couldn't run `flutter analyze` locally (the toolchain requires macOS 14+; this machine is on 13.0) — relying on CI. The change is self-contained; `resolveEmojiGlyph`/`parseEmojiToken` were already imported and used in the file.
- Worth a manual check: react with a standard unicode emoji from the picker and confirm a single pill at count 1.

### DIFF
--- a/lib/features/messaging/controllers/accord_messages.dart
+++ b/lib/features/messaging/controllers/accord_messages.dart
@@ -321,7 +321,7 @@ class AccordMessagesController extends _$AccordMessagesController {
     final message = state?.firstWhereOrNull((m) => m.id == messageId);
     if (message == null) return;
     final existing = message.reactions?.firstWhereOrNull(
-      (r) => _emojiName(r) == emojiName,
+      (r) => _emojiName(r) == _emojiKey(emojiName, emojiId),
     );
     final adding = !(existing?.includesMe ?? false);
     final token = emojiId == null
@@ -408,8 +408,9 @@ class AccordMessagesController extends _$AccordMessagesController {
     final message = current.firstWhereOrNull((m) => m.id == messageId);
     if (message == null) return;
 
+    final key = _emojiKey(emojiName, emojiId);
     final reactions = [...(message.reactions ?? const <AccordReaction>[])];
-    final index = reactions.indexWhere((r) => _emojiName(r) == emojiName);
+    final index = reactions.indexWhere((r) => _emojiName(r) == key);
 
     if (added) {
       if (index >= 0) {
@@ -420,7 +421,9 @@ class AccordMessagesController extends _$AccordMessagesController {
       } else {
         reactions.add(
           AccordReaction(
-            emoji: {'id': emojiId, 'name': emojiName},
+            // Store the canonical key (glyph for unicode) so an optimistic pill
+            // and its later gateway echo dedup to one — see [_emojiKey].
+            emoji: {'id': emojiId, 'name': key},
             count: 1,
             includesMe: isOwn,
           ),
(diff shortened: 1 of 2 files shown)


krazyjakee:
## Review findings — ranked by severity

### 🔴 HIGH — Bug: `clearReactionEmoji` comparison broken after `_emojiKey` introduction

**File:** `lib/features/messaging/controllers/accord_messages.dart:452`

`_emojiName(r)` now delegates to `_emojiKey`, which resolves unicode emoji to their canonical glyph (e.g. `hamburger` → `🍔`). But `clearReactionEmoji` was left comparing that canonical glyph against the raw `emojiName` argument, which arrives from the gateway `reaction.clear_emoji` event and can be a shortcode. The filter never matched, so the clear was silently a no-op.

The same is true for reactions loaded from REST before the PR that still have their shortcode stored as the name: `_emojiName(r)` resolves to the glyph, but the comparison was against the shortcode.

**Fix applied** (`ec3087b`):
```dart
// before
.where((r) => _emojiName(r) != emojiName)
// after
.where((r) => _emojiName(r) != _emojiKey(emojiName, null))
```

---

### 🟡 MEDIUM — No tests for reaction dedup logic

The core invariant introduced by this PR (shortcode + glyph echo → one pill) had no test coverage. Given the bug found above, tests are particularly important here.

**Added** in `test/features/messaging/accord_messages_reactions_test.dart` (`ec3087b`):
- Shortcode optimistic pill + glyph gateway echo collapse to one (the primary PR scenario)
- Existing pill tap via glyph works correctly
- REST-loaded shortcode-named reaction removed via glyph
- Custom emoji still keyed by name, not resolved to a glyph
- `clearReactionEmoji` with shortcode, glyph, and cross-emoji isolation
- Count decrement and `includesMe` guarding

---

### 🟢 LOW — No other issues found

- **Code quality:** The `_emojiKey` helper is clearly named, well-commented, and the three callsites all route through it consistently. `_emojiName` as a thin wrapper is readable.
- **File size:** 482 lines — well under threshold.
- **Correctness of the primary fix:** `toggleReaction` and `applyReaction` are correct; the optimistic add canonicalises via `_emojiKey` and stores the glyph, so the gateway echo (`🍔`) finds and merges into the pill rather than creating a second one.
- **No Discord endpoints or regressions introduced.**

> Note: Flutter/Dart toolchain is not available in this container, so tests weren't run locally — relying on CI for the green signal.


---

PR #132, opened 2026-06-13
QR-code 2FA setup and consolidated account settings

krazyjakee:
## What changed

Two related account/security UX improvements.

### 1. QR-code 2FA enrollment
The 2FA setup previously showed only the TOTP secret as plain text, forcing users to type it into their authenticator app by hand. Now it renders the server's `otpauth` URI as a scannable **QR code** (using `qr_flutter`, already a dependency).

- QR sits on a white quiet-zone background so it scans reliably regardless of the app theme.
- The secret remains as copyable text (with a copy button) as a manual-entry fallback for servers that only return a bare secret.

### 2. Consolidated account settings
Account security (password, 2FA, delete account) was only reachable from a "Account settings" item buried in the avatar/status popup menu — separate from every other profile/account setting, which all live in **Settings → Account**.

- Added a **"Password & Security"** tile to Settings → Account (under "Edit profile") that opens the same dialog.
- Removed the "Account settings" item from the status popup, so the avatar button is now a **pure status picker** (online/idle/dnd/invisible + custom status).
- Retitled the dialog from "Account settings" to "Password & Security" to match its new entry point.

**New end-to-end path:** Settings → Account → Password & Security → Enable 2FA → scan the QR code.

## Why
- Manual secret entry is error-prone; QR scanning is the expected 2FA enrollment flow.
- One predictable home for account management instead of two divergent entry points.

## Reviewer notes
- No new dependencies — `qr_flutter` was already in `pubspec.yaml`/`pubspec.lock`.
- `flutter analyze --no-fatal-infos` passes clean on all three changed files.
- Not yet exercised in a running app (no device booted in this environment); changes are UI-only and analyzer-verified.

## Files
- `lib/features/user/views/accord_account_settings.dart` — QR code + copy button; dialog retitle
- `lib/features/settings/views/accord_settings_screen.dart` — "Password & Security" tile
- `lib/features/user/components/self_status_button.dart` — remove account item; status-picker only

### DIFF
--- a/lib/features/settings/views/accord_settings_screen.dart
+++ b/lib/features/settings/views/accord_settings_screen.dart
@@ -10,6 +10,7 @@ import 'package:bonfire/features/settings/views/settings_backup.dart';
 import 'package:bonfire/features/error_reporting/controllers/error_reporting.dart';
 import 'package:bonfire/features/updates/views/updates_page.dart';
 import 'package:bonfire/features/settings/models/accord_settings.dart';
+import 'package:bonfire/features/user/views/accord_account_settings.dart';
 import 'package:bonfire/features/user/views/accord_profile_edit.dart';
 import 'package:bonfire/shared/app_info.dart';
 import 'package:bonfire/features/voice/views/voice_settings_screen.dart';
@@ -233,6 +234,15 @@ class AccordSettingsScreen extends ConsumerWidget {
             subtitle: const Text('Display name, bio, avatar'),
             trailing: const Icon(Icons.chevron_right),
             onTap: () => showAccordProfileEdit(context),
+          ),
+          ListTile(
+            leading: Icon(Icons.lock_outline, color: colors.dirtyWhite),
+            title: const Text('Password & Security'),
+            subtitle: const Text(
+              'Password, two-factor authentication, delete account',
+            ),
+            trailing: const Icon(Icons.chevron_right),
+            onTap: () => showAccordAccountSettings(context),
           ),
           if (ref.watch(connectionsControllerProvider).hasMultiple)
             ListTile(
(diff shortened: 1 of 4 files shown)


krazyjakee:
## Code Review — PR #132

Reviewed `accord_account_settings.dart`, `accord_settings_screen.dart`, and `self_status_button.dart`. The overall approach is clean and the consolidation is well-motivated. Three issues fixed in commit `bffccb8`, ranked below by severity.

---

### 🟡 Medium — Copy button copies raw `otpauth://` URI instead of the bare secret

**File:** `lib/features/user/views/accord_account_settings.dart`

`_enable()` parses the server response into two optional fields:
```dart
_secret = data['secret']?.toString();          // may be null
_otpauth = data['otpauth_uri']?.toString();    // may be null
```
When a server only returns `otpauth_uri` (no separate `secret`), the SelectableText and copy button both fell back to `_otpauth`, surfacing the full URI (`otpauth://totp/user?secret=BASE32&issuer=…`) as the "secret" to copy. A user pasting this into an authenticator that expects a bare Base32 key would fail.

**Fix:** Added `extractTotpSecret(String? uri)` — a package-visible helper that regex-extracts the `?secret=` parameter from an otpauth URI and passes non-URI strings through unchanged. Display and copy now use `_secret ?? extractTotpSecret(_otpauth) ?? ''`. Six unit tests added in `test/features/user/totp_uri_test.dart`.

---

### 🔵 Low — QR error-correction level was L (7%)

**File:** `lib/features/user/views/accord_account_settings.dart`

`QrImageView` defaults to `QrErrorCorrectLevel.L`, the minimum reliability tier. On-screen scans are more prone to glare and partial occlusion than printed codes; `QrErrorCorrectLevel.M` (15%) is the standard choice for software-rendered QR codes.

**Fix:** Added `errorCorrectionLevel: QrErrorCorrectLevel.M` to `QrImageView`.

---

### 🔵 Low — TOTP code field accepted unlimited input and had no keyboard submit

**File:** `lib/features/user/views/accord_account_settings.dart`

The 6-digit code `TextField` had no `maxLength`, so users could accidentally type extra digits without visual feedback, leading to silent verification failure. The keyboard "Done" action also had no effect, requiring a tap on the button.

**Fix:** Added `maxLength: 6`, `counterText: ''` (hides the counter label), `textInputAction: TextInputAction.done`, and `onSubmitted: (_) { if (!_busy) _verify(); }`.

---

No issues found in `accord_settings_screen.dart` or `self_status_button.dart` — both changes are clean.

🤖 Generated with [Claude Code](https://claude.ai/claude-code)

---

PR #142, opened 2026-06-14
fix(channels): use ReorderableListView.onReorder; gate CI on stable too

### LINKED ISSUE #138: Build error on master: accord_home_channels.dart uses undefined onReorderItem / missing onReorder
Discovered while working on #134. These are **pre-existing** on `master` (not introduced by that PR — the file is untouched there).

## Errors
\`flutter analyze\` reports two errors in \`lib/features/spaces/views/accord_home_channels.dart\`:

\`\`\`
error • The named parameter 'onReorder' is required, but there's no corresponding argument • lib/features/spaces/views/accord_home_channels.dart:855:12 • missing_required_argument
error • The named parameter 'onReorderItem' isn't defined • lib/features/spaces/views/accord_home_channels.dart:859:7 • undefined_named_parameter
\`\`\`

Looks like a reorderable-list widget API mismatch (`onReorderItem` vs the now-required `onReorder`), likely from a dependency version bump.

## Why it matters
These are hard errors, so `flutter analyze` is non-clean on `master` and any build touching this file fails. Worth confirming against CI's pinned Flutter (the repo pins `beta` via `.fvmrc`; reproduced here with 3.38.1 stable) in case it's a local-toolchain artifact — but the API change should be resolved regardless.

## Fix
Update the call site at `accord_home_channels.dart:855-859` to the current widget API (`onReorder`), or pin the widget package to the version whose API matches.

krazyjakee:
Fixes #138.

## Problem
`lib/features/spaces/views/accord_home_channels.dart` called `ReorderableListView.builder` with `onReorderItem:`, which is not a parameter on Flutter's built-in `ReorderableListView` — it requires `onReorder:`. On stable this is two hard `flutter analyze` errors (`missing_required_argument` + `undefined_named_parameter`), so `flutter analyze` is non-clean on `master` and any build touching the file fails.

## Fix
- Renamed `onReorderItem:` → `onReorder:`.
- The two APIs differ in `newIndex` semantics: `onReorderItem` reported the target index in **post-removal** coordinates, while Flutter's `onReorder` reports it **pre-removal** (the dragged slot is still counted). Added the canonical `if (newIndex > oldIndex) newIndex -= 1;` at the top of `_onReorder` so the existing logic — which indexes into the post-removal list — stays correct and drag-**down** moves don't land one slot too low.
- Updated the now-inaccurate comment.

Verified clean with stable 3.38.1 (the toolchain in the issue):
```
Analyzing accord_home_channels.dart...
No issues found!
```

## Why it reached master, and the CI guard
The error landed in #107 and CI was **green** — because the analyze gate only ran the pinned **`beta`** channel, which tolerated `onReorderItem`. A channel-skew break sailed through.

This PR matrixes the `analyze` job over **`[beta, stable]`** (`fail-fast: false`), so a widget/SDK API that's valid on one channel but a hard error on the other now fails the merge gate. Cache keys include the channel so the legs don't clobber each other.

**Note:** this makes `stable` a hard blocker, including for releases (`release.yml` reuses this gate via `workflow_call`). The tree is currently clean on stable. If you'd prefer `stable` be advisory (report-only) until the port settles, that's a one-line `continue-on-error` change — shout.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -27,7 +27,9 @@ concurrency:
   cancel-in-progress: ${{ github.event_name == 'pull_request' }}
 
 env:
-  # The repo pins the Flutter channel in .fvmrc; keep this in sync with it.
+  # Channel used by the *build* job below. The *analyze* job now matrix-tests
+  # both beta and stable (see strategy.matrix.channel), so it ignores this.
+  # Keep in sync with .fvmrc and release.yml.
   FLUTTER_CHANNEL: beta
   # Workspace-local pub cache so actions/cache can persist it (incl. the
   # flutter_webrtc git checkout + its downloaded libwebrtc binary).
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -35,8 +35,19 @@ env:
 
 jobs:
   analyze:
-    name: Analyze & Test
+    name: Analyze & Test (${{ matrix.channel }})
     runs-on: ubuntu-latest
+    strategy:
+      # Run every channel even if one fails, so a stable-only break and a
+      # beta-only break are both reported in the same run.
+      fail-fast: false
+      matrix:
+        # Gate on both the pinned dev channel (.fvmrc) AND stable. A widget/SDK
+        # API that diverges between channels — e.g. ReorderableListView's
+        # onReorder vs onReorderItem (#138, #107) — is a hard error on one and
+        # silently tolerated on the other; analyzing a single channel lets that
+        # land on master. Catching it here is the cheap smoke test.
+        channel: [beta, stable]
     steps:
       - name: Checkout
         uses: actions/checkout@v4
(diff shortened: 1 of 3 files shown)


krazyjakee:
## Code Review — findings and fixes (commit `03041e5`)

### Severity ranking

#### 🟡 Medium — No tests for the index-fix (fixed)

The core bug fix (`if (newIndex > oldIndex) newIndex -= 1`) is the right canonical pattern, but it had zero test coverage. A future refactor that removes or misplaces the decrement would silently reintroduce the drag-down-lands-one-slot-too-low bug.

**Fixed:** Added 7 unit tests to `test/features/channels/channel_sort_test.dart` in a new `reorder index adjustment` group:
- drag-down by one slot (the "no-op" case — item stays in place)
- drag-down two slots
- drag-down to last position (Flutter passes `newIndex == itemCount`)
- drag-up by one slot, drag-up to first, drag-up multiple slots
- regression guard that documents the exact wrong list produced _without_ the decrement, so reviewers can see at a glance what the bug looked like

The helpers (`_adjustReorder` / `_applyReorder`) are pure functions defined locally in the test file — no production-code changes required.

---

#### 🔵 Low — `FLUTTER_CHANNEL` env var comment was misleading (fixed)

After the matrix change, the `analyze` job no longer reads `${{ env.FLUTTER_CHANNEL }}` — it reads `${{ matrix.channel }}` instead. The env var is still needed (the `build` job still uses it at line 182), but the top-level comment implied it controlled both jobs.

**Fixed:** Updated the comment to state explicitly that `FLUTTER_CHANNEL` is now only for the `build` job and that `analyze` uses the matrix.

---

#### ℹ️ Info — Pre-existing `_flatten` ordering inconsistency (not fixed, out of scope)

`_ChannelDragListState._flatten` places uncategorized channels **before** all categories; `_ChannelReorderState._flatten` (in `accord_channel_reorder.dart`) places them **after**. This inconsistency predates this PR and isn't introduced or worsened by it. Worth a follow-up to align the two, but it's out of scope here.

---

The PR's core changes (`onReorderItem` → `onReorder` + canonical decrement + matrix CI) are correct. No issues found with the logic itself.

---

PR #145, opened 2026-06-14
feat(updates): background download + one-click apply for self-updates

krazyjakee:
## Summary

Reworks self-updating so getting up to date takes a **single click**:

- **Background download** — as soon as a check finds a newer release, the matching build is downloaded and verified silently in the background (no banner while this runs).
- **Notification appears after the download finishes** — the update banner stays hidden until the build is staged, then shows *"Update ready — … Tap to restart & install."*
- **One click → restart** — tapping applies the already-staged build: desktop quits and the swap helper relaunches into the new version; Android hands the staged APK to the system installer. No intermediate page, no separate download step.

This replaces the previous multi-click flow (banner → Updates page → "Download & install" → wait → install).

## What changed

- Added `UpdatePhase.ready` plus `stagedArchivePath` / `preparedVersion` to `UpdateState`, with `updateReady` / `downloading` getters.
- Split the old `installUpdate()` into:
  - `prepareUpdate()` — background download + verify, staged at `ready`. Fired automatically by `check()`. Guards make it safe to call on every hourly check and re-stage when a newer release ships.
  - `applyUpdate()` — applies the staged build, or downloads-then-applies as a fallback.
  - Shared `_downloadAndVerify()` helper.
- `UpdateBanner` now stays hidden during the background download and only shows the one-click "Update ready" banner once staged. Platforms that can't self-install (web/iOS, or a failed background download) fall back to the old "tap to view" banner.
- Updates page button shows "Restart & install" (restart icon) when staged and reuses the staged download.

## Notes for reviewers

- **Web is unchanged** — still uses the service-worker reload path.
- No codegen needed (Riverpod provider signature unchanged).
- `flutter analyze lib/features/updates` is clean; existing `update_controller_test.dart` passes.
- The actual binary swap + restart couldn't be exercised locally (needs a real desktop build against a published GitHub release); verified via analyzer + unit tests only.

### DIFF
--- a/lib/features/updates/controllers/update_controller.dart
+++ b/lib/features/updates/controllers/update_controller.dart
@@ -16,8 +16,10 @@ part 'update_controller.g.dart';
 
 /// Where an in-place install currently is. Drives the update UI's progress and
 /// disabled states; [idle] also covers "not started" and "completed on Android"
-/// (where the system installer takes over and the app keeps running).
-enum UpdatePhase { idle, downloading, verifying, installing, failed }
+/// (where the system installer takes over and the app keeps running). [ready]
+/// means the new build has been downloaded and verified in the background and is
+/// staged for a one-click apply (restart-and-swap).
+enum UpdatePhase { idle, downloading, verifying, ready, installing, failed }
 
 /// Snapshot of the update checker.
 @immutable
@@ -31,6 +33,8 @@ class UpdateState {
     this.phase = UpdatePhase.idle,
     this.progress = 0.0,
     this.installError,
+    this.stagedArchivePath,
+    this.preparedVersion,
   });
 
   /// The latest release fetched, or null if none/unknown.
@@ -59,11 +63,27 @@ class UpdateState {
   /// The last install failure's user-facing message, or null.
   final String? installError;
 
+  /// Path to the downloaded-and-verified bundle staged for a one-click apply,
+  /// set when [phase] reaches [UpdatePhase.ready]. Null until then.
+  final String? stagedArchivePath;
+
+  /// The version currently being prepared or staged (download → ready), used to
+  /// dedupe repeated background prepares and to detect a stale staged archive
+  /// when a newer release ships. Null when nothing has been prepared.
+  final String? preparedVersion;
+
+  /// Whether a download/verify is actively in flight in the background.
+  bool get downloading =>
+      phase == UpdatePhase.downloading || phase == UpdatePhase.verifying;
+
   /// Whether an install is actively in flight (download/verify/swap).
   bool get installing =>
       phase == UpdatePhase.downloading ||
       phase == UpdatePhase.verifying ||
       phase == UpdatePhase.installing;
+
+  /// Whether a verified build is staged and one click away from being applied.
+  bool get updateReady => phase == UpdatePhase.ready && stagedArchivePath != null;
 
   /// Whether [latest] is a newer build than the running one. Pre-releases are
   /// ignored unless this build is itself a pre-release (matching the reference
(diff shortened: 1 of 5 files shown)


krazyjakee:
## Code Review — findings & fixes (commit e918008)

All issues below were fixed in the follow-up commit pushed to this branch.

---

### 🔴 High — Dead code: `UpdateState.downloading` getter

`bool get downloading` was defined in `UpdateState` (returned `true` during the `downloading` or `verifying` phase) but was never referenced anywhere in the codebase — not in the views, not in the controller, not in tests. It also overlaps almost completely with the existing `installing` getter (which covers the same two phases plus `installing`), making it a source of confusion about which one to call. **Removed.**

---

### 🟡 Medium — `copyWith` couldn't nil-out `stagedArchivePath` / `preparedVersion`

The `copyWith` pattern uses `value ?? this.value`, so passing `null` for either field was a no-op — once set, they could never be cleared. This is the same gap that `clearInstallError` exists to solve for `installError`.

**Impact:** After Android successfully hands the APK to the system installer, the code set `phase = idle` but left `stagedArchivePath` pointing at the old temp file. A subsequent hourly check then couldn't distinguish "already staged for this version" from "stale reference after install handed off", leading to an unnecessary re-download.

**Fix:** Added `bool clearStaged = false` to `copyWith`, and called `copyWith(phase: idle, clearStaged: true)` in `applyUpdate` after the Android system-installer hand-off.

---

### 🟡 Medium — `prepareUpdate` dedup guard was too narrow

The old guard:
```dart
if (state.phase == UpdatePhase.ready && state.preparedVersion == version) return;
```
only fired when `phase == ready`. After the `clearStaged` fix above, this is fine for the steady state. But the version-based check is more expressive and covers any future call ordering:

**Fix:** Simplified to:
```dart
if (state.preparedVersion == version && state.stagedArchivePath != null) return;
```
This is equivalent in the steady state and more readable about the intent ("we already have a staged archive for this version").

---

### 🟡 Medium — Missing tests for new `updateReady` getter and `clearStaged`

**Added to `update_controller_test.dart`:**
- `updateReady` is false in initial state
- `updateReady` is false when phase is `ready` but `stagedArchivePath` is null
- `updateReady` is false when `stagedArchivePath` is set but phase is not `ready`
- `updateReady` is true when both conditions hold
- `copyWith(clearStaged: true)` nils out both staged fields
- `copyWith` default preserves staged fields across phase transitions

**Added to `update_banner_test.dart`:**
- Shows "Update ready … Tap to restart" text when `updateReady == true`
- Collapses "Update ready" banner when `dismissedVersion` matches the release

---

### ✅ No issues found

- **File sizes:** all three changed files are well under 500 lines.
- **`update_banner.dart` logic:** the `showViewBanner` condition correctly guards on `updateAvailable` so non-newer releases (e.g. pre-release on stable) never surface. The removal of `!updateAvailable` from the early exit is safe because both downstream rendering paths are independently guarded.
- **`applyUpdate` fallback path:** correctly re-downloads when nothing is staged or when the staged version is stale relative to the current offer.
- **Progress indicator:** the Updates page `LinearProgressIndicator` still shows during background `prepareUpdate` runs (phase is `downloading`), which is intentional and fine — the banner stays hidden, but the page gives feedback if the user navigates there.
- **Web/iOS path:** unchanged; `prepareUpdate` returns early on platforms where `canInstallInPlace == false`.


---

krazyjakee:
Closing as superseded: master already contains the background-download + one-click self-update feature (landed via #93, commit 833d5169), including the `clearStagedArchive`/`clearPreparedVersion` copyWith flags this PR introduced. Rebasing onto master would discard this branch's commits in favor of the already-merged version, so closing rather than rebasing.

PR #147, opened 2026-06-15
fix(windows): silence experimental/coroutine deprecation (STL1011)

krazyjakee:
## Problem
The latest Release run failed at **CI / Build Windows**, which (being gated) skipped the **Google Play** deploy job and blocked the Play release.

The GitHub Windows runner image updated to a new MSVC toolchain (VS 18 / MSVC 14.51). That compiler promotes the `<experimental/coroutine>` deprecation to a **hard error** (`C2338` / `STL1011`). Three Flutter plugins still include that header and fail to compile:
- `audioplayers_windows`
- `permission_handler_windows`
- `flutter_local_notifications_windows`

This is an upstream toolchain regression, not a code change in this repo.

## Fix
Add `_SILENCE_EXPERIMENTAL_COROUTINE_DEPRECATION_WARNINGS` as a project-wide compile definition in `windows/CMakeLists.txt`, alongside the existing `UNICODE` definition, so it propagates to all targets and plugin subdirectories.

## Verification
Can't build Windows on the maintainer's macOS 13 machine; this is the standard MSVC-recommended remedy for STL1011. The Windows build leg of the Release run will confirm.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #148, opened 2026-06-15
ci(play): upload signed AAB artifact for manual first upload

krazyjakee:
## Why
The Google Play deploy job builds a valid signed AAB but the upload fails repeatedly with:

```
Google Api Error: Server error - Internal error encountered.
```

Two consecutive runs failed identically, the second failing even earlier (during "Preparing aab"). This is the well-known symptom of a **brand-new Play app with no first release** — the Play Developer API rejects uploads with an opaque 500 until the app's first release is created manually in the Console.

## Change
- Save the signed AAB as a workflow artifact (`daccord-playstore-aab`, `if: always()`) so we can grab the exact signed bundle and do the one-time manual upload via the Play Console. API deploys work after that.
- Guard the `_SILENCE_EXPERIMENTAL_COROUTINE_DEPRECATION_WARNINGS` define (from #147) behind `if(MSVC)`.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -701,6 +701,18 @@ jobs:
             --dart-define=APP_STORE=true \
             --dart-define=SENTRY_DSN="${SENTRY_DSN:-}"
 
+      - name: Upload signed AAB artifact
+        # Keep the exact signed bundle even if the Play API upload fails, so the
+        # first-ever release can be uploaded manually via the Play Console (a
+        # brand-new app rejects API uploads with an opaque 500 until its first
+        # release is created by hand).
+        if: always()
+        uses: actions/upload-artifact@v4
+        with:
+          name: daccord-playstore-aab
+          path: build/app/outputs/bundle/release/app-release.aab
+          if-no-files-found: warn
+
       - name: Upload to Google Play
         env:
           PLAY_JSON_KEY_PATH: ${{ runner.temp }}/play-service-account.json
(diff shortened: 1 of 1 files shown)


krazyjakee:
## Code Review — findings & fixes (ranked by severity)

### Medium — Commit message / PR description are inaccurate
The commit message ("guard MSVC coroutine define") and the PR description ("Guard the `_SILENCE_EXPERIMENTAL_COROUTINE_DEPRECATION_WARNINGS` define behind `if(MSVC)`") both claim that change is in this PR. It is **not** — `windows/CMakeLists.txt` is unchanged here. The `if(MSVC)` guard was already applied in the second squash commit of PR #147 (`62f45b148f5`). No code fix possible for an already-pushed commit message, but noting it to avoid confusion during future blame/history reads.

### Low — `retention-days` missing; defaults to 90 days (wasteful for a one-time artifact)
The AAB is only needed until the first manual Play Console release is created — after that, subsequent deploys go via the API and the artifact serves no purpose. Defaulting to 90 days accumulates unnecessary storage against the Actions quota.

**Fixed:** added `retention-days: 30`.

### Negligible — `if: always()` fires on workflow cancellation
`if: always()` includes the cancelled state, which triggers the artifact upload step even when a developer cancels the run mid-build. With `if-no-files-found: warn` the impact is just a spurious warning, but `if: success() || failure()` is semantically correct and matches the documented intent.

**Fixed:** replaced `if: always()` with `if: success() || failure()`.

---

Both fixes applied in commit `c482418` on this branch.

---

PR #152, opened 2026-06-16
Add regression tests for platformAsset selection logic

krazyjakee:
## Summary
Adds comprehensive test coverage for the `UpdateController.platformAsset()` selection logic to prevent a recurring bug where unextractable package formats (`.deb`, `.rpm`, `.appimage`, etc.) were incorrectly chosen for in-place installation, causing "Unsupported archive" crashes.

## Changes
- **Import `universal_platform`** to enable platform-specific test assertions
- **Add test helper functions:**
  - `setLatest()` — sets up a mock release with specified assets
  - `_asset()` — creates test asset objects
  - `_fullReleaseAssets` — realistic asset list mirroring a real multi-platform release (including both `.deb` and `.tar.gz` for Linux, in GitHub's alphabetical order)
  - `_hostInstallableExt` — getter that returns the only extractable extension the current host platform supports (`.apk` on Android, `.zip` on Windows, `.dmg` on macOS, `.tar.gz` on Linux, `null` on web/iOS)

- **Add regression test group** `'platformAsset selection (Unsupported archive regression)'` with five tests:
  1. **never hands an unextractable package to the in-place installer** — verifies that `.deb`, `.rpm`, `.appimage`, `-setup.exe`, `.msi`, `.pkg` are never selected
  2. **selects the extension the host installer can actually apply** — confirms the correct platform-specific format is chosen
  3. **prefers .tar.gz over .deb when the .deb is listed first** — reproduces the exact GitHub ordering bug and verifies the fix
  4. **a package-only release has no in-place asset but is downloadable** — ensures graceful fallback when only unextractable formats exist
  5. **platformAssetUrl prefers .tar.gz over .deb for manual download** — validates download URL selection respects priority
  6. **canInstallInPlace is true only when an extractable asset exists** — confirms the flag correctly reflects availability

## Implementation Details
- Tests are platform-aware: Linux-specific tests skip on other platforms; web/iOS tests handle the `null` case
- The `_fullReleaseAssets` list intentionally mirrors the real v0.2.4 release structure to catch the exact ordering trap that caused the original bug
- All tests use the existing `makeContainer()` and `controllerOf()` helpers for consistency with the test suite

https://claude.ai/code/session_01YMPGSUtRSXJExgsvLxPoCm

### DIFF
--- a/lib/features/updates/controllers/update_controller.dart
+++ b/lib/features/updates/controllers/update_controller.dart
@@ -256,9 +256,13 @@ class UpdateController extends _$UpdateController {
   /// macOS, and hands a `.apk` to the system installer on Android. Package
   /// formats the swap helper can't apply (`.deb`/`.rpm`/`.appimage`, setup
   /// `.exe`/`.msi`) are deliberately excluded — they're download-only.
+  ///
+  /// Windows uses the more-specific `windows-x86_64.zip` suffix before the
+  /// generic `.zip` so the web build bundle (`daccord-web.zip`) is never
+  /// mistakenly selected ahead of the actual Windows installer.
   List<String> get _installableExts {
     if (UniversalPlatform.isAndroid) return const ['.apk'];
-    if (UniversalPlatform.isWindows) return const ['.zip'];
+    if (UniversalPlatform.isWindows) return const ['windows-x86_64.zip', '.zip'];
     if (UniversalPlatform.isMacOS) return const ['.dmg'];
     if (UniversalPlatform.isLinux) return const ['.tar.gz'];
     return const [];
@@ -270,7 +274,9 @@ class UpdateController extends _$UpdateController {
   /// apply in place.
   List<String> get _downloadExts {
     if (UniversalPlatform.isAndroid) return const ['.apk'];
-    if (UniversalPlatform.isWindows) return const ['.zip', '.exe', '.msi'];
+    if (UniversalPlatform.isWindows) {
+      return const ['windows-x86_64.zip', '.zip', '-setup.exe', '.exe', '.msi'];
+    }
     if (UniversalPlatform.isMacOS) return const ['.dmg', '.pkg', '.zip'];
     if (UniversalPlatform.isLinux) {
       return const ['.tar.gz', '.deb', '.rpm', '.appimage'];
--- a/test/features/updates/update_controller_test.dart
+++ b/test/features/updates/update_controller_test.dart
@@ -1,3 +1,5 @@
+import 'dart:io';
+
 import 'package:bonfire/features/settings/controllers/settings.dart';
 import 'package:bonfire/features/updates/controllers/update_controller.dart';
 import 'package:bonfire/features/updates/models/app_release.dart';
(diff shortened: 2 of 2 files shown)


krazyjakee:
## Code Review — findings and fixes (commit ef464e1)

Reviewed `test/features/updates/update_controller_test.dart` and the production `UpdateController` it covers. Three issues found; all fixed and pushed to this branch.

---

### 🔴 HIGH — Windows `.zip` ambiguity produces same bug class as the regression being guarded

**Where:** `update_controller.dart` `_installableExts` / `_downloadExts`, plus the test's `_fullReleaseAssets` and `'selects the extension the host installer can actually apply'`.

`_fullReleaseAssets` correctly mirrors the real alphabetical GitHub asset order, which places `daccord-web.zip` **before** `daccord-windows-x86_64.zip`. On Windows, `_installableExts` returned `['.zip']`, so `_assetForExts` picked the first `.zip` match — `daccord-web.zip`, the web bundle — as the in-place Windows installer. This is exactly the same failure mode as the `.deb`-over-`.tar.gz` bug this PR guards against, just for Windows.

The test's assertion `asset!.name.endsWith('.zip')` passed for `daccord-web.zip`, masking the bug entirely.

**Fix (production):** `_installableExts` for Windows now returns `['windows-x86_64.zip', '.zip']` and `_downloadExts` returns `['windows-x86_64.zip', '.zip', '-setup.exe', '.exe', '.msi']` — the more-specific suffix is checked first, then the generic `.zip` fallback for future naming conventions.

**Fix (test):** Added a Windows regression test parallel to the Linux one:
```dart
test('prefers windows zip over web zip when web zip is listed first', () {
  if (!UniversalPlatform.isWindows) return;
  // daccord-web.zip sorts before daccord-windows-x86_64.zip alphabetically
  setLatest(c, [_asset('daccord-web.zip'), _asset('daccord-windows-x86_64.zip')]);
  expect(asset!.name, equals('daccord-windows-x86_64.zip'));
});
```
Also tightened the `'selects the extension the host installer can actually apply'` assertion on Windows to require the name contains `'windows'`, not just ends with `'.zip'`.

---

### 🟡 MEDIUM — `dart:io` import placed after `package:` imports

**Where:** `update_controller_test.dart` lines 1–9 (pre-fix).

Dart's `directives_ordering` lint requires `dart:` imports before `package:` imports. The `dart:io` import was placed last, after all `package:` imports (this was pre-existing in the file; the PR added the `universal_platform` import in the same wrong position).

**Fix:** Moved `dart:io` to line 1, with a blank line separating it from the `package:` block.

---

### 🟢 LOW — Platform-conditional tests use silent `return` instead of `markTestSkipped`

**Where:** All tests with `if (!UniversalPlatform.isLinux) return;` / `if (!UniversalPlatform.isWindows) return;`.

Using `return` makes the test report as **passed** on non-matching platforms rather than **skipped**. This is a common pattern in Flutter tests and not a bug, but it means CI on macOS/Windows counts these as passing without executing them. Left unfixed since changing to `markTestSkipped` would be a noise-only refactor and the pattern is consistent with the existing test suite style.

---

PR #176, opened 2026-06-25
Codebase audit: DRY/consistency cleanup of the view tier

krazyjakee:
## What this is

An organization/consistency audit of `lib/` followed by the fixes. The codebase was already strong on fundamentals (one Riverpod codegen style, single gateway-wiring hub, 100% package-absolute imports, snake_case throughout). The work here targets the two areas that had drifted: **duplicated view-tier boilerplate** and a few **inconsistent/incorrect conventions**.

Net result: **70 files, +271/−292** (a net reduction) with no behavior changes intended.

## Changes (by commit)

**1. Shared infrastructure + doc/dep drift** (`618bae9`)
- New `lib/shared/components/async_state_views.dart`: `LoadingView`, `InlineError`.
- `accordInitial()` helper (the `name.isEmpty ? '?' : name[0].toUpperCase()` idiom, ~24 copies).
- `AccordMemberAvatar` generalized — presence dot is now optional, so it covers the no-presence avatars too.
- `showInfoSnack()` companion to the existing `showErrorSnack()`.
- `messageTimestampString()` — kills the duplicated weekday/month tables in the message row.
- Moved the gateway dispatcher `accord_event_handler.dart` from `events/utils/` → `events/controllers/` (it's the gateway-wiring hub, not a stateless util).
- Logged the two previously-silent empty `catch` blocks in `space_cache.dart`.
- Dropped `freezed`/`freezed_annotation`/`json_serializable`/`dart_mappable_builder` — **zero usages** in `lib/`, no `build.yaml` references them.
- Corrected `CLAUDE.md`: client models are hand-rolled (not `json_serializable`); the per-server `AccordClient` is owned by `AccordAuth` (not `connections.dart`); documented the real gateway-dispatcher + controller data-access pattern.

**2. Apply the helpers across views** (`efe7500`)
- `LoadingView` replaces inline `Center(child: CircularProgressIndicator())` at ~28 sites.
- `accordInitial`, `showInfoSnack`, and `AccordMemberAvatar` applied where they cleanly fit.
- In-button/footer fixed-size spinners and avatars with bespoke initial styling were intentionally left as-is.

**3. Drop dead `last-guild-channels` Hive box** (`3c46225`)
- The box was opened at startup but never read/written anywhere — leftover Bonfire guild-channel memory and the last live Discord-vocabulary identifier in the code. No migration needed (nothing consumed it).

**4. Consolidate error arms into `InlineError`** (`b208895`)
- ~22 `Text(error, bodySmall + colors.red)` error arms → `InlineError`. Dead `colors` locals cleaned up. Only true error arms with the exact shape were converted (button foregroundColors, icon tints, bold danger labels, `colorScheme.error` left alone).

**5. Unify per-feature widget dirs under `views/`** (`b73de3a`)
- `channels/components/`, `messaging/components/` (incl. `box/`), `user/components/self_status_button.dart` → `views/`; removed empty `notifications/components/` and `voice/components/`. Pure move + import-path rewrite (lib + test).

## Deliberately deferred (rationale)

- **Reinstating a per-feature `repositories/` layer** — the audit flagged that data access lives in controllers/views rather than a repository layer. Building one out across 5 features is a rewrite, which `CLAUDE.md` explicitly steers away from ("a port, not a rewrite"). Instead, the docs were corrected to describe the actual architecture.
- **View-suffix unification (`_page`/`_view`/`_screen`) + class-name alignment + the half-applied `accord_` prefix** — high import/identifier churn for modest gain; better as a focused follow-up than bundled here.

## Verification note

This environment has no Dart/Flutter toolchain, so the changes were verified by static analysis (import resolution including `part of` libraries, no duplicate/orphaned imports, no dangling symbols) rather than `flutter analyze`/`flutter test`. CI will be the first full compile — I'll drive it to green.

---
_Generated by [Claude Code](https://claude.ai/code/session_01SpNEK8oWh7ryG1yX43QerF)_


krazyjakee:
## Code review findings

Reviewed the full diff (70 files, 5 commits). The PR is structurally solid — the helper extraction is correct, imports are consistent, and the `part of` files correctly inherit from `accord_home.dart`. Three issues found and fixed in commit `ee64d9c`.

---

### 🔴 High — Style regression in `connections_settings_page.dart`

**`lib/features/settings/views/connections_settings_page.dart`, `_body()` error branch**

The full-region error state previously used `bodyMedium` (larger readable text for an empty-screen case). When converted to `InlineError` it silently downgraded to `bodySmall`, making the error text noticeably smaller. The PR description explicitly says *"Only true error arms with the exact shape were converted"* — `bodyMedium` is not that shape.

**Fix:** Reverted to the original `Center(child: Text(..., bodyMedium.red))`. `LoadingView` in the same method was correct and kept.

---

### 🟡 Medium — Two `InlineError`-eligible arms missed in `updates_page.dart`

**`lib/features/updates/views/updates_page.dart`, lines ~80 and ~192**

Both `update.error` and `update.installError` display errors as `bodySmall.red` inside a `Padding` — exactly the `InlineError(_, centered: false)` shape that the PR targets, but they were not converted. Left as inconsistent manual inline patterns.

**Fix:** Converted both to `InlineError(…, centered: false)` and added the `async_state_views` import.

---

### 🟡 Medium — No tests for the three new pure functions

`accordInitial`, `messageTimestampString`, and `showInfoSnack` are all pure/side-effectful utilities with clear contracts, and the codebase already has test files for their modules (`rest_result_ext_test.dart`, `message_time_test.dart`). Edge cases like `null`, empty string, whitespace-only, and month/weekday indexing are the sort of thing that silently breaks on refactor.

**Fix:** Added tests covering:
- `accordInitial`: null, `""`, whitespace-only, leading whitespace trim, single char, lower→upper
- `messageTimestampString`: known date format, minute padding, all 12 month names, all 7 weekday names (new `test/features/member/member_display_test.dart` + extended `message_time_test.dart`)
- `showInfoSnack`: widget test verifying the snackbar text appears (extended `rest_result_ext_test.dart`)

---

### ℹ️ Note — `showInfoSnack` applied to a few error messages

`showInfoSnack(context, 'Failed to open direct message')` and `showInfoSnack(context, 'Incorrect PIN')` are error conditions, not informational confirmations. The function name is semantically misleading here. However, the original code also used a plain unstyled `SnackBar` for both, so there is **no visual regression** — both before and after show an identical plain snackbar. Left as-is; if the convention matters a rename or overload can follow as a separate cleanup.

---

No file over 1000 lines, no 3D/rendering changes, no breaking structural issues. CI will be the first full compile check.

---

PR #178, opened 2026-07-09
fix(updates): make Linux self-update work for .deb (system) installs

krazyjakee:
## The bug

On Linux, the self-updater silently no-ops for `.deb` installs. Symptom: the "Update ready — tap to restart & install" banner appears, clicking it quits the app, it comes back — **still on the old version**.

**Root cause:** the updater assumed a user-writable install root. The `.deb` installs to root-owned `/opt/daccord` (`/opt` is `root:root`), so the swap helper's first step —

```bash
mv "/opt/daccord" "/opt/daccord.old" 2>/dev/null   # Permission denied — silently swallowed
```

— fails, the `if` block is skipped, nothing swaps, and it relaunches the same old `/opt/daccord/daccord`. Download + verify succeed, so it *looks* like it worked. The same latent problem affects macOS `/Applications` and Windows `Program Files` installs.

## The fix

1. **Gate in-place installs on a writable install root.** A cheap, cached probe checks whether the install root's *parent* dir is writable (exactly what the swap `mv` needs). Package-manager installs no longer offer the broken one-click apply.

2. **Real privileged path for `.deb` installs (the common Linux case).** When the root isn't writable but the release ships a `.deb` and `pkexec` is present, the updater now:
   - downloads the new `.deb` and SHA-256-verifies it,
   - runs `pkexec dpkg -i <deb>` → polkit shows an admin-password dialog,
   - on success, dpkg overwrites `/opt/daccord` (safe while running — Linux keeps the old inode until exit), the app quits, and a detached helper relaunches the new binary.
   - On a cancelled prompt or dpkg failure it throws a clear message ("Update needs administrator approval." / the dpkg error) and **stays on the current version** — no silent false-success, and it doesn't quit.

3. **Honest fallback.** Without `pkexec` (or on a non-writable root generally), the UI falls back to a manual download, now preferring the `.deb` the user's package manager applies over the portable tarball.

## Changes

- **`update_installer_io.dart`** — `isInstallRootWritable` + `hasPrivilegedInstaller` probes (cached, with `@visibleForTesting` seams); `_installLinux` routes a `.deb` through the new `_installLinuxDeb` (pkexec + verified relaunch).
- **`update_controller.dart`** — `_installableExts` now encodes the writability / privilege gating for all desktop platforms; added `requiresPrivilegedInstall` so the UI can warn.
- **`update_banner.dart` / `updates_page.dart`** — admin-aware wording ("Tap to install (admin required)", "Install (admin)").
- **`update_installer_stub.dart`** — web API parity.
- **Tests** — cover the writable, non-writable-without-pkexec, and non-writable-with-pkexec paths.

## Reviewer notes

- **Rollout caveat:** this ships in a new release; an already-installed build has the old helper, so the *first* update after this lands still runs the old (broken) path. That first fixed `.deb` likely needs a one-time manual `dpkg -i`; every update after is one-click.
- `pkexec dpkg -i` needs an interactive polkit agent (desktop sessions provide one). No apt repo is involved — it reinstalls the downloaded package directly, keeping dpkg's DB consistent.
- The privileged flow was not exercised end-to-end in this change (it would really reinstall daccord + trigger a password prompt); logic is covered by unit tests and the mechanics were validated on a real `/opt` install.

All 458 tests pass; `flutter analyze` clean for the touched files.

### DIFF
--- a/lib/features/updates/controllers/update_controller.dart
+++ b/lib/features/updates/controllers/update_controller.dart
@@ -158,6 +158,9 @@ class UpdateController extends _$UpdateController {
     // App Store builds update through the store; never check GitHub or arm the
     // periodic timer (see [kAppStoreBuild]).
     if (kAppStoreBuild) return;
+    // The web build ships whatever the server serves — there's nothing to
+    // download or install, so never check for releases or show the banner.
+    if (UniversalPlatform.isWeb) return;
     _timer ??= Timer.periodic(_throttle, (_) {
       if (ref.read(settingsControllerProvider).autoUpdateCheck) check();
     });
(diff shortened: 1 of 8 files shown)


krazyjakee:
## Automated review

Reviewed the diff (`update_installer_io.dart`, `update_controller.dart`, `update_installer_stub.dart`, `update_banner.dart`, `updates_page.dart`, `update_controller_test.dart`) against the three focus areas below. No 3D/rendering code is touched, so that section doesn't apply.

### Findings

**1. Missing test coverage for the new admin-required UI (Medium — fixed in `4580c6d`)**
`requiresPrivilegedInstall` itself was unit-tested, but the UI it drives — `UpdateBanner`'s "Tap to install (admin required)." text and `UpdatesScreen`'s `_installLabel` "Install (admin)" / "Download & install (admin)" wording — had zero test coverage. Pushed:
- `test/features/updates/update_banner_test.dart`: new widget test asserting the admin-required banner text renders when `requiresPrivilegedInstall` is true (mirrors the existing "restart & install" test, extending `_FakeUpdateController` with an overridable getter).
- `test/features/updates/updates_page_test.dart` (new file): three widget tests covering the `Install (admin)`, `Restart & install`, and `Download & install (admin)` button labels on `UpdatesScreen`.

**2. `_probePkexec` only checks 3 hardcoded paths (Low — not fixed, flagging for awareness)**
`update_installer_io.dart`'s `_probePkexec()` checks `/usr/bin/pkexec`, `/bin/pkexec`, `/usr/local/bin/pkexec`. This covers Debian/Ubuntu (the stated target of this fix) but would miss `pkexec` installed elsewhere (e.g. NixOS's `/run/current-system/sw/bin/pkexec`). Since the privileged path is Linux-`.deb`-specific and Debian/Ubuntu ship pkexec at `/usr/bin/pkexec`, this is a minor robustness gap rather than a bug — a `PATH`-based lookup (e.g. `Process.runSync('which', ['pkexec'])`) would be more portable if that ever matters. Not changed, since it's outside the stated scope and the current behaviour is correct for the target platform.

### Everything else checked out
- Writability probe (`isInstallRootWritable`) correctly probes the *parent* of the install root, matching the `mv`/rename semantics the swap helper actually needs, and is consistent across the Windows/macOS/Linux swap scripts.
- `_installLinuxDeb`'s `pkexec` exit-code handling (126 = dialog dismissed, 127 = not authorized) matches pkexec's documented behavior, and failures correctly throw `UpdateInstallException` without quitting the app (verified against `applyUpdate`'s catch block — the app stays on the current version on failure).
- `update_installer_stub.dart` (web) keeps full API parity with the new `isInstallRootWritable`/`hasPrivilegedInstaller`/debug-override surface, so the conditional export doesn't break the web build.
- No file in the diff approaches the 1000-line split threshold.
- Was unable to run `flutter analyze`/`flutter test` locally in this environment (no Flutter SDK available), but CI (`Analyze & Test` stable + beta) passed on the original commit `27d5ad3`, and I pushed the new tests as a follow-up commit rather than amending it.

— Claude (automated review)

---

PR #177, opened 2026-07-03
docs: add Flutter complexity audit report

krazyjakee:
Prioritized audit of lib/ and pubspec.yaml for overengineering:
~21 unused dependencies, ~30 sites hand-rolling existing shared
helpers, duplicated channel-reorder and message-row implementations,
the accord_auth god repository, test-only voice_logic duplication,
and a phased, behavior-neutral refactoring plan.

Co-Authored-By: Claude Fable 5 <noreply@anthropic.com>
Claude-Session: https://claude.ai/code/session_01YVE6hihmi3ty9PheAJfqhV

PR #179, opened 2026-07-09
fix(router): stop bouncing sub-routes home when logged in

krazyjakee:
## Problem

While signed in, opening **Settings** made the UI twitch as if it were about to open, then instantly snap back. **Switch account** appeared to do nothing, and **Server administration** was affected the same way.

## Root cause

The logged-in redirect on the `/` route guarded on `state.matchedLocation`:

```dart
final location = state.matchedLocation;
if (location != '/' && location != '/login' && location != '/register') {
  return null;
}
// logged in → redirect to /spaces
```

The redirect is attached to the parent `/` route. For a **parent-route** redirect, go_router sets `state.matchedLocation` to that route's *own* matched segment — which is always `/` — not the full incoming location (confirmed in go_router's `RouteMatch.buildState`: `matchedLocation` = the route's own segment, `uri` = the full destination).

So navigating to `/settings`, `/switcher`, or `/admin` while logged in was seen as `/`, passed the guard, and got redirected straight to `/spaces` — the twitch, and the dead Switch-account button.

Regressed in #177, when the login redirect moved from a `build()` post-frame callback into a go_router route-level redirect.

## Fix

Guard on the full destination (`state.uri.path`) instead of the per-route `matchedLocation`, so only `/`, `/login`, and `/register` bounce a logged-in user home — sub-routes open normally. Added a comment documenting the `matchedLocation` vs `uri.path` gotcha to prevent regression.

## Testing

- `flutter analyze --no-fatal-infos lib/router/controller.dart` → No issues found.
- Verified `/settings`, `/switcher`, and `/admin` all reach their screens while logged in; `/`, `/login`, `/register` still redirect home.

### DIFF
--- a/lib/router/controller.dart
+++ b/lib/router/controller.dart
@@ -33,7 +33,12 @@ final rootNavigatorKey = GlobalKey<NavigatorState>();
 /// not the full incoming location. Using `matchedLocation` here would treat
 /// every sub-route (`/settings`, `/admin`, `/switcher`) as `/` and bounce it
 /// home while logged in.
-String? _redirectLoggedInToHome(BuildContext context, GoRouterState state) {
+///
+/// Public (not `_`-prefixed) and [visibleForTesting] so tests can exercise it
+/// against a real [GoRouter] instance without depending on the app's actual
+/// screen widgets.
+@visibleForTesting
+String? redirectLoggedInToHome(BuildContext context, GoRouterState state) {
   final location = state.uri.path;
   if (location != '/' && location != '/login' && location != '/register') {
     return null;
@@ -52,7 +57,7 @@ final routerController = GoRouter(
         routes: [
           GoRoute(
             path: '/',
-            redirect: _redirectLoggedInToHome,
+            redirect: redirectLoggedInToHome,
             builder: (context, state) => const AccordLoginScreen(),
             routes: [
               GoRoute(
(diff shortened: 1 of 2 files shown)


krazyjakee:
## Automated review

Reviewed against code quality, file size/structure, and test coverage (no 3D/rendering code touched, so no visual review needed). Verified locally with a full Flutter SDK (installed in the review sandbox): `dart run build_runner build`, `flutter analyze`, and `flutter test` (469 tests) all pass.

### Findings

1. **Test coverage (medium) — fixed.** The regression fixed here (parent-route redirect reading `state.matchedLocation` instead of `state.uri.path`) had no test guarding it, so it could silently reappear. Pushed a commit that:
   - Renames `_redirectLoggedInToHome` → `redirectLoggedInToHome` and marks it `@visibleForTesting`, so it's exercisable from a test without dragging in the app's real screens.
   - Adds `test/router/controller_test.dart`, which builds a real nested `GoRouter` (mirroring the app's route tree with placeholder screens) and asserts:
     - `/settings`, `/admin`, `/switcher` stay put while logged in (the actual regression).
     - `/`, `/login`, `/register` still redirect to `/spaces` while logged in.
     - `/` does not redirect while logged out.
   - I confirmed this test suite genuinely catches the regression: reverting the fix locally (`state.uri.path` → `state.matchedLocation`) fails exactly the 3 sub-route tests, as expected.

2. **Code quality — no issues.** The fix itself (one-line guard change) is correct and minimal; the added doc comment clearly explains the `matchedLocation` vs `uri.path` gotcha for future readers. Consistent with surrounding style.

3. **File size/structure — no action needed.** `lib/router/controller.dart` is 99 lines, well under the 1000-line threshold.

4. **3D/visual — not applicable.** This PR only touches router redirect logic; no 3D model or rendering code paths are affected.

No other issues found.


---

PR #180, opened 2026-07-10
fix(members/auth): stop roster spinning forever; sign out on expired token

krazyjakee:
## What & why

The member roster **loaded forever and never resolved**, even after an app restart. Investigation surfaced two independent problems behind that one symptom.

### 1. Roster had no failure path (`lib/features/member`)
`AccordMembersController._load` set `state` **only on success**, and accordkit's HTTP client has **no request timeout**. So a hung, `4xx`/`5xx`, or throwing `members.list` call left `state == null` — and since the provider is `@Riverpod(keepAlive: true)`, it never retried or recovered. The only error UI was gated on the *gateway* being unreachable, which a failed REST call doesn't trip → permanent spinner.

**Fix:** `_load` now wraps the fetch in a **20s timeout** with **bounded retry** (3 attempts, backoff), and on exhausted failure flips a new `membersLoadFailedProvider`. The roster then shows a **"Couldn't load members" + Retry** card (`ServerUnreachable` gained an optional `title`) instead of spinning; Retry re-runs the load. Genuinely-empty spaces still render an empty roster (no false error).

### 2. Expired token still rendered cached data (`lib/features/authentication` + `packages/accordkit`)
The deeper cause: a dead/expired token still rendered **cached spaces and channels** while every authenticated REST call `401`'d in the background — which is what made this look like a member-list-specific hang. accordkit exposes **no token-refresh endpoint**, so silent recovery isn't possible.

**Fix:** central 401 handling.
- `AccordRest` gains an `onUnauthorized` hook, fired on any `401`, threaded through `AccordClient`.
- `AccordAuth` wires each **live** connection to `_handleUnauthorized`, which signs the account out via `removeAccount` — switching to another live account if one exists, otherwise logging out to the **login screen**. A `_signingOut` guard collapses the burst of simultaneous 401s into a single sign-out.
- Login/register use throwaway unauthenticated clients (no `onUnauthorized`), so a wrong-password `401` never triggers a global sign-out.

## Verified live
Ran on Linux desktop against `chat.daccord.gg` with an expired session:

```
Session 312534760092073984@https://chat.daccord.gg is unauthorized (token invalid/expired); signing it out.
```

Sign-out fired exactly **once** (guard held), and the app landed on the login screen instead of a broken home. Roster retry state renders in place of the infinite spinner.

## Reviewer notes
- No behavioural change for healthy sessions — `onUnauthorized` only fires on real `401`s.
- Multi-account: if a background account's token dies, only that connection is dropped; the active session is untouched.
- `dart analyze` (accordkit) and `flutter analyze --no-fatal-infos` on all touched files are clean. `accord_members.g.dart` is regenerated Riverpod codegen for the new provider.

### DIFF
--- a/lib/features/authentication/repositories/accord_auth.dart
+++ b/lib/features/authentication/repositories/accord_auth.dart
@@ -63,6 +63,11 @@ class AccordAuth extends _$AccordAuth {
   final Map<String, _Conn> _connections = {};
   String? _activeKey;
 
+  /// Connection keys currently being torn down after a 401, so the burst of
+  /// simultaneous unauthorized responses (spaces, members, emojis, …) triggers
+  /// exactly one sign-out per connection.
+  final Set<String> _signingOut = {};
+
   /// The active connection's client, or null when signed out. Repositories
   /// should obtain this via the logged-in [AccordAuthLoggedIn] state rather than
   /// reaching in here.
@@ -312,6 +317,30 @@ class AccordAuth extends _$AccordAuth {
         .set(ConnectionStatus.disconnected);
     ref.read(readyControllerProvider.notifier).setReady(false);
     state = const AccordAuthLoggedOut();
+  }
+
+  /// Reacts to a `401 Unauthorized` from connection [key]: its token is invalid
+  /// or expired and accordkit exposes no refresh path, so the session can't be
+  /// recovered silently. Sign the affected account out via [removeAccount],
+  /// which switches to another live server if one exists or otherwise
+  /// [logout]s to the login screen — matching "log me back in" rather than
+  /// leaving cached spaces/channels rendered behind a dead token.
+  ///
+  /// Guarded by [_signingOut] so the burst of simultaneous 401s (spaces,
+  /// members, emojis, …) tears the connection down exactly once; a no-op once
+  /// the connection is already gone.
+  Future<void> _handleUnauthorized(String key) async {
+    if (_signingOut.contains(key)) return;
+    final conn = _connections[key];
+    if (conn == null) return;
+    _signingOut.add(key);
+    debugPrint('Session $key is unauthorized (token invalid/expired); '
+        'signing it out.');
+    try {
+      await removeAccount(conn.session);
+    } finally {
+      _signingOut.remove(key);
+    }
   }
 
   // ── Add-a-server (multi-connection) ────────────────────────────────────────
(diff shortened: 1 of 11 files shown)


krazyjakee:
## Review: code quality, structure, test coverage

Scope covered: `accord_auth.dart`'s `onUnauthorized`/`_handleUnauthorized` wiring, `accord_members.dart`'s retry/timeout/failure-flag `_load`, the `ServerUnreachable` title param, and the `accordkit` REST changes. No file in this diff exceeds 1000 lines, so no split was needed, and there are no 3D/rendering changes in scope.

### Findings & fixes (most → least significant)

1. **Missing test coverage for all of the new runtime behavior (medium, fixed).** Nothing in the diff exercised the retry/backoff/failure-flag logic, the new `onUnauthorized` hook, or the `ServerUnreachable` title override — exactly the kind of risky, easy-to-regress code path this review is meant to catch (the retry loop in particular: three attempts, two different backoffs, a 20s timeout, and a shared `Riverpod` flag that a later refactor could easily break silently). Added:
   - `packages/accordkit/test/rest_test.dart`: `onUnauthorized` fires on 401 for both `makeRequest` and `makeMultipartRequest`, and stays silent on other statuses/success. (`test_helpers.dart`'s `mockRest` gained an `onUnauthorized` passthrough to make this testable, per the package's own testability conventions.)
   - `test/features/member/accord_members_load_test.dart`: drives `AccordMembersController._load` through a real `AccordClient` backed by `http`'s `MockClient` — covers a genuinely-empty roster (not a failure), retry-then-succeed, retry-exhausted (sets the failure flag), and a fresh load after Retry clearing it. This runs the actual `Future.delayed` backoffs rather than stubbing them, so the retry timing itself is under test.
   - `test/shared/server_unreachable_test.dart`: default vs. custom title/message, and the Retry button's wiring.

   All new tests pass, and the existing suite is unaffected: accordkit `dart analyze` (clean) + `dart test` (144 passed), and `flutter analyze --no-fatal-infos` (clean, no new issues) + `flutter test` (477 passed, incl. the 8 new tests) — verified locally on Flutter 3.46.0-0.3.pre (beta), matching `.fvmrc`.

2. **Doc comments overstated when the failure flag resets (low, fixed).** `MembersLoadFailed`'s doc said it's "Reset to false whenever a load (re)starts," and the roster's comment said retrying "clears the failed flag on the way in" — but `_load` only clears it on a *successful* load; the actual reset-before-retry happens in the Retry button's own `.set(false)` call in `accord_member_list.dart`. Not a functional bug (the two call sites already agree in practice), but the comment could mislead a future maintainer into removing the view-side reset as "redundant," which would actually change behavior. Reworded both comments to state where the reset really happens.

Nothing else stood out as a correctness, naming, duplication, or error-handling issue — the retry/backoff/timeout logic, the `_signingOut` de-dup guard in `AccordAuth._handleUnauthorized`, and the `onUnauthorized` threading through `accordkit` all read as sound and consistent with the surrounding code's conventions.

Pushed as a follow-up commit on this branch (`a6637eb`).


---

PR #181, opened 2026-07-15
fix: member roster loading, session expiry, and message long-press menu

krazyjakee:
## What changed

This branch bundles three related mobile/UX reliability fixes:

### 1. Message long-press menu (this session)
On Android, long-pressing a message body did nothing — only long-pressing the **avatar** opened the actions/reaction menu. Users couldn't add a reaction from the message itself.

**Cause:** the message row's `GestureDetector` used the default `HitTestBehavior.deferToChild`, which only fires when a child absorbs the touch. On mobile, plain-text content renders as a bare non-selectable `RichText` (no gesture recognisers), and the wrapping `Container` is transparent — so touches over the text hit nothing. The avatar worked only because `AccordMemberAvatar` paints a background color and is therefore hit-testable.

**Fix:** set `HitTestBehavior.opaque` on the message/thread/forum `GestureDetector`s so the whole row claims long-press and secondary-tap (right-click). Scrolling and mention/link taps are unaffected — vertical drag isn't claimed, and the row has no `onTap` outside bulk-select mode, so quick taps still reach link/mention recognisers.

Files:
- `lib/features/messaging/views/message_pane/message_row.dart`
- `lib/features/messaging/views/thread_view.dart`
- `lib/features/messaging/views/forum_view.dart` (already mostly worked via its `InkWell`; made explicit for consistency)

### 3. Sign out on expired token
(from the earlier `fix(members/auth)` commit on this branch.)

## Testing
- `flutter analyze --no-fatal-infos` — no issues on the changed files.
- Long-press now opens **Add reaction / Reply / …** anywhere on a message, thread reply, and forum card (verify on a device with `flutter run --flavor github`).

### DIFF
--- a/lib/features/messaging/views/forum_view.dart
+++ b/lib/features/messaging/views/forum_view.dart
@@ -377,6 +377,9 @@ class _PostRow extends ConsumerWidget {
       clipBehavior: Clip.antiAlias,
       margin: EdgeInsets.zero,
       child: GestureDetector(
+        // Opaque so the whole card is hit-testable for long-press, matching
+        // the message pane and thread rows.
+        behavior: HitTestBehavior.opaque,
         onLongPressStart: (d) => onMenu(d.globalPosition),
         onSecondaryTapUp: (d) => onMenu(d.globalPosition),
         child: InkWell(
--- a/lib/features/messaging/views/message_pane/message_row.dart
+++ b/lib/features/messaging/views/message_pane/message_row.dart
@@ -260,6 +260,12 @@ class _MessageRowState extends ConsumerState<_MessageRow> {
       onEnter: (_) => setState(() => _hovered = true),
       onExit: (_) => setState(() => _hovered = false),
       child: GestureDetector(
+        // Opaque so the whole row is hit-testable: on mobile a plain-text
+        // message renders as a bare RichText with no recognisers, and the
+        // surrounding Container is transparent, so with the default
+        // `deferToChild` a long-press over the text hits nothing and never
+        // fires. Only the avatar (which paints a background) responded before.
+        behavior: HitTestBehavior.opaque,
         onLongPressStart:
             widget.selecting ? null : (d) => _showActionsMenu(d.globalPosition),
         onSecondaryTapUp:
--- a/lib/features/messaging/views/thread_view.dart
+++ b/lib/features/messaging/views/thread_view.dart
@@ -496,6 +496,10 @@ class _MessageLineState extends ConsumerState<_MessageLine> {
       onEnter: (_) => setState(() => _hovered = true),
       onExit: (_) => setState(() => _hovered = false),
       child: GestureDetector(
+        // Opaque so a long-press anywhere on the row (not just the avatar)
+        // opens the menu; plain-text content is a transparent RichText that
+        // wouldn't register a hit under the default `deferToChild`.
+        behavior: HitTestBehavior.opaque,
         onLongPressStart: (d) => _showMenu(d.globalPosition),
         onSecondaryTapUp: (d) => _showMenu(d.globalPosition),
         child: Padding(
--- a/packages/accordkit/lib/src/core/accord_config.dart
+++ b/packages/accordkit/lib/src/core/accord_config.dart
@@ -13,6 +13,17 @@ class AccordConfig {
 
   static const int heartbeatIntervalDefault = 45000;
 
+  /// Upper bound applied to the server-advertised `heartbeat_interval`.
+  ///
+  /// The heartbeat doubles as the only keepalive traffic on an otherwise-idle
+  /// gateway socket. Many network intermediaries (home NATs, VPNs, corporate
+  /// firewalls, some CDN edges) silently drop idle WebSockets after ~30s — and
+  /// the server's default 45s interval is longer than that, so the socket is
+  /// culled (close 1006) before the first heartbeat ever fires, causing an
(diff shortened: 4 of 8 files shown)


krazyjakee:
## Review

Reviewed for code quality, file size, and test coverage (this PR doesn't touch 3D rendering, so no visual-review pass applies). Scope: the 401-triggered sign-out plumbing (`accordkit`/`accord_auth.dart`), the members-roster retry/failure UI, and the `HitTestBehavior.opaque` long-press fixes.

### Findings & fixes (pushed in `63b7a7c`)

1. **[Correctness — fixed] `AccordRest.makeRawRequest` didn't fire `onUnauthorized` on a 401**, unlike `makeRequest` and `makeMultipartRequest`. `makeRawRequest` is used for the plugin source/bundle download endpoints, which are token-authenticated — a 401 there (expired/revoked token) would silently fail instead of tripping `AccordAuth._handleUnauthorized` and signing the account out like every other endpoint does. Added the same `if (response.statusCode == 401) onUnauthorized?.call();` check used by the other two methods.

2. **[Test coverage — fixed] No tests existed for the new `onUnauthorized` callback.** Added regression tests in `packages/accordkit/test/rest_test.dart` covering all three request methods (`makeRequest`, `makeMultipartRequest`, and the now-fixed `makeRawRequest`), plus a negative test confirming a non-401 failure (403) doesn't fire the callback. Extended `mockRest()` in `test_helpers.dart` with an optional `onUnauthorized` param to support this.

### Reviewed, no change made

- **`_handleUnauthorized` calling `removeAccount`** fully removes the saved account (not just a soft logout) — confirmed this reuses the same `removeAccount` already wired to the explicit "sign out"/"remove account" UI actions elsewhere (`switcher.dart`, `accord_account_settings.dart`), so the behavior is consistent with existing intentional semantics, not novel. The `_signingOut` guard correctly dedupes a burst of simultaneous 401s (spaces/members/emojis/…) into a single teardown per connection — verified the synchronous check-then-add happens entirely before the first `await`, so there's no interleaving race even with several REST calls failing in the same event-loop tick.
- **Members-controller retry loop** (`accord_members.dart`): the 3-attempt/20s-timeout/1s-2s-backoff retry can, in the worst case, leave a background `_load()` invocation still in-flight when a manual retry (`ref.invalidate`) starts a fresh one — a stale-write-after-dispose race. This pre-exists the PR (a single slow request could already race a manual retry before this change) and the PR doesn't add a new failure mode, just a longer window; the codebase doesn't use `ref.mounted` guards anywhere else either, so a fix here would be inconsistent with current conventions. Flagging for awareness rather than blocking.
- **File sizes**: all touched files are well under the 1000-line threshold (largest is `message_row.dart` at 863 lines); no splitting needed.
- **`HitTestBehavior.opaque` changes** (message row / thread row / forum card): verified `onTap` stays `null` outside bulk-select mode and no drag handlers are registered, so scrolling and mention/link tap recognizers inside the row are unaffected — opaque only changes whether the row claims hits over its own transparent regions, not gesture-arena priority against descendant recognizers.
- **`accord_members.g.dart`** codegen output for the new `MembersLoadFailed` family matches the existing `AccordMembersController` family's generated shape; couldn't run `build_runner`/`dart test` in this environment (no Dart/Flutter SDK installed), so this is a structural read rather than a generated-and-diffed verification — worth a `flutter analyze`/`flutter test` pass in CI to confirm.

### Needs more context (not blocking)

- No test harness exists in this repo for Riverpod controllers wired to a live `AccordClient` (the members retry logic and `AccordAuth`'s 401→sign-out path both fall in this category). Testing those end-to-end would need a fake/mock `AccordClient` injection point that doesn't currently exist — out of scope to build for this PR, but worth tracking as a gap if this class of bug (silent-hang / silent-auth-failure) recurs.

---

krazyjakee:
## CI status

- **Analyze & Test (stable)** — ✅ passing (analyze + test, including the new `onUnauthorized` regression tests).
- **Analyze & Test (beta)** — ❌ failing, but **not because of anything in this PR**.

### What's failing

5 analyzer errors, all in `packages/livekit_client/lib/src/participant/local.dart` (lines 380, 385, 390, 399, 403) — `unchecked_use_of_nullable_value` on `publishOptions.videoCodec` / `publishOptions.degradationPreference`. That file:
- isn't touched by this PR (last changed in `#100`, months ago),
- is part of the vendored `livekit_client` fork that backs voice/video/screen-share (flagged in `CLAUDE.md` as actively maintained, not to be stubbed),
- fails only on the **beta** Flutter channel — the **stable** job analyzes the same file cleanly.

This matches exactly the scenario `ci.yml`'s own comment calls out: beta and stable are run in parallel specifically to catch "a widget/SDK API that diverges between channels... silently tolerated on the other." The beta channel's analyzer has apparently tightened its null-flow inference around a `publishOptions!`-then-bare-`publishOptions.field` pattern that stable's analyzer still accepts. Since `publishOptions` is asserted non-null once (`publishOptions!`) earlier in the same function but then accessed bare afterward, the real fix is either re-deriving a promoted local (`final opts = publishOptions!;` used throughout) or adding `!`/`?.` at each flagged site — but that's a judgment call about WebRTC track-publishing code I'd rather not make unilaterally on a PR that isn't about voice at all.

### Recommendation

I'd treat this as a separate, pre-existing issue rather than block this PR on it — happy to open a follow-up fix for `local.dart` if wanted, or take a stab at it here if you'd rather bundle it in. Not making that change without a nod first since it touches the voice/video publish path.

---

PR #182, opened 2026-07-29
fix(router): stop nesting the signed-in screens under the sign-in route

krazyjakee:
## The report

> When I click the "settings" button in the bottom left of the app it does nothing other than seemingly refresh the app. It is doing this to me on two different computers.
> — `#bug-reports`, 2026-07-28

## What's going on

That symptom is #179: the logged-in redirect lived on `/`, and `/spaces`, `/settings`, `/admin`, `/switcher` were all **children** of `/`, so a parent-route redirect saw `state.matchedLocation` pinned to `/` and bounced every one of them home. Settings twitched open and closed — which reads exactly like "refreshed the app".

The guard fix landed in #179 on **2026-07-09 22:12 UTC**. `v0.2.6` was cut **2026-07-09 11:21 UTC** — about eleven hours earlier. So the reporter is on the newest release and it does not contain the fix. **Cutting a release ships the user-visible fix on its own**; this PR removes the thing that made the bug possible so it can't come back a third time (it was introduced by #177 and had already survived one round of review).

## What this changes

- **`/spaces`, `/settings`, `/admin`, `/switcher` are now siblings of `/`, `/login`, `/register`**, not children. Each sign-in route carries `redirectLoggedInToHome` itself, so a signed-in destination can never inherit it. The guard on `state.uri.path` stays as belt-and-braces if the routes are ever nested again.
- **The sign-in screen is no longer mounted in the navigator stack underneath the whole app.** That nesting is also what made Android's back swipe on home pop down to sign-in (#125) and what let `AccordLoginScreen`'s auth listener call `go('/spaces')` from under a screen the user was actually looking at.
- **Screens opened from home use `push` consistently** — Settings already did; Switch account and Server administration now match. `/spaces` stays beneath them, so system back and the app bar's arrow (`canPop() ? pop() : go('/spaces')`) return to home instead of unwinding to the sign-in route. `go` is left for the cases that genuinely replace the stack: signing in, signing out, returning home.
- **The switcher's Back button pops when it can**, falling back to `/spaces` for a cold deep link — same shape as Settings and Admin. Previously it always jumped to `/spaces`, which was wrong when it was opened from the signed-out sign-in screen.

## Tests

`test/router/controller_test.dart` mirrors the new route shape and keeps all of #179's redirect cases. Added coverage for the path the bug report actually walks, which #179's test never touched — it only exercised `go`, and the gear button is the one `push` in the app:

- `push('/settings')` from `/spaces` opens Settings, then pops back to home.
- Same for `/switcher` and `/admin`.
- At `/spaces`, the sign-in screen is not in the tree and `canPop()` is `false` — this fails on the old nesting and is the direct regression test for #125.

## Verification

`flutter analyze` / `flutter test` were not run locally — no Flutter SDK in this environment — so CI is the gate on this one.

## Not fixed here

"Add account" in the switcher does `go('/login')`, which the logged-in redirect bounces to `/spaces`, so it does nothing while signed in. Pre-existing, same family of bug, but it needs an intentional "let me reach sign-in while signed in" signal rather than a routing tweak. Worth its own issue.

---
_Generated by [Claude Code](https://claude.ai/code/session_018mHj4GybdoXLJNizb2g7vu)_

### DIFF
--- a/lib/features/authentication/views/accord_login.dart
+++ b/lib/features/authentication/views/accord_login.dart
@@ -266,11 +266,12 @@ class _AccordLoginScreenState extends ConsumerState<AccordLoginScreen> {
   Widget build(BuildContext context) {
     final state = ref.watch(accordAuthProvider);
 
-    // This only covers a login that completes *while this screen is showing*.
-    // Landing on a login route while *already* logged in (e.g. tapping "back"
-    // out of /admin or /settings, whose router parent is this screen) never
-    // reaches build: the router redirects it straight home (see
-    // `_redirectLoggedInToHome` in `lib/router/controller.dart`).
+    // Covers a login that completes *while this screen is showing* — a state
+    // change doesn't re-run router redirects. Landing on a sign-in route while
+    // *already* logged in never reaches build: the router redirects it straight
+    // home (see `redirectLoggedInToHome` in `lib/router/controller.dart`), and
+    // the signed-in screens are siblings of these routes, so this screen is not
+    // mounted underneath them.
     ref.listen(accordAuthProvider, (previous, next) {
       if (next is AccordAuthLoggedIn) {
         final spaceId = _pendingJoinSpaceId;
@@ -322,7 +323,9 @@ class _AccordLoginScreenState extends ConsumerState<AccordLoginScreen> {
           onBrowse: () => setState(() => _view = _LoggedOutView.browse),
           onManualConnect: () =>
               setState(() => _view = _LoggedOutView.credentials),
-          onSwitchAccount: _hasAccounts ? () => context.go('/switcher') : null,
+          onSwitchAccount: _hasAccounts
+              ? () => context.push('/switcher')
+              : null,
         ),
         maxWidth: 480,
       ),
@@ -344,7 +347,7 @@ class _AccordLoginScreenState extends ConsumerState<AccordLoginScreen> {
           tosAccepted: _tosAccepted,
           onBack: _atFlowRoot ? null : _goBack,
           onModeChanged: _onModeChanged,
-          onSwitchAccount: () => context.go('/switcher'),
+          onSwitchAccount: () => context.push('/switcher'),
           onGeneratePassword: _generatePassword,
           onTosChanged: (v) => setState(() => _tosAccepted = v),
           onTosLinkTap: _openTos,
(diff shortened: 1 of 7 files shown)


krazyjakee:
## Review

Focused on code quality, file size, and test coverage for this diff (`lib/router/controller.dart` + the four screens it touches + `test/router/controller_test.dart`). No 3D/rendering changes here, so that section doesn't apply.

### Findings, ranked by severity

**1. (Medium — fixed, pushed) CI red on "Analyze & Test (beta)"**
The beta-channel analyze job was failing with 5 `unchecked_use_of_nullable_value` errors in `packages/livekit_client/lib/src/participant/local.dart` (`_publishVideoTrack`'s `negotiate()` closure). This is **not caused by this PR's diff** — that file isn't touched by any commit here, and every other PR run in the last few weeks (07-09 through 07-18) was green on both channels. The likely cause: the `beta` channel is an unpinned moving target (this run resolved to `beta-3.47.0-0.2.pre`), and a newer analyzer's flow analysis stopped promoting `publishOptions` to non-null inside the closure (it's reassigned in the enclosing scope after the closure captures it, so promotion doesn't apply there — one call site already worked around this with `publishOptions!`, five others didn't).

Fixed in `1e3c6cd`: capture a single non-null `final options = publishOptions!;` at the top of the closure and use that instead of asserting at each of the 5 call sites. Pushed directly to this branch since it was blocking merge and unrelated to the diff under review.

**2. (Low — informational, not changed) "Add account" in the switcher is a no-op while signed in**
`switcher.dart`'s "Add account" button calls `context.go('/login')`, which `redirectLoggedInToHome` immediately bounces back to `/spaces` for a logged-in user. This is pre-existing (unchanged by this PR) and already called out in the PR description under "Not fixed here" as needing its own issue — leaving it as-is, just confirming the analysis is correct and scope is intentional.

### What looks good
- The route-flattening (`/spaces`, `/settings`, `/admin`, `/switcher` as siblings of `/`, `/login`, `/register` instead of children) is the right fix for both #179 and #125 — verified the redirect guard (`state.uri.path`, not `matchedLocation`) and the push/pop consistency across all four call sites (`accord_login.dart`, `switcher.dart`, `accord_settings_screen.dart`, `accord_home.dart`).
- Checked for any leftover relative-path navigation (`.go('login')`, `.goNamed`, etc.) that the old nested structure would have allowed — none found; everything already used absolute paths, so the flattening is a clean drop-in.
- `test/router/controller_test.dart` mirrors the new route shape and adds direct regression coverage for both bugs (push `/settings`/`/switcher`/`/admin` from home and pop back, plus "home does not stack on the sign-in screen" for #125) — this is good, targeted coverage; nothing changed-behavior-wise looked untested.
- No file in the diff is anywhere near the 1000-line split threshold.

No changes needed to the PR's own diff — the router/screen changes are correct and well-tested as submitted.


---

PR #184, opened 2026-08-04
fix(messaging): surface attachment failures instead of failing silently

krazyjakee:
## The complaint

Users reported that they "can't attach MP3s" and assumed a permission was blocking them. Nothing in the client gates attachments by type or by `attach_files` — `attachments.dart` already maps `mp3` → `audio/mpeg`, and the multipart upload path is sound. What was actually happening is that **every way this flow could fail, failed in silence**, which is indistinguishable from being blocked.

## Three silent failure points

1. **Unreadable files vanished.** `_pickFiles` kept any picked file with bytes and dropped the rest with no message:
   ```dart
   if (file.bytes != null) _attachments.add(file);
   ```
   A cloud- or provider-backed file the platform can't read into memory produced no chip and no explanation.

2. **The 25 MB limit was documentation-only.** `docs/messaging/file-sharing.md` documented it; nothing enforced it. Music routinely crosses it where the images this flow was built around never did — a 320 kbps MP3 passes 25 MB at about ten minutes. The whole file was read, base64'd into a multipart body and uploaded in full before the server's rejection came back.

3. **`_send` discarded its result.** There was no `else` on `if (ok)`, no `SnackBar`, and no error state anywhere in `composer.dart`. `sendWithAttachments` logged to `debugPrint` and returned `false`. A 413, a 403 for missing `attach_files`, and a rejected MIME type all presented identically: the chip stayed put and the message didn't send.

## Changes

- **`lib/features/messaging/utils/attachment_limits.dart`** (new) — `kMaxAttachmentBytes`, `formatFileSize`, and `screenAttachments`, which splits picked files into accepted and rejected-with-a-reason. Pure logic in `utils/` so it's unit-testable.
- **`composer.dart`** — `_pickFiles` and both paste paths route through a shared `_addFiles` that screens and reports. New `_error` state renders as a dismissible `InlineError` above the composer, cleared on the next attach or send.
- **`accord_messages.dart`** — `sendWithAttachments` returns `String?` (null on success) instead of `bool`, passing the server's own reason up. Its one caller is the composer.
- **`rest_result_ext.dart`** — adds `errorMessageOr`, which renders the bare server message. The existing `errorOr` renders `AccordError.toString()`, i.e. `AccordError(code: …, message: …)` — right for an admin screen, noise in the composer. The ~50 existing `errorOr` call sites are untouched.
- **`docs/messaging/file-sharing.md`** — drops the drag-and-drop claim. There is no `DropTarget`, `DropRegion`, or `desktop_drop` usage anywhere in `lib/`, so dragging a file onto the composer was a second silent route to the same complaint. Documents what the limit now does.

Plain-text send failures were silent for the same reason, so they now show a generic message too — the alternative was leaving one branch of `_send` mute.

## Tests

- `test/features/messaging/attachment_limits_test.dart` (new) — size formatting, the accept/reject split, both rejection reasons, and that the constant matches the documented 25 MB.
- `test/shared/rest_result_ext_test.dart` — three cases for `errorMessageOr`, including that the `AccordError(` wrapper doesn't leak into user-facing text.

**I could not run `flutter analyze` or `flutter test` — no Flutter SDK in this container.** Changes were verified by inspection only; CI is the real check.

## What this does not fix

This makes the true cause visible; it doesn't raise the ceiling. If your deployment's server-side cap really is 25 MB, large music files still won't upload — they'll now just say why. Raising the cap is an `accordserver` change, outside this repo. **If your limit differs from 25 MB, `kMaxAttachmentBytes` needs updating to match** — I hardcoded the documented figure since I had no way to query a live server.

The memory profile is also unchanged: `pickFiles(withData: true)` still loads every file eagerly, and `form.build()` plus `bodyBytes` copy it twice more. Screening now rejects oversize files before the upload, but a large file is still fully resident before it's rejected. Streaming uploads would be a separate change.

https://claude.ai/code/session_01KejDtFqkZEdSUJz5H3diNM

---
_Generated by [Claude Code](https://claude.ai/code/session_01KejDtFqkZEdSUJz5H3diNM)_

### DIFF
--- a/docs/messaging/file-sharing.md
+++ b/docs/messaging/file-sharing.md
@@ -11,9 +11,11 @@ You can share files and images directly in any text channel or DM.
 
 ## Uploading Files
 
-Attach a file to your message using the attachment button in the composer. You can also **drag and drop** files into the message area.
+Attach a file to your message using the attachment button in the composer. Drag and drop is not supported yet.
 
-The maximum file size is **25 MB**. Files larger than this will show an error and will not be uploaded.
+The maximum file size is **25 MB** per file. Larger files are rejected as soon as you pick them, with the file's size shown next to the limit -- nothing is uploaded. Music and video files are the ones that most often run into this; a 320 kbps MP3 passes 25 MB at around ten minutes.
+
+If the server refuses an upload for another reason -- you don't have the **Attach Files** permission in that channel, for example -- the reason it gives is shown above the composer.
 
 ## Pasting Images
 
--- a/lib/features/messaging/controllers/accord_messages.dart
+++ b/lib/features/messaging/controllers/accord_messages.dart
@@ -2,6 +2,7 @@ import 'package:accordkit/accordkit.dart';
 import 'package:bonfire/features/messaging/utils/emoji_catalog.dart';
 import 'package:bonfire/features/error_reporting/controllers/error_reporting.dart';
 import 'package:bonfire/shared/utils/client_access.dart';
+import 'package:bonfire/shared/utils/rest_result_ext.dart';
 import 'package:collection/collection.dart';
 import 'package:flutter/foundation.dart';
 import 'package:riverpod_annotation/riverpod_annotation.dart';
@@ -242,14 +243,23 @@ class AccordMessagesController extends _$AccordMessagesController {
   /// Sends [content] plus file [files] to this channel via
   /// `messages.createWithAttachments`. Each file is a map with `filename`,
   /// `content` (bytes) and optional `content_type`. Falls back to a plain
-  /// [send] when there are no files. Returns true on success.
-  Future<bool> sendWithAttachments(
+  /// [send] when there are no files.
+  ///
+  /// Returns null on success, or the failure message to show the user. The
+  /// server's own reason is passed through — a rejected upload (too large, no
+  /// `attach_files` permission, unsupported type) is otherwise indistinguishable
+  /// from a dead Send button.
+  Future<String?> sendWithAttachments(
     AccordClient client,
     String content,
     List<Map<String, dynamic>> files, {
     String? replyTo,
   }) async {
-    if (files.isEmpty) return send(client, content, replyTo: replyTo);
+    if (files.isEmpty) {
+      return await send(client, content, replyTo: replyTo)
+          ? null
+          : 'Failed to send message.';
+    }
     final data = <String, dynamic>{'content': content.trim()};
     if (replyTo != null) data['reply_to'] = replyTo;
 
@@ -260,14 +270,14 @@ class AccordMessagesController extends _$AccordMessagesController {
     );
     if (!result.ok) {
       debugPrint('Failed to send attachments to $channelId: ${result.error}');
-      return false;
+      return result.errorMessageOr('Failed to send attachments.');
     }
     final message = result.data;
     if (message is AccordMessage) addMessage(message);
     // Action breadcrumb for opt-in error reporting — the fact a message was
     // sent, never its content.
     ref.read(errorReportingControllerProvider.notifier).messageSent();
-    return true;
+    return null;
   }
 
   /// Appends a newly-received message, ignoring duplicates (e.g. the gateway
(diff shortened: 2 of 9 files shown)


krazyjakee:
## Review findings

Scoped to code quality, file structure, and test coverage per the review checklist. No 3D model/render code is touched, so no visual review was needed. File sizes stayed well under 1000 lines (largest touched file is 650 lines), so no split was needed.

### 1. (Medium — fixed) Plain-text send failures still lost the server's reason

`sendWithAttachments`'s no-attachments branch wrapped `send()`'s bare `bool` in a hardcoded `'Failed to send message.'`:

```dart
if (files.isEmpty) {
  return await send(client, content, replyTo: replyTo)
      ? null
      : 'Failed to send message.';
}
```

`send()` itself discards `result.error` behind a `debugPrint`. This is the exact silent-failure pattern this PR sets out to fix, just one level up — a rate limit, a banned-word filter, or a permission error on a text-only message would still show a generic message instead of the server's real reason, while the same failure on an attachment upload would show the specific one. `composer.dart`'s `_send()` also re-derived this same bool→string mapping in a second, duplicated ternary.

**Fix:** extracted `_createMessage`, the single place that calls `messages.create` and turns the `RestResult` into an error message (`result.errorMessageOr(...)`, matching the attachment path). `send()` and `sendWithAttachments`'s no-files branch both delegate to it now, and `composer.dart._send()` always calls `sendWithAttachments` (which already handled the no-files case), dropping the duplicated branch and its own hardcoded fallback string.

### 2. (Low — fixed) No test coverage for `send`/`sendWithAttachments`

Tests were added for `attachment_limits.dart` and `errorMessageOr`, but the controller methods that changed behavior (`sendWithAttachments`'s error passthrough in particular) had no direct test at all.

**Fix:** added `test/features/messaging/accord_messages_send_test.dart`, following the existing `MockClient`-backed pattern from `accord_members_load_test.dart`. Covers: `send` success/failure, `sendWithAttachments` with no files (success, and specifically that a failure surfaces the server's message rather than the generic fallback — the regression-guard for finding #1), and `sendWithAttachments` with files (server message on failure, generic fallback when the server sends none).

### Not changed — flagged for awareness

- `docs/messaging/file-sharing.md`'s copy and the inline-error UI (dismissible `InlineError` above the composer) read fine and match existing conventions; no changes made there.
- `_createMessage`'s empty-content guard now returns a message (`'Message is empty.'`) instead of the previous bare `false`. Every current caller (`composer.dart`, `thread_view.dart`, `voice_text_panel.dart`) already guards against empty text before calling `send`/`sendWithAttachments`, so this is unreachable via the UI today — it's there so a future caller that skips the guard fails loudly instead of silently succeeding.

### Caveat — could not run automated checks

No Dart/Flutter SDK is available in this sandbox, so `flutter analyze` and `flutter test` could not be run here. The diff was reviewed manually against the surrounding code, the accordkit REST/error-envelope source, and existing test patterns (`accord_members_load_test.dart`, `error_reporting_test.dart` for the Hive setup the new test needs). CI on this PR will be the first actual run — I'm subscribed and will fix anything it turns up.

### Housekeeping note

A duplicate PR (#185) was opened against this same underlying commit before this PR was found — closing it as a duplicate of this one.

---

krazyjakee:
## Follow-up review finding

While reviewing the now-closed duplicate (#185) I found one more issue and ported the fix here since this is the canonical PR:

**Low — `errorMessageOr` didn't guard against a whitespace-only server message.** `lib/shared/utils/rest_result_ext.dart`: the emptiness check was `message.isEmpty` on the raw `error?.message`, so a server response with `message: " "` (present but blank) would pass through and render as an empty inline error banner in the composer instead of falling back to the caller's generic message.

**Fixed** in `849ada4`: the check now trims before testing for emptiness. Added a regression test (`returns fallback when the message is whitespace only`) to `test/shared/rest_result_ext_test.dart`.

Everything else from the earlier review (code quality, file size, test coverage) already checked out clean — see the previous review comment on this PR.

---

PR #185, opened 2026-08-04
fix(messaging): surface attachment failures instead of failing silently

krazyjakee:
## Summary

Attaching a file that gets rejected (too large, unreadable, no `attach_files` permission, server error) used to fail with no visible feedback — the Send button just looked dead. Every silent path is now visible:

- Files are screened at pick time against the documented 25 MB limit (`lib/features/messaging/utils/attachment_limits.dart`), naming the file and both sizes on rejection instead of silently dropping it.
- `sendWithAttachments` now returns the server's own failure reason (`String?`) instead of a bare `bool`, via a new `RestResult.errorMessageOr` extension that renders the bare server message rather than the `AccordError(code: …)` wrapper.
- The composer shows send/attach failures in a dismissible inline error above the message box.
- Docs (`docs/messaging/file-sharing.md`) updated to drop the drag-and-drop claim (unimplemented) and document the new behavior.

### Follow-up from self-review

Reviewing the initial change surfaced the same silent-failure gap one level up: the no-attachments path of `sendWithAttachments` wrapped `send()`'s bare `bool` in a hardcoded `'Failed to send message.'`, discarding the real server reason — the exact problem this PR exists to fix, just for plain-text sends. Extracted a shared `_createMessage` helper so `send()` and `sendWithAttachments`'s no-files branch both surface the same server-provided reason, removed the duplicated ternary in `composer.dart`, and added controller-level test coverage (`test/features/messaging/accord_messages_send_test.dart`) that didn't exist before for `send`/`sendWithAttachments`.

## Test plan

- [x] `test/features/messaging/attachment_limits_test.dart` — file screening (size limit, unreadable files, size formatting)
- [x] `test/shared/rest_result_ext_test.dart` — `errorMessageOr`
- [x] `test/features/messaging/accord_messages_send_test.dart` (new) — `send`/`sendWithAttachments` success/failure, and specifically that the no-files path surfaces the server's message rather than a generic fallback
- [ ] `flutter analyze` / `flutter test` — **could not be run**: no Dart/Flutter SDK is installed in this sandbox. Changes were reviewed manually line-by-line against the surrounding code and existing test patterns (`accord_members_load_test.dart`, `error_reporting_test.dart`) instead.
- [ ] Manual UI verification (attach an oversize file, attach an unreadable file, trigger a server-side rejection) — not performed, no running app/device in this environment.

🤖 Generated with [Claude Code](https://claude.ai/code)

https://claude.ai/code/session_01KejDtFqkZEdSUJz5H3diNM

---
_Generated by [Claude Code](https://claude.ai/code/session_01G69JTFuHoebbfVpR7GVFTx)_

### DIFF
--- a/docs/messaging/file-sharing.md
+++ b/docs/messaging/file-sharing.md
@@ -11,9 +11,11 @@ You can share files and images directly in any text channel or DM.
 
 ## Uploading Files
 
-Attach a file to your message using the attachment button in the composer. You can also **drag and drop** files into the message area.
+Attach a file to your message using the attachment button in the composer. Drag and drop is not supported yet.
 
-The maximum file size is **25 MB**. Files larger than this will show an error and will not be uploaded.
+The maximum file size is **25 MB** per file. Larger files are rejected as soon as you pick them, with the file's size shown next to the limit -- nothing is uploaded. Music and video files are the ones that most often run into this; a 320 kbps MP3 passes 25 MB at around ten minutes.
+
+If the server refuses an upload for another reason -- you don't have the **Attach Files** permission in that channel, for example -- the reason it gives is shown above the composer.
 
 ## Pasting Images
 
--- a/lib/features/messaging/controllers/accord_messages.dart
+++ b/lib/features/messaging/controllers/accord_messages.dart
@@ -2,6 +2,7 @@ import 'package:accordkit/accordkit.dart';
 import 'package:bonfire/features/messaging/utils/emoji_catalog.dart';
 import 'package:bonfire/features/error_reporting/controllers/error_reporting.dart';
 import 'package:bonfire/shared/utils/client_access.dart';
+import 'package:bonfire/shared/utils/rest_result_ext.dart';
 import 'package:collection/collection.dart';
 import 'package:flutter/foundation.dart';
 import 'package:riverpod_annotation/riverpod_annotation.dart';
@@ -242,14 +243,23 @@ class AccordMessagesController extends _$AccordMessagesController {
   /// Sends [content] plus file [files] to this channel via
   /// `messages.createWithAttachments`. Each file is a map with `filename`,
   /// `content` (bytes) and optional `content_type`. Falls back to a plain
-  /// [send] when there are no files. Returns true on success.
-  Future<bool> sendWithAttachments(
+  /// [send] when there are no files.
+  ///
+  /// Returns null on success, or the failure message to show the user. The
+  /// server's own reason is passed through — a rejected upload (too large, no
+  /// `attach_files` permission, unsupported type) is otherwise indistinguishable
+  /// from a dead Send button.
+  Future<String?> sendWithAttachments(
     AccordClient client,
     String content,
     List<Map<String, dynamic>> files, {
     String? replyTo,
   }) async {
-    if (files.isEmpty) return send(client, content, replyTo: replyTo);
+    if (files.isEmpty) {
+      return await send(client, content, replyTo: replyTo)
+          ? null
+          : 'Failed to send message.';
+    }
     final data = <String, dynamic>{'content': content.trim()};
     if (replyTo != null) data['reply_to'] = replyTo;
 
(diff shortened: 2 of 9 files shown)


krazyjakee:
## Review findings

Scoped to code quality, file structure, and test coverage per the review checklist. No 3D model/render code is touched, so no visual review was needed. File sizes stayed well under 1000 lines (largest touched file is 650 lines), so no split was needed.

### 1. (Medium — fixed) Plain-text send failures still lost the server's reason

`sendWithAttachments`'s no-attachments branch wrapped `send()`'s bare `bool` in a hardcoded `'Failed to send message.'`:

```dart
if (files.isEmpty) {
  return await send(client, content, replyTo: replyTo)
      ? null
      : 'Failed to send message.';
}
```

`send()` itself discards `result.error` behind a `debugPrint`. This is the exact silent-failure pattern the PR sets out to fix, just one level up — a rate limit, a banned-word filter, or a permission error on a text-only message would still show a generic message instead of the server's real reason, while the same failure on an attachment upload would show the specific one. `composer.dart`'s `_send()` also re-derived this same bool→string mapping in a second, duplicated ternary.

**Fix:** extracted `_createMessage`, the single place that calls `messages.create` and turns the `RestResult` into an error message (`result.errorMessageOr(...)`, matching the attachment path). `send()` and `sendWithAttachments`'s no-files branch both delegate to it now, and `composer.dart._send()` always calls `sendWithAttachments` (which already handled the no-files case), dropping the duplicated branch and its own hardcoded fallback string.

### 2. (Low — fixed) No test coverage for `send`/`sendWithAttachments`

The PR added tests for the new `attachment_limits.dart` utilities and `errorMessageOr`, but the controller methods that changed behavior (`send`'s return semantics were relied on more broadly now, `sendWithAttachments`'s error passthrough) had no direct test at all, before or after the PR.

**Fix:** added `test/features/messaging/accord_messages_send_test.dart`, following the existing `MockClient`-backed pattern from `accord_members_load_test.dart`. Covers: `send` success/failure, `sendWithAttachments` with no files (success, and specifically that a failure surfaces the server's message rather than the generic fallback — this is the regression-guard for finding #1), and `sendWithAttachments` with files (server message on failure, generic fallback when the server sends none).

### Not changed — flagged for awareness

- `docs/messaging/file-sharing.md`'s new copy and the inline-error UI (dismissible `InlineError` above the composer) read fine and match existing conventions; no changes made there.
- `_createMessage`'s empty-content guard now returns a (previously bare `false`) message `'Message is empty.'` instead of silently succeeding — every current caller (`composer.dart`, `thread_view.dart`, `voice_text_panel.dart`) already guards against empty text before calling `send`/`sendWithAttachments`, so this is unreachable via the UI today; it's there so a future caller that skips the guard fails loudly instead of silently.

### Caveat — could not run automated checks

No Dart/Flutter SDK is available in this sandbox (`dart`/`flutter` not on `PATH`, no cached pub packages), so `flutter analyze` and `flutter test` could not be run here. The diff was reviewed manually against the surrounding code, the accordkit REST/error-envelope source, and existing test patterns (`accord_members_load_test.dart`, `error_reporting_test.dart` for the Hive setup the new test needs). CI on this PR will be the first actual run of `flutter analyze`/`flutter test` against these changes — I'm subscribed to this PR and will fix anything CI turns up.

---

krazyjakee:
Closing as a duplicate — #184 already opened this same change (identical head commit) from a different branch, and the review fix + tests from this PR have been pushed there instead: #184.

---

krazyjakee:
## Automated review

Reviewed against code quality, file size/structure, and test coverage. No 3D model/rendering code is touched, so that review dimension doesn't apply.

**CI**: `Analyze & Test` was green on both `stable` and `beta` for `d30962e` before this review's commit.

**File size**: `composer.dart` (643 lines) and `accord_messages.dart` (494 lines) are both well under the 1000-line split threshold.

### Findings

1. **Low — `errorMessageOr` didn't guard against a whitespace-only server message.** `lib/shared/utils/rest_result_ext.dart`: the emptiness check was `message.isEmpty` on the raw `error?.message`, so a server response with `message: " "` (present but blank) would pass through and render as an empty inline error banner in the composer instead of falling back to the caller's generic message. **Fixed** in `5077b9b`: the check now trims before testing for emptiness. Added a regression test (`returns fallback when the message is whitespace only`) to `test/shared/rest_result_ext_test.dart`.

2. **Info — two other observations, not changed:**
   - `composer.dart` has no widget tests before or after this PR (attach/dismiss/error-banner flow included). Given the Riverpod + platform-channel scaffolding a proper widget test would need, and that this environment has no Flutter SDK to verify one against, I didn't add one blind rather than risk landing something broken. The new logic that *is* independently testable (`screenAttachments`, `errorMessageOr`, `send`/`sendWithAttachments`) already has solid coverage from this PR.
   - `docs/messaging/file-sharing.md`'s "drag and drop is not supported yet" claim checks out — no `DropTarget`/`DropRegion`/`desktop_drop` usage anywhere in `lib/`.

Everything else — the `_createMessage` extraction, `screenAttachments`, the `_error` banner wiring in the composer, and the new test files — checked out on manual read: correct control flow, no duplicated-logic or missing-error-handling issues, and consistent with surrounding conventions (matches the `errorOr`/`listOrLog` extension patterns already in `rest_result_ext.dart`).

---

PR #186, opened 2026-08-04
fix(updates): ship the Linux bundle as .tgz to unstrand <=0.2.6 .deb installs

krazyjakee:
## The problem

#178 fixed self-update for Linux system (`.deb`) installs — but it shipped **in v0.2.7**, the release that affected clients can't apply. Their selector is frozen at:

```dart
if (UniversalPlatform.isLinux) return const ['.tar.gz'];   // v0.2.6, no writable-root check
```

So a `.deb` install under root-owned `/opt` downloads the portable tarball, extracts it, and runs the swap helper — whose `mv "$dst" "$dst.old"` fails with EACCES, is swallowed by `2>/dev/null`, and then relaunches the *old* binary. The app restarts, the version never changes, and nothing is shown to the user.

Found on a real machine still on 0.2.6 hours after 0.2.7 shipped. The forensic trail: `daccord-linux-x86_64.tar.gz` downloaded 20:10:57, extracted to `staged/` 20:11:28, `apply.sh` self-deleted (so the helper ran to completion), no `.old` directory anywhere, `/opt/daccord` untouched since Jul 9, and no `pkexec`/`dpkg` activity in the polkit or dpkg logs. The helper logic itself is fine — replayed against a writable sandbox root it swaps correctly.

## Why this can't be fixed in the client

Those installs run frozen code. Any fix we merge reaches them only through a release they are structurally unable to install. The one input we still control is **what the release contains**.

## The fix

Ship the Linux bundle as `.tgz`. With no `.tar.gz` asset present, a v0.2.6 client:

| | before | after |
|---|---|---|
| `platformAsset()` | the `.tar.gz` | `null` |
| `canInstallInPlace` | `true` | `false` |
| `prepareUpdate()` | silently downloads a doomed swap | no-ops |
| banner | nothing (stays hidden until "ready") | **"Update available — tap to view"** |
| Updates page | — | **"Download"** → falls through `_downloadExts` to the **`.deb`** |

A silent no-op becomes a working manual-install prompt.

Supporting changes:
- **Release guard** — the release job fails if any asset ends in `.tar.gz`, since publishing one re-arms the bug. Verified it fires on a regressed asset set and stays silent on the new one.
- **Release body** — carries a manual-install notice for affected users. The in-app Updates page renders this (`AppRelease.notes`), so it reaches them inside the app, not just on GitHub.
- `.tar.gz` stays **accepted** by the selector so older releases remain applicable; only the *published* suffix changes.

## Scope

Windows and macOS need no equivalent: `installer.iss` is `PrivilegesRequired=lowest` so `{autopf}` resolves per-user and writable, and `/Applications` is admin-writable. Only root-owned `/opt` was affected. Disarming Windows wouldn't even work — with the windows zip removed, a v0.2.6 client's `.zip` fallback would select `daccord-web.zip`.

## Trade-off

v0.2.7 **portable** (writable-root) installs match `.tar.gz`, so they fall back to a manual download for one cycle; one-click returns for them at v0.2.8. The `.deb` population — the recommended install, and the broken one — keeps one-click via `pkexec` throughout. Reversible if you'd rather weight it the other way.

## Testing

- `flutter test` — 516 passed, including two new cases: `.tgz` is applied on a writable root, and legacy `.tar.gz` bundles stay installable.
- `flutter analyze --no-fatal-infos` — clean on the touched directories.
- Release guard exercised both ways.
- `pkexec dpkg -i` (the exact command the v0.2.7 path runs) verified end-to-end on a real 0.2.6 → 0.2.7 upgrade.

## Follow-ups (not in this PR)

1. **An apt repo** — `.deb` users would update through `apt` and never depend on the in-app updater, making this class of bug unreachable.
2. **Post-apply verification** — #178 gates on a *probe* predicting the swap will work, not a check that it did. Stamping the attempted version before quitting and comparing on next launch would make any future silent no-op report itself.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -278,9 +278,17 @@ jobs:
 
       - name: Package (Linux)
         if: matrix.platform == 'linux'
+        # `.tgz`, NOT `.tar.gz` — this suffix is load-bearing. Clients at v0.2.6
+        # and earlier pick a `.tar.gz` asset unconditionally on Linux and try to
+        # swap it into a `.deb` install's root-owned /opt, which silently fails
+        # and relaunches the old build; those clients are frozen, so withholding
+        # the suffix is the only way to break them out of the loop. With no
+        # `.tar.gz` present they fall back to offering the `.deb` as a manual
+        # download. See update_controller._installableExts and the guard in the
+        # release job below.
         run: |
           cd build/linux/x64/release/bundle
-          tar czf "$GITHUB_WORKSPACE/${{ matrix.artifact }}.tar.gz" .
+          tar czf "$GITHUB_WORKSPACE/${{ matrix.artifact }}.tgz" .
 
       - name: Build .deb (Linux)
         if: matrix.platform == 'linux'
@@ -359,7 +367,7 @@ jobs:
           name: ${{ matrix.artifact }}
           path: |
             ${{ matrix.artifact }}.zip
-            ${{ matrix.artifact }}.tar.gz
+            ${{ matrix.artifact }}.tgz
             ${{ matrix.artifact }}.apk
             ${{ matrix.artifact }}.deb
             daccord-windows-x86_64-setup.exe
@@ -378,6 +386,21 @@ jobs:
           pattern: daccord-*
           merge-multiple: true
 
+      - name: Guard against a .tar.gz asset
+        # Publishing any asset whose name ends in `.tar.gz` re-arms the broken
+        # self-updater in clients ≤ v0.2.6: on Linux they select it regardless of
+        # whether the install root is writable, so a `.deb` install downloads it,
+        # runs a swap against root-owned /opt that can't succeed, and relaunches
+        # the old build with no error shown. The Linux bundle ships as `.tgz` so
+        # those clients find nothing installable and offer the `.deb` instead.
+        run: |
+          cd artifacts
+          if ls | grep -q '\.tar\.gz$'; then
+            echo "::error::Refusing to publish a .tar.gz asset — it silently breaks self-update on clients <= v0.2.6. Ship the Linux bundle as .tgz."
+            ls
+            exit 1
+          fi
+
       - name: Generate checksums
         # SHA256SUMS.txt lets the in-app self-updater verify a downloaded bundle
         # before swapping it in (see update_controller._expectedSha), and gives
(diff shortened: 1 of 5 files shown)


krazyjakee (commented):
## Automated review

Reviewed the diff (`.github/workflows/release.yml`, `docs/MIGRATION.md`, `lib/features/updates/controllers/update_controller.dart`, `test/features/updates/update_controller_test.dart`) against code quality, file size, and test coverage. No `flutter`/`dart` SDK is available in this sandbox, so I could not run `flutter analyze`/`flutter test` myself — relying on the "Analyze & Test (beta)" check, which is green, plus manual tracing of the logic.

### Findings

**No functional defects found.** Specifically verified:
- `update_installer_io.dart:_extract` already branches on `name.endsWith('.tar.gz') || name.endsWith('.tgz')` (line 220), so the swap helper correctly applies a downloaded `.tgz` — the client-side extraction path was already `.tgz`-aware before this PR, it just never had an asset to exercise it.
- `_installableExts`/`_downloadExts` priority-list changes are self-consistent: `.tgz` is tried before `.tar.gz` everywhere, and `.tar.gz` stays in every list so a legacy release remains installable/downloadable.
- No other repo reference (`dist/`, other workflows, other docs) still hardcodes `daccord-linux-x86_64.tar.gz` — `docs/MIGRATION.md` was the only doc needing the update, and it was updated.
- The new release-workflow guard (`grep -q '\.tar\.gz$'`) runs after `Download all artifacts` and before `Generate checksums`/`Create GitHub Release`, so a regression fails the job before anything is published — correct ordering.
- Test coverage: both new cases (`.tgz` applies on a writable root; legacy `.tar.gz` stays installable) are present, and the pre-existing `.tar.gz`-vs-`.deb` ordering/priority tests were correctly left as `.tar.gz` (they test the legacy-compat path, not the new default) rather than blanket-renamed.
- All four touched files are well under the 1000-line split threshold (519 / 557 / 768 / 64 lines).
- No 3D/rendering code is touched — visual review doesn't apply here.

**One non-blocking observation (low severity, not fixed):** the new `body:` block in the "Create GitHub Release" step (release.yml) unconditionally prepends the "Linux `.deb` users on v0.2.6 or earlier…" notice to *every* future release's notes, not just the one that ships this fix — there's no version gate or expiry. The in-code comment already flags this ("Drop this notice once ≤ v0.2.6 installs have aged out"), so it reads as a deliberate, human-tracked trade-off rather than an oversight, and I didn't want to unilaterally change release-messaging behavior on a judgment call outside this PR's stated scope. Worth a reminder/issue so it doesn't silently ship on v0.3.0+.

No fixes pushed — nothing met the bar of an actual bug once traced through the extraction path and asset-priority lists.

---

krazyjakee:
Reviewed and pushed one fix. The approach is sound — `update_installer_io._extract` already accepts `.tgz` alongside `.tar.gz`, so the published suffix change needs no installer work, and a v0.2.6 Linux client's `_downloadExts` has no `.zip` entry, so it can't misselect `daccord-web.zip` the way the Windows path would have.

**The gap:** this PR's whole recovery path is "download the `.deb` and install it by hand," but `docs/getting-started/installation.md` sent Linux users to `daccord-linux-x86_64.zip` and `daccord-linux-arm64.zip` — neither has ever been published. The `.deb`, the one file stranded users need, wasn't listed at all. Confirmed against `gh release view v0.2.7`:

```
daccord-android.apk
daccord-linux-x86_64.deb
daccord-linux-x86_64.tar.gz
daccord-macos-universal.dmg
daccord-web.zip
daccord-windows-x86_64-setup.exe
daccord-windows-x86_64.zip
SHA256SUMS.txt
```

macOS was also listed as `daccord-macos.dmg` vs. the real `daccord-macos-universal.dmg`. Table now matches what the workflow actually publishes (portable bundle under its new `.tgz` name), the Linux section leads with the `.deb`, and it carries the same one-time manual-install notice the release body does — so the in-app notice and the docs agree.

Pre-existing staleness rather than something this PR introduced, but it defeats the recovery path this PR is built around, so it belongs here.

PR #187, opened 2026-08-04
fix(channels): clear voice channels' unread badge and ack it to the server

krazyjakee:
A voice channel's unread dot never went away — e.g. **Dev Talk** stayed lit no matter what.

## Root cause

`_openChannel` in `accord_home.dart` gated the read-clear on `channel?.type != 'voice'`, so tapping a voice channel joined the call and opened the tab but never marked it read or POSTed `channels.ack`. The only voice-side clear lived in `VoiceTextPanel.initState`, with two gaps:

1. it runs only when the user explicitly opens the call's chat panel (off by default), and
2. it cleared **local** state only — with no ack, the next gateway READY re-hydrated the badge from the server's `unread` array, so the dot came straight back.

The context menu's *Mark as read* had a related hole: it acked only the newest **cached** message. A voice channel's history is almost never cached, so the ack was skipped and the clear didn't survive a reconnect.

## Changes

- **New `lib/features/channels/utils/mark_channel_read.dart`** — one helper that does the local `markRead` *and* the `channels.ack`, acking the newest cached message or falling back to the channel's `last_message_id`. Takes an optional `serverKey` so a call pinned to a background server acks on the right connection.
- **`accord_home.dart`** — dropped the `!= 'voice'` guard; `_markChannelRead` delegates to the helper.
- **`voice_text_panel.dart`** — acks (with the `last_message_id` fallback) instead of clearing locally.
- **`channel_context_menu.dart`** — *Mark as read* gained the `last_message_id` fallback, so it sticks for never-opened channels.
- **`accord_message_content.dart`** — mention-tap channel open acks too, same phantom-unread fix.

Net effect: voice channels now clear exactly like text channels, and every clear moves the server's read position so it survives a reconnect.

## Testing

- `flutter analyze --no-fatal-infos` — 0 errors (only the pre-existing infos).
- `flutter test` — 483 passing.
- Not verified live: reproducing needs a signed-in session with an unread voice channel. The ack goes through the same `client.channels.ack` path text channels already use.

### DIFF
--- a/lib/features/channels/utils/mark_channel_read.dart
+++ b/lib/features/channels/utils/mark_channel_read.dart
@@ -1,0 +1,41 @@
+import 'package:bonfire/features/authentication/repositories/accord_auth.dart';
+import 'package:bonfire/features/channels/controllers/read_state.dart';
+import 'package:bonfire/features/messaging/controllers/accord_messages.dart';
+import 'package:bonfire/features/server/controllers/connections.dart';
+import 'package:flutter_riverpod/flutter_riverpod.dart';
+
+/// Clears [channelId]'s unread badge locally *and* POSTs `channels.ack` so the
+/// server's read position catches up too.
+///
+/// Both halves matter: the local clear hides the dot now, the ack is what stops
+/// the READY payload's `unread` array from re-lighting it on the next connect.
+/// The acked position is the newest cached message, falling back to
+/// [fallbackMessageId] (the channel's `last_message_id`) when the message cache
+/// is empty — which is the common case for a channel whose messages were never
+/// opened (any voice channel joined without its chat panel) and for a *phantom*
+/// unread whose message has since been deleted.
+///
+/// [serverKey] pins the ack to a specific connection; it defaults to the active
+/// one. Pass it explicitly when the channel may live on a background server
+/// (e.g. a voice call pinned to a server the user has since navigated away
+/// from).
+void markChannelRead(
+  WidgetRef ref,
+  String channelId, {
+  String? serverKey,
+  String? fallbackMessageId,
+}) {
+  final key = serverKey ?? ref.read(connectionsControllerProvider).activeKey;
+  if (key == null) return;
+  ref.read(readStateControllerProvider(key).notifier).markRead(channelId);
+
+  final messages = ref.read(accordMessagesControllerProvider(channelId));
+  final lastId =
+      messages?.isNotEmpty == true ? messages!.last.id : fallbackMessageId;
+  if (lastId == null) return;
+  ref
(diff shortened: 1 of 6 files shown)


krazyjakee:
## Review

Went through the diff against the three focus areas.

**Code quality** — solid, no issues found.
- `markChannelRead`'s use of `messages!.last.id` as the newest message is correct: `AccordMessagesController.build`'s doc comment and `_load` both confirm the cache is stored oldest→newest, and the REST result is explicitly reversed to match.
- Null-safety throughout is careful: `clientForKey(key)` returning `null` (no live connection) short-circuits via `?.channels.ack(...)` rather than throwing, so the local `markRead` clear always happens even when the ack can't be sent — which is the actual fix for the voice-channel bug (message history is rarely cached for those channels).
- The four call sites (`accord_home.dart`, `channel_context_menu.dart`, `accord_message_content.dart`, `voice_text_panel.dart`) were all migrated consistently to the shared helper, and the old duplicated logic was fully removed rather than left as dead code.
- Import cleanup is correct — no stale `read_state`/`accord_messages` imports left behind at the call sites now routed through `mark_channel_read.dart`.

**File size / structure** — no file in the diff is anywhere close to the 1000-line threshold (largest touched file is `accord_home.dart` at 652 lines); no splitting needed.

**Test coverage** — this was the one gap. The new `mark_channel_read.dart` had zero tests, and it's exactly the kind of helper worth covering (key-resolution logic, the null-client no-op path). Added `test/features/channels/mark_channel_read_test.dart` with 4 cases, using a `Consumer`-captured `WidgetRef` against a real `ProviderContainer` (no HTTP mocking needed, since the interesting behavior — local unread clearing — doesn't depend on the ack REST call succeeding):
- explicit `serverKey` clears local state with no cached messages and no live client (the voice-channel scenario this PR fixes)
- falls back to the active connection when no `serverKey` is passed
- no-ops when there's neither a `serverKey` nor an active connection
- an explicit `serverKey` wins over an unrelated active connection

I did not add a test asserting on the `channels.ack` REST call's arguments — `clientForKey` reads `AccordAuth`'s private `_connections` map, which has no public test seam short of driving a full mocked login flow, and it isn't proportionate for this fix. The `messages.last.id` vs. `fallbackMessageId` selection logic is exercised implicitly by inspection above rather than directly asserted, for the same reason.

**Note:** I could not run `flutter analyze` or `flutter test` locally (no Flutter SDK in this container) — verified by inspection only; CI is the authoritative check. `Analyze & Test` was still in progress on the head commit as of this review.

No 3D/visual changes in this diff (channel unread-state logic only), so no rendering review was needed.

Pushed as `7a53fb4` on this branch (test-only addition, no production code changed).

---

PR #188, opened 2026-08-04
fix(messaging): send forum thread replies and DMs on Enter

krazyjakee:
## Problem

Typing a reply in a forum post and pressing Enter inserted a newline instead of sending.

## Cause

The forum/thread reply composer (`lib/features/messaging/views/thread_view.dart`) is a multiline `TextField` (`minLines: 1, maxLines: 4`) with no `textInputAction`. Flutter defaults such a field to `TextInputAction.newline`, and `EditableText.performAction` deliberately does nothing for `newline` on a multiline field — so Enter only inserted a line break and `onSubmitted` was never called. The main channel composer works because it already sets `textInputAction: TextInputAction.send`.

The DM conversation composer (`lib/features/user/views/accord_direct_messages_conversations.dart`) had the identical defect, so it's fixed the same way here.

## Change

- `textInputAction: TextInputAction.send` on both composers, matching `message_pane/composer.dart`.
- Each gets a `FocusNode` that is re-focused after a successful send — submitting unfocuses the field, so without this you'd have to click back in for every message.

The forum *post creation* dialog (`post_composer_dialog.dart`) is intentionally untouched: Enter should insert newlines in a post body, and it has an explicit Post button.

## Testing

- `flutter analyze --no-fatal-infos` clean on both files.
- No behaviour change on mobile beyond the keyboard's action key becoming "send", consistent with the main composer.

### DIFF
--- a/lib/features/messaging/views/message_pane/composer.dart
+++ b/lib/features/messaging/views/message_pane/composer.dart
@@ -365,7 +365,9 @@ class _ComposerState extends ConsumerState<_Composer> {
       _saveDraft(widget.channelId, '');
       _lastTypingSent = null;
       widget.onCancelReply?.call();
-      _focusNode.requestFocus();
+      // After the frame that re-enables the field: the send disabled it, and a
+      // disabled TextField refuses focus requests outright.
+      refocusAfterFrame(_focusNode);
     }
   }
 
--- a/lib/features/messaging/views/message_pane/message_pane.dart
+++ b/lib/features/messaging/views/message_pane/message_pane.dart
@@ -45,6 +45,7 @@ import 'package:bonfire/shared/components/context_menu.dart';
 import 'package:bonfire/shared/utils/client_access.dart';
 import 'package:bonfire/shared/utils/confirm_dialog.dart';
 import 'package:bonfire/shared/utils/platform.dart';
+import 'package:bonfire/shared/utils/refocus.dart';
 import 'package:bonfire/theme/theme.dart';
 import 'package:cached_network_image/cached_network_image.dart';
 import 'package:collection/collection.dart';
--- a/lib/features/messaging/views/thread_view.dart
+++ b/lib/features/messaging/views/thread_view.dart
@@ -16,6 +16,7 @@ import 'package:bonfire/features/spaces/utils/message_time.dart';
 import 'package:bonfire/shared/components/async_state_views.dart';
 import 'package:bonfire/shared/components/context_menu.dart';
 import 'package:bonfire/shared/utils/confirm_dialog.dart';
+import 'package:bonfire/shared/utils/refocus.dart';
 import 'package:bonfire/theme/theme.dart';
 import 'package:flutter/material.dart';
 import 'package:flutter/services.dart';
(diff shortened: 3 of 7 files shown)


krazyjakee:
## Automated review

Reviewed against code quality, file size/structure, test coverage, and visual-change criteria.

**No issues found — no code changes pushed.**

### Code quality
The fix is a minimal, correct wiring change: `textInputAction: TextInputAction.send` + a `FocusNode` that's re-requested after a successful send. This is an exact match for the identical, already-shipped pattern in the main channel composer (`lib/features/messaging/views/message_pane/composer.dart:445-451,361`) — same field order (`controller` → `focusNode` → `enabled` → `minLines`/`maxLines` → `textInputAction`), same `if (!mounted) return;` guard before `setState`, same "only refocus on success" behavior. Both `FocusNode`s are correctly disposed in `dispose()`. No error handling, naming, or duplication concerns beyond what already exists identically in the reference composer (not introduced by this diff).

### File size / structure
Both touched files are well under the 1000-line threshold (`thread_view.dart`: 611 lines, `accord_direct_messages_conversations.dart`: 595 lines). No split needed.

### Test coverage
No widget-level test exercises this Enter-to-send wiring — but that's true repo-wide, including for the reference composer this PR mirrors (`message_pane/composer.dart`, already in production, also untested at this level). The existing test suite for messaging (`forum_view_logic_test.dart`, `federation_messages_test.dart`, etc.) covers pure logic and Riverpod controllers directly; nothing constructs a full `ProviderScope` + mocked `AccordClient` for these stateful composer widgets. Adding that harness here would be a disproportionate, unrelated addition for a 2-file/21-line consistency fix, so I'm flagging it as a suggested follow-up rather than blocking this PR:
- A shared, extracted "message input field" widget (currently near-duplicated three times: composer.dart, thread_view.dart, accord_direct_messages_conversations.dart) would make the Enter-to-send contract independently testable via `tester.testTextInput.receiveAction(TextInputAction.send)`. Worth doing only if/when a fourth composer shows up or this logic needs to change again — not warranted for this fix alone per the repo's minimal-diff convention.

### 3D model / visual changes
N/A — no 3D/rendering code touched.

### CI
`Analyze & Test (beta)` passed; `Analyze & Test (stable)` was still running at review time — worth a final glance once it completes.

---

krazyjakee:
Reviewed and pushed a fix: the `FocusNode.requestFocus()` calls in this PR were silently dropped, so the "click back in for every message" problem they were meant to solve survived.

All three composers disable their `TextField` while a send is in flight (`enabled: !_sending`). A disabled `TextField` pushes `canRequestFocus = false` onto its `FocusNode`, and `FocusNode.requestFocus` **returns early without queueing** when `canRequestFocus` is false. Restoring focus in the same synchronous turn as `setState(() => _sending = false)` runs *before* the rebuild that re-enables the field, so the request lands while the node still refuses it.

Verified with a widget test driving the real send cycle — focus, submit via `TextInputAction.send`, settle:

| refocus strategy | focused after send |
|---|---|
| `node.requestFocus()` (this PR as opened) | `false` |
| deferred to a post-frame callback | `true` |

Added `refocusAfterFrame()` in `lib/shared/utils/` and applied it to the two composers here plus `message_pane/composer.dart`, which carried the same latent bug and is the composer the other two were modelled on. `test/shared/refocus_test.dart` asserts both rows of that table, so the reason the helper exists is pinned rather than left as a comment.

Also rebased onto master to pick up #184, which rewrote `_send` in `composer.dart`. Locally: `flutter analyze --no-fatal-infos` clean (9 pre-existing infos) and 504 tests passing.

The Enter-to-send change itself was correct as written.

PR #189, opened 2026-08-04
feat(messaging): attach files by dragging them onto the composer

krazyjakee:
Files could only become attachments via the picker or a clipboard paste — dragging one from a file manager onto the message box did nothing.

This wraps the composer in a `desktop_drop` `DropTarget` and funnels dropped items through the same screening and pending-attachment path the picker uses, so they render as the usual chips and go out with the same send call. While a drag hovers, the box outlines and the hint reads "Drop files to attach".

### Details

- **Size is checked before reading.** `length()` runs ahead of `readAsBytes()`, so a dropped 4 GB video is rejected rather than pulled into memory first.
- **Failures are visible.** Folders, oversize files and unreadable files surface in the composer's existing inline error instead of vanishing. `_addFiles` gained an `alsoRejected` parameter so caller-side rejections merge with the ones `screenAttachments` produces, and the oversize wording reuses `attachment_limits`.
- **macOS sandbox.** The item's security-scoped bookmark is opened around the read, so files dragged in from outside the container work in the App Store build (`files.user-selected.read-write` is already entitled).
- **Covered targets.** The drop target is disabled while sending and while a dialog or sheet is on top (`ModalRoute.isCurrent`) — `desktop_drop` otherwise keeps delivering events to targets that aren't visible.

Adds `desktop_drop: ^0.7.1` (Linux / Windows / macOS / Web / Android; a no-op on iOS, where it only installs an inbound method-call handler and never invokes the channel).

Scope note: this covers the main channel composer. The thread-reply composer in `thread_view.dart` has no attachment support at all — no attach button, no upload path — so there was nothing to hook a drop into there.

### Testing

- `flutter analyze --no-fatal-infos lib/` — clean (9 pre-existing infos, none in messaging)
- `flutter test` — 506 passing
- `flutter build linux --debug` — succeeds, native plugin registered

### DIFF
--- a/docs/messaging/file-sharing.md
+++ b/docs/messaging/file-sharing.md
@@ -11,7 +11,7 @@ You can share files and images directly in any text channel or DM.
 
 ## Uploading Files
 
-Attach a file to your message using the attachment button in the composer. Drag and drop is not supported yet.
+Attach a file to your message using the attachment button in the composer, or drag files from your file manager and drop them onto the composer.
 
 The maximum file size is **25 MB** per file. Larger files are rejected as soon as you pick them, with the file's size shown next to the limit -- nothing is uploaded. Music and video files are the ones that most often run into this; a 320 kbps MP3 passes 25 MB at around ten minutes.
 
--- a/lib/features/messaging/utils/attachment_limits.dart
+++ b/lib/features/messaging/utils/attachment_limits.dart
@@ -21,6 +21,23 @@ String formatFileSize(int bytes) {
   return '$bytes bytes';
 }
 
+/// Message for a file that couldn't be read into memory (a cloud- or
+/// provider-backed file the platform declined to hand over bytes for).
+/// Shared by the picker and drag-and-drop paths so the wording never drifts
+/// between the two.
+String unreadableAttachmentMessage(String name) =>
+    "$name couldn't be read. Copy it to local storage and try again.";
+
+/// Message for a file that exceeds [maxBytes]. Shared by the picker and
+/// drag-and-drop paths so the wording never drifts between the two.
+String oversizeAttachmentMessage(
+  String name,
+  int size, {
+  int maxBytes = kMaxAttachmentBytes,
+}) =>
+    '$name is ${formatFileSize(size)} — the limit is '
+    '${formatFileSize(maxBytes)}.';
+
 /// The outcome of screening picked files: the ones that can be attached, plus a
 /// line per file that can't be, naming it and why.
 class AttachmentScreening {
@@ -53,15 +70,12 @@ AttachmentScreening screenAttachments(
   for (final file in files) {
     final bytes = file.bytes;
     if (bytes == null) {
-      rejections.add(
-        "${file.name} couldn't be read. Copy it to local storage and try again.",
-      );
+      rejections.add(unreadableAttachmentMessage(file.name));
       continue;
     }
     if (bytes.length > maxBytes) {
       rejections.add(
-        '${file.name} is ${formatFileSize(bytes.length)} — the limit is '
-        '${formatFileSize(maxBytes)}.',
+        oversizeAttachmentMessage(file.name, bytes.length, maxBytes: maxBytes),
       );
       continue;
     }
--- a/lib/features/messaging/utils/dropped_entity.dart
+++ b/lib/features/messaging/utils/dropped_entity.dart
@@ -1,0 +1,12 @@
+/// Directory detection for files dropped onto the composer.
+///
+/// `desktop_drop` only types directories as `DropItemDirectory` on macOS and
+/// web — its shared `performOperation` handler, which Linux and Windows both
+/// use, wraps every dropped path in a `DropItemFile` with no directory check.
+/// Without this, a folder dropped on Linux or Windows falls through to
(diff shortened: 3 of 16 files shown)


krazyjakee (commented):
Automated review — pushed a follow-up commit (`df0b848`) to this branch with fixes. Ranked findings:

**1. (Low) Stale doc — "Drag and drop is not supported yet"**
`docs/messaging/file-sharing.md` still told users drag-and-drop wasn't supported, which this PR makes false. Fixed the line to describe the new behavior.

**2. (Low) Duplicated rejection-message wording (code quality)**
`_onDrop`'s oversize/"couldn't be read" strings in `composer.dart` were hand-copied from `screenAttachments()` in `attachment_limits.dart` rather than shared, so the two paths (picker vs. drop) could silently drift out of sync if either wording changes later. Extracted `oversizeAttachmentMessage()` and `unreadableAttachmentMessage()` into `attachment_limits.dart` and pointed both call sites at them.

**3. (Low) Missing test coverage for the new messages**
The extracted helpers are now pure functions, so I added unit tests in `attachment_limits_test.dart` covering both, including one asserting the drop path's wording matches `screenAttachments()`'s exactly.

**Other things checked, no issues found:**
- Verified against the `desktop_drop` 0.7.1 API (`DropItem.length()`, `readAsBytes()`, `path`, `extraAppleBookmark`, `DesktopDrop.instance.start/stopAccessingSecurityScopedResource(bookmark:)`) — usage matches the published signatures.
- Size-before-read ordering, directory rejection, scoped-access try/finally, and the `ModalRoute.isCurrent`/`_sending` gating on the `DropTarget` all look correct; no leaked scoped-resource handles or setState-after-dispose paths found.
- File sizes are well within the 1000-line guideline (composer.dart is 739 lines); no split needed.
- No 3D model/rendering code involved — visual review step doesn't apply to this PR.

**Not done, flagging as a gap rather than fixing silently:** there's no widget-level test exercising `_onDrop`/drag-hover state directly — doing so would require new platform-channel mocking scaffolding for `desktop_drop` that doesn't exist in this repo yet (there's no precedent for mocking `file_picker`'s channel either). I limited test additions to the pure logic that was extractable without that infrastructure; a follow-up could add a `TestDefaultBinaryMessenger` harness for `desktop_drop` if deeper coverage is wanted.

I don't have a Flutter/Dart SDK available in this environment, so I could not run `flutter analyze` or `flutter test` to confirm — please run CI to verify.

---

krazyjakee:
Reviewed. Verified the parts that carry risk, found one defect, pushed a fix.

## Fixed: dropped folders are misreported on Linux and Windows

The `item is DropItemDirectory` branch only fires on **macOS and web**. desktop_drop's shared `performOperation` handler — the one Linux and Windows both use — wraps every dropped path in a `DropItemFile` with no directory check (`channel.dart:100`); only `performOperation_macos` and the web implementation ever construct a `DropItemDirectory`.

So on Linux/Windows a dropped folder fell through to `XFile.length()`. Confirmed against the real API:

```
length() on a directory THREW: FileSystemException: Cannot retrieve length of file (OS Error: Is a directory, errno = 21)
```

That's caught by the `catch (_)` and reported as `unreadableAttachmentMessage` — *"X couldn't be read. Copy it to local storage and try again"* — advice that cannot help someone who dropped a folder, on the two platforms most likely to be dropping one. Now checks the path alongside the type check, behind the repo's usual conditional-export split so web compiles against a stub.

## Verified, no change needed

- **iOS** — the "no-op" claim holds. desktop_drop declares macos/linux/windows/android/web only, and `DesktopDrop.init()` does nothing but `setMethodCallHandler`, which is Dart-side registration. No `MissingPluginException` path.
- **macOS sandbox** — `AppStore.entitlements` does carry `files.user-selected.read-write`. Release/DebugProfile run unsandboxed.
- **The `enable` toggle can't strand `_dragging`** — `DropTarget.didUpdateWidget` fires `onDragExited` when it goes false mid-drag.
- **`PlatformFile.path` on web** — set but never read (`_send` uses `.bytes`/`.name`/`.extension`, `_AttachmentChip` uses `.name`), so file_picker's web-throwing `path` getter is never hit.
- **Web build** succeeds, including on the new stub path.

## Not fixed — worth a follow-up

**The drop zone is only the composer strip.** It's a ~50px bar under a full-height pane, and dropping on the message list does nothing at all: no hover outline, no error. That's the silent-failure mode #184 was written to eliminate, in a new form — and dropping "on the conversation" is where most people will aim. Left alone because moving the target up to the pane means lifting `_attachments`/`_error`/`_addFiles` out of `_ComposerState` or reaching in through a GlobalKey, which is more than this PR's scope. Happy to do it separately.

**Per-file limit only.** A multi-file drop makes bulk selection much easier than the picker did, and 20 × 24 MB all land in memory as `Uint8List` before the multipart body copies them again. Pre-existing (#184 flagged the same profile), just easier to reach now.

## Build coverage

CI's `build` job is `workflow_dispatch`-only, so a PR adding native plugin code across five platforms gets **zero** build coverage from the merge gate. I dispatched the matrix against this branch — https://github.com/DaccordProject/daccord/actions/runs/30950848950 — and will hold the merge on Android/Windows, which are the ones neither of us has built.

Issue #141, opened 2026-06-14
DM call parity follow-ups: mid-call mute/deafen broadcast + PiP for DM calls


krazyjakee:
Follow-ups explicitly called out in #134's description, tracked here so they don't get lost:

- **Mid-call mute/deafen isn't broadcast over the gateway for DM calls.** The gateway voice-state op is space-scoped, so for a DM call only the *initial* join state is sent; local mute/deafen toggles work locally but peers don't see them update. (Group-call join/leave presence still updates via `voice.state_update`.) Needs a DM-scoped voice-state update path (likely an accordserver change).
- **PiP overlay remains space-only.** Its reopen path is keyed on a space, so a DM call stays in the full-screen view with no picture-in-picture. Generalize the PiP key to handle a space-less (DM) call.

Surfaced during the UX review of #134.

_Severity: low — both are known parity gaps, not regressions._

krazyjakee:
Half of this is now done; the other half is blocked on `accordserver`.

**PiP for DM calls — done** (#136, daa7ba8e). `voice_pip_overlay.dart` no longer early-returns on `spaceId == null`, and the reopen path is DM-aware: it calls `showFullScreenVoice(channelId, spaceId: null, channelName: <dm title>)` instead of routing through the space channel opener, which can't resolve a DM channel. Audio-only DM calls still show no PiP, consistent with space behaviour.

**Mid-call mute/deafen broadcast — blocked** (#135). I verified the blocker rather than assuming it: `GatewaySocket.updateVoiceState` takes `spaceId` as a non-nullable positional and puts `space_id` in the payload, so `VoiceController._sendVoiceStateUpdate()`'s early return isn't an oversight — there is nothing for it to call. No REST fallback exists either; the voice endpoint surface is `getInfo`/`join`/`leave`/`ring`/`declineCall`/`cancelCall`/`listRegions`/`getStatus`, and `join` carries only the *initial* self-mute/self-deaf, which is precisely why peers see the opening state and nothing after. Full detail in #135.

Leaving this open tracking the mute/deafen half only. It closes when accordserver gains a DM-scoped voice-state route and accordkit relaxes the `spaceId` signature — the client change here is then a few lines.

krazyjakee:
The remaining half is now written — mid-call mute/deafen broadcast is covered by DaccordProject/accordserver#46 (server) and #203 (client). PiP for DM calls landed earlier in #136 / daa7ba8e.

[...]

Issue #136, opened 2026-06-14
DM calls: support the PiP overlay (currently space-only)


krazyjakee:
Follow-up from #127 / #134 (DM voice/video calls).

## Problem
The picture-in-picture overlay (`VoicePipOverlay`) doesn't appear for DM calls. While in a DM call you can't navigate away and keep a floating call window the way you can for space voice channels — the DM call lives only in the full-screen voice view.

## Cause
`lib/features/voice/views/voice_pip_overlay.dart` early-returns when `spaceId == null`:

```dart
if (!voice.isConnected || channelId == null || spaceId == null || viewingVoice) {
  return const SizedBox.shrink();
}
```

and its `onOpen(channelId, spaceId)` reopen path is keyed on a **space** (`_openChannel(channelId, spaceId: spaceId)` in `accord_home.dart`), which doesn't resolve DM channels.

## What's needed
- Allow the PiP to show for `spaceId == null` calls (it only renders when there's a video track anyway).
- Give it a DM-aware reopen path: tapping the PiP should reopen the full-screen DM call view (`showFullScreenVoice(channelId, spaceId: null, channelName: <dm title>)`) rather than routing through the space channel opener.

## Notes
Audio-only DM calls won't show a PiP regardless (no video track to preview) — that's consistent with the space behaviour.

PR #203, opened 2026-08-05
Broadcast mid-call mute/deafen in DM calls

### LINKED ISSUE #135: DM calls: mid-call mute/deafen not broadcast over the gateway
Follow-up from #127 / #134 (DM voice/video calls).

## Problem
For DM/group-DM calls, mid-call mute and deafen toggles are **not** broadcast to the other participant(s). The voice stack sends the *initial* self-mute/self-deaf state via the REST `voice/join`, and local toggles work for the user, but subsequent changes aren't propagated.

## Cause
`VoiceController._sendVoiceStateUpdate()` routes through the gateway `updateVoiceState` op, which is **space-scoped** (requires a non-null `spaceId` and runs space-permission checks server-side). DM calls have `spaceId == null`, so the method early-returns:

```dart
void _sendVoiceStateUpdate() {
  final channelId = state.channelId;
  final spaceId = state.spaceId;
  if (channelId == null || spaceId == null) return; // DM calls bail here
  ...
}
```

## What's needed
- Server: a way to update + route voice state for DM channels to participants (analogous to how the join now routes `voice.state_update` to `target_user_ids`). May be a server-side change in DaccordProject/accordserver.
- Client: once the server supports it, send DM voice-state updates on mute/deafen/video/screenshare toggles for `spaceId == null` channels.

## Impact
Low-ish: each side sees the others' *initial* state and join/leave; live mic/headset indicators just won't update mid-call. Video/screen-share tiles still appear because those are driven by LiveKit track events, not voice-state.

krazyjakee:
Closes #135, and the remaining half of #141 (PiP landed in #136).

## The bug

Toggling mute, deafen, video or screen share during a DM call changed nothing for the other side. Peers saw the state the call *opened* with — carried by `voice/join` — and never an update after that.

## Why it wasn't just a missing call

`_sendVoiceStateUpdate` early-returned on `spaceId == null`, which is always true in a DM. That wasn't an oversight: `updateVoiceState` took `spaceId` as a non-nullable positional and put it in the payload, and the server's `VOICE_STATE_UPDATE` was space-scoped by construction, so there was nothing to call. No REST fallback either — the voice surface is `getInfo`/`join`/`leave`/`ring`/`declineCall`/`cancelCall`/`listRegions`/`getStatus`, and only `join` carries self-mute/self-deaf.

## What changed

[accordserver#46](https://github.com/DaccordProject/accordserver/pull/46) makes `space_id` optional and resolves the scope from the channel row, routing DM voice state to the call's participants. This is the client half:

- `updateVoiceState` takes a nullable `spaceId` on both `GatewaySocket` and the `AccordClient` façade. Existing callers pass positionally, so they're unaffected.
- `_sendVoiceStateUpdate` guards only on `channelId` and passes `state.spaceId` straight through.

That's the whole change — 25 lines, most of them comments explaining why the guard was there.

## Dependency

**Needs accordserver#46 deployed to have any effect.** Against an older server the extra op is simply ignored, so this is safe to merge first; it just won't do anything until the server side lands.

## Testing

`dart analyze` clean and all 147 accordkit tests pass, including a new case asserting the DM payload — null `space_id`, real `channel_id`, flags carried. `flutter analyze --no-fatal-infos` shows only pre-existing infos, and all 942 Flutter tests pass.

The server side is covered by 9 new tests in accordserver#46, including a peer actually receiving a mid-call mute over a DM.

### DIFF
--- a/lib/features/voice/controllers/voice.dart
+++ b/lib/features/voice/controllers/voice.dart
@@ -657,10 +657,13 @@ class VoiceController extends _$VoiceController {
 
   void _sendVoiceStateUpdate() {
     final channelId = state.channelId;
-    final spaceId = state.spaceId;
-    if (channelId == null || spaceId == null) return;
+    if (channelId == null) return;
+    // `spaceId` is null during a DM call. That used to mean bailing out — the
+    // gateway op was space-scoped, so peers saw the mute/deafen state we opened
+    // the call with and nothing after it (#135). The server now resolves the
+    // scope from the channel and routes DM updates to the call's participants.
     _client?.updateVoiceState(
-      spaceId,
+      state.spaceId,
       channelId,
       selfMute: state.selfMute,
       selfDeaf: state.selfDeaf,
--- a/packages/accordkit/lib/src/core/accord_client.dart
+++ b/packages/accordkit/lib/src/core/accord_client.dart
@@ -152,9 +152,10 @@ class AccordClient {
           {Map<String, dynamic> activity = const {}}) =>
       gateway.updatePresence(status, activity: activity);
 
-  /// Sends a voice state update through the gateway.
+  /// Sends a voice state update through the gateway. [spaceId] is null for DM
+  /// and group DM calls, which have no parent space.
   void updateVoiceState(
-    String spaceId,
+    String? spaceId,
     String? channelId, {
     bool selfMute = false,
     bool selfDeaf = false,
--- a/packages/accordkit/lib/src/gateway/gateway_socket.dart
+++ b/packages/accordkit/lib/src/gateway/gateway_socket.dart
@@ -389,8 +389,13 @@ class GatewaySocket {
   }
 
   /// Sends a voice state update. [channelId] may be null to disconnect.
+  ///
+  /// [spaceId] is null for DM and group DM calls, which have no parent space —
+  /// the server resolves the scope from the channel and routes the update to
+  /// the call's participants. When non-null it must be the channel's own space;
+  /// the server ignores an update that claims a different one.
   void updateVoiceState(
-    String spaceId,
+    String? spaceId,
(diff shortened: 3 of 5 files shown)


krazyjakee:
Reviewed against the three focus areas (code quality, file size/structure, test coverage). No 3D model/rendering code is touched, so that section doesn't apply.

## Findings

**1. Missing regression coverage at the actual bug site — fixed**
The bug was `VoiceController._sendVoiceStateUpdate` (`lib/features/voice/controllers/voice.dart`) early-returning when `spaceId == null`. The PR's new test (`packages/accordkit/test/gateway_test.dart`) only proves the low-level wire payload is correct for `GatewaySocket.updateVoiceState(null, ...)` — it never exercises the guard that actually caused #135, so a future regression that reintroduces the early return would pass CI untouched.

Added `test/features/voice/voice_dm_state_update_test.dart`: wires a real `AccordClient` to a fake `GatewayConnection`, drives `VoiceController.setMute()` through a DM-connected controller state, and asserts the actual sent frame carries a null `space_id` with the channel id and flags — plus a contrasting space-call case to confirm the non-DM path still sends `space_id`. Pushed as `a4fea4d`.

**2. Code quality — no issues**
The guard change, nullable-`spaceId` threading through `AccordClient`/`GatewaySocket`, and doc comments are all correct, minimal, and consistent with the surrounding code (matches the existing nullable-`spaceId` convention already used by `VoiceController.join`).

**3. File size/structure — no issues**
All touched files are well under the 1000-line threshold (largest is `gateway_socket.dart` at 866 lines).

I wasn't able to run `dart analyze` / `dart test` locally (no Dart/Flutter SDK in this sandbox) — relied on the CI run instead, which was green on both `stable` and `beta` prior to my commit.


---

PR #205, opened 2026-08-05
ci(windows): support Certum cloud code signing via SimplySign

krazyjakee:
## Why

SignPath Foundation declined our application, so the Windows Authenticode path needs a certificate we can actually buy. [Certum Open Source Code Signing in the Cloud](https://certum.store/open-source-code-signing-on-simplysign.html) is **€49/year** — the previous "~$200–600/year" figure in our docs was for Certum's commercial OV cert, not the OSS product.

The catch: its key is non-exportable, so the existing `WINDOWS_CERT_PFX_BASE64` flow can't use it.

## How the cloud key is reached

Certum publish **no signing API and no login CLI**. The key is reachable only through SimplySign Desktop, which mounts it as a *virtual smart card* — once a session is open the certificate appears in `Cert:\CurrentUser\My` and `signtool` signs with it by thumbprint. Unlike DigiCert/SSL.com there's no `/dlib` to point at, so this is a second credential mode rather than a new vendor library.

## Changes

- **`dist/sign-windows.ps1`** — new *store* mode: `WINDOWS_CERT_SHA1` signs via `/sha1`, keeping the private key where it is. The `.pfx` path is untouched and still covers self-signed rehearsal certs and pre-2023 certificates. Also covers a USB token on a self-hosted runner.
- **`dist/simplysign-login.ps1`** *(new)* — installs SimplySign Desktop via winget, derives the TOTP from the enrolment secret, drives the login GUI, polls for the mounted certificate, exports `WINDOWS_CERT_SHA1` through `$GITHUB_ENV`.
- **`.github/workflows/release.yml`** — runs the login once per job before the two signing steps; `WINDOWS_SIGN_AVAILABLE` now accepts either credential, plus a new `SIMPLYSIGN_AVAILABLE` flag following the existing `env:`-resolves-secrets pattern.
- **`docs/release-signing.md`** — corrected pricing, recorded the SignPath denial, documented the SimplySign mechanics and their costs.

## The honest caveat

Step 3 of the login **drives a GUI with `SendKeys`**. It depends on the runner having an interactive desktop and on Certum not rearranging the login dialog's tab order, and it will break without warning at some point. Two things make that acceptable rather than reckless:

- the script exits 0 on **every** failure path, so a broken login leaves `WINDOWS_CERT_SHA1` unset and the release ships unsigned — the same outcome as owning no certificate;
- the signing steps remain `continue-on-error: true`, so a tagged release can never be lost to it.

Also worth stating plainly:

- **This collapses 2FA to 1FA.** `SIMPLYSIGN_TOTP_SECRET` is the second factor; storing it beside the account ID means anyone with repo-secret access can sign as us. Inherent to unattended signing with this product, not introduced by this PR.
- **Session lifetime is unverified** — Certum don't document it. The login runs once and both signing steps reuse the session. If sessions turn out to be shorter than a build, the installer-signing step fails first and the fix is to re-run the login before it.

## Testing

- Workflow YAML parses; step gating and ordering verified.
- TOTP + base32 ported line-for-line to Python and checked against **all five RFC 6238 SHA-1 test vectors** — all pass, base32 decoder round-trips.
- The GUI automation is necessarily **untested** until a certificate exists. First real validation is a tagged release after purchase.

## Follow-up (not in this PR)

1. Buy the cert — issued to an individual (`Open Source Developer, <name>`), needs identity-document verification, so budget a few days.
2. Set secrets `SIMPLYSIGN_USER` + `SIMPLYSIGN_TOTP_SECRET`, and repo variable `WINDOWS_TIMESTAMP_URL=http://time.certum.pl`.
3. Re-apply to SignPath Foundation once Daccord has a longer public track record — free and far more robust than this.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -19,7 +19,8 @@ name: Release
 #           → `fastlane mac dmg` signs the .app (Developer ID + hardened
 #             runtime), signs the .dmg, notarizes it with notarytool and staples
 #             the ticket. Absent → plain unsigned .dmg (Gatekeeper warns).
-#   Windows WINDOWS_CERT_PFX_BASE64 (+ WINDOWS_CERT_PASSWORD)
+#   Windows SIMPLYSIGN_USER + SIMPLYSIGN_TOTP_SECRET  (Certum cloud key), or
+#           WINDOWS_CERT_PFX_BASE64 (+ WINDOWS_CERT_PASSWORD)  (exportable .pfx)
 #           → Authenticode-signs daccord.exe, the bundled DLLs and the Inno
 #             Setup installer via dist/sign-windows.ps1. Absent → unsigned
 #             binaries (SmartScreen warns).
@@ -128,8 +129,12 @@ jobs:
       # otherwise the lane starts, fails on a missing ENV.fetch, and we take the
       # slow path to the same unsigned DMG.
       DEVID_AVAILABLE: ${{ secrets.DEVELOPER_ID_CERT_P12 != '' && secrets.ASC_KEY_P8_BASE64 != '' }}
-      # Authenticode signing of the Windows .exe / DLLs / installer.
-      WINDOWS_SIGN_AVAILABLE: ${{ secrets.WINDOWS_CERT_PFX_BASE64 != '' }}
+      # Authenticode signing of the Windows .exe / DLLs / installer. Either
+      # credential will do: an exportable .pfx, or a Certum cloud key mounted by
+      # dist/simplysign-login.ps1 (which exports WINDOWS_CERT_SHA1 for the
+      # signing steps to pick up).
+      WINDOWS_SIGN_AVAILABLE: ${{ secrets.WINDOWS_CERT_PFX_BASE64 != '' || (secrets.SIMPLYSIGN_USER != '' && secrets.SIMPLYSIGN_TOTP_SECRET != '') }}
+      SIMPLYSIGN_AVAILABLE: ${{ secrets.SIMPLYSIGN_USER != '' && secrets.SIMPLYSIGN_TOTP_SECRET != '' }}
     strategy:
       fail-fast: false
       matrix:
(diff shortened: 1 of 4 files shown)


krazyjakee:
Reviewed this PR against code quality, file structure, and test coverage, and pushed a fix commit (ab905d5).

## Findings

**1. Confirmed bug — `SIMPLYSIGN_TIMEOUT_SEC` parsing zeroed the timeout on a malformed value (Medium, fixed)**

`dist/simplysign-login.ps1` (pre-fix, line 65) did:
```powershell
[int]::TryParse($env:SIMPLYSIGN_TIMEOUT_SEC, [ref]$timeoutSec) | Out-Null
```
`Int32.TryParse`'s `out`/`ref` parameter is reset to `0` on failure, it is **not** left at its previous value. So a typo'd or malformed `SIMPLYSIGN_TIMEOUT_SEC` (e.g. `"18O"`) silently collapsed the wait for the virtual smart card to mount from the documented 180s default down to 0s — the deadline loop would run zero iterations and the login would be reported as failed before Certum's session ever had a chance to open. Verified the bug reproduces and the fix resolves it with an isolated repro script against the actual `TryParse` call.

Note the PR's own `sign-windows.ps1` (`Find-SignTool`) already does this correctly elsewhere (`if ([version]::TryParse($name, [ref]$parsed)) { $parsed } else { ... }`), so the fix brings `simplysign-login.ps1` in line with the existing pattern in this same PR rather than introducing a new one.

**2. No test coverage for the TOTP/base32 implementation (Medium, fixed)**

The PR description states the TOTP and base32 code was "ported line-for-line to Python and checked against all five RFC 6238 SHA-1 test vectors," but that verification wasn't captured anywhere in the repo — a future edit to the bit-packing or HMAC offset math would have no automated way to catch a regression. The GUI-automation half of the script genuinely can't be tested in CI (needs a real SimplySign install + interactive desktop), but the TOTP/base32 functions are pure and deterministic.

Fix: extracted `Get-Base32Secret`, `ConvertFrom-Base32`, and `Get-Totp` out of `simplysign-login.ps1` into `dist/lib/totp.ps1` (dot-sourced by the login script, no behavior change), added an injectable timestamp parameter to `Get-Totp` for determinism, and added `dist/lib/totp.Tests.ps1` (Pester) covering:
- all five RFC 6238 Appendix B SHA-1 test vectors,
- RFC 4648 base32 round-trip vectors, case-insensitivity, and padding/whitespace/hyphen tolerance,
- `otpauth://` URI secret extraction, including the missing-`secret=` error path.

All 13 cases pass locally against the extracted implementation (installed PowerShell 7.4.6 + Pester 6 in a sandbox to verify — this repo has no prior Pester/PowerShell test infra).

**3. Wired the new tests into CI**

Added a `powershell-tests` job to `.github/workflows/ci.yml` (installs Pester, runs `dist/lib/totp.Tests.ps1`, fails the job on any red test). Runs on `ubuntu-latest` alongside `analyze`; since `release.yml` calls `ci.yml` via `workflow_call` and gates on the whole thing (`needs: ci`), this also gates tagged releases, consistent with how `analyze` already gates them.

## Not changed

- **File size / structure**: no file in the diff exceeds 1000 lines (largest touched file, `release.yml`, is 1085 lines total but only ~34 lines changed by this PR and is CI config, not a natural split candidate for this diff).
- **3D model / visual changes**: not applicable — this PR only touches Windows code-signing CI plumbing and docs.
- The GUI-automation path (SendKeys against the SimplySign login dialog) is exactly as fragile as the PR already documents, and remains genuinely untestable outside a real Windows box with SimplySign installed — left as-is per the PR's own stated caveats.


---

PR #206, opened 2026-08-05
fix(ci): guard SIMPLYSIGN_TIMEOUT_SEC parsing, add TOTP unit tests

krazyjakee:
## Why

Follow-up to #205 (Certum SimplySign cloud code signing), which merged at `d2e4731` before this fix could land. Reviewing that PR (code quality / structure / test coverage) turned up one real bug and a test-coverage gap in `dist/simplysign-login.ps1`.

## Changes

- **`dist/simplysign-login.ps1`** — fixed `SIMPLYSIGN_TIMEOUT_SEC` parsing. `[int]::TryParse`'s `ref` parameter is reset to `0` on failure, not left unchanged, so a malformed timeout value silently collapsed the wait for the SimplySign virtual smart card to mount from the documented 180s default down to 0s — the deadline loop would run zero iterations and the login would be reported failed before Certum's session ever had a chance to open. Now falls back to the 180s default and warns on an unparsable value, matching the pattern `sign-windows.ps1`'s `Find-SignTool` already uses elsewhere in the same feature.
- **`dist/lib/totp.ps1`** *(new)* — extracted the pure `Get-Base32Secret` / `ConvertFrom-Base32` / `Get-Totp` functions out of `simplysign-login.ps1` so they're testable independently of the GUI automation (which needs a real SimplySign install + interactive desktop and can't run in CI). `Get-Totp` gained an injectable `-Now` parameter for determinism; production callers still default to `UtcNow`, so behavior is unchanged.
- **`dist/lib/totp.Tests.ps1`** *(new)* — Pester suite covering all five RFC 6238 Appendix B SHA-1 test vectors, RFC 4648 base32 round-trip vectors (case-insensitivity, padding/whitespace/hyphen tolerance), and `otpauth://` URI secret extraction including the missing-`secret=` error path. #205's description claimed this was checked manually against a one-off Python port; nothing was previously committed to catch a regression.
- **`.github/workflows/ci.yml`** — new `powershell-tests` job (installs Pester, runs the suite above, fails the job on any red test) alongside `analyze`. Since `release.yml` calls `ci.yml` via `workflow_call` and gates on it (`needs: ci`), this also gates tagged releases.

## Testing

- Installed PowerShell 7.4.6 + Pester in a sandbox (this repo had no prior PowerShell/Pester test infra) and ran the new suite: 13/13 passing.
- Reproduced the `TryParse` bug in isolation (malformed `SIMPLYSIGN_TIMEOUT_SEC` → old code yields `timeoutSec = 0`; fixed code warns and falls back to 180) before and after the fix.
- Parsed both modified `.ps1` files with the PowerShell AST parser to confirm no syntax errors, and ran `simplysign-login.ps1` end-to-end (with/without the SimplySign env vars set) to confirm it still exits 0 gracefully outside Windows/without SimplySign installed.

## Not in scope

No file in this diff exceeds 1000 lines; no 3D/visual changes. The GUI-automation path (`SendKeys` against the SimplySign login dialog) remains genuinely untestable outside a real Windows box with SimplySign installed, exactly as #205 already documents.

🤖 Generated with [Claude Code](https://claude.ai/code)

---
_Generated by [Claude Code](https://claude.ai/code/session_01L1EDJ82Hn7AEoavqFqv7ML)_

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -112,8 +112,8 @@ jobs:
   powershell-tests:
     name: PowerShell unit tests
     runs-on: ubuntu-latest
-    # Only the pure TOTP/base32 math (dist/lib/totp.ps1) is covered here — the
-    # GUI automation in dist/simplysign-login.ps1 needs a real SimplySign
+    # Only the pure, dependency-free helpers under dist/lib are covered here —
+    # the GUI automation in dist/simplysign-login.ps1 needs a real SimplySign
     # install and an interactive desktop, so it isn't exercised in CI.
     steps:
       - name: Checkout
@@ -128,7 +128,9 @@ jobs:
         run: |
           Import-Module Pester -MinimumVersion 5.0.0
           $config = New-PesterConfiguration
-          $config.Run.Path = 'dist/lib/totp.Tests.ps1'
+          # Discovers every *.Tests.ps1 under dist/lib, so new lib modules get
+          # CI coverage automatically instead of needing this path updated.
+          $config.Run.Path = 'dist/lib'
           $config.Run.Exit = $true
           $config.Output.Verbosity = 'Detailed'
           Invoke-Pester -Configuration $config
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -109,6 +109,30 @@ jobs:
       - name: Test
         run: flutter test
 
+  powershell-tests:
+    name: PowerShell unit tests
+    runs-on: ubuntu-latest
+    # Only the pure TOTP/base32 math (dist/lib/totp.ps1) is covered here — the
+    # GUI automation in dist/simplysign-login.ps1 needs a real SimplySign
+    # install and an interactive desktop, so it isn't exercised in CI.
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v4
+
+      - name: Install Pester
+        shell: pwsh
+        run: Install-Module -Name Pester -Force -SkipPublisherCheck -MinimumVersion 5.0.0
+
+      - name: Run tests
+        shell: pwsh
+        run: |
+          Import-Module Pester -MinimumVersion 5.0.0
+          $config = New-PesterConfiguration
+          $config.Run.Path = 'dist/lib/totp.Tests.ps1'
+          $config.Run.Exit = $true
+          $config.Output.Verbosity = 'Detailed'
+          Invoke-Pester -Configuration $config
+
   build:
     name: Build ${{ matrix.name }}
     needs: analyze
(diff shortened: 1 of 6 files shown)


krazyjakee:
Automated review pass (code quality / structure / test coverage — no 3D/visual code in this diff, so that check doesn't apply). Pushed a fix commit (`592ee8a`) to this branch.

## Findings, ranked by severity

1. **[High] `SIMPLYSIGN_TIMEOUT_SEC` fix didn't cover the same failure mode via a different input** — `dist/simplysign-login.ps1`
   The original bug in this PR's description is: an unparsable timeout value collapses the mount-wait deadline to 0s, so the wait loop runs zero iterations and the login is reported failed before the SimplySign card can ever mount. The fix added a `TryParse` success check, but `TryParse("0", ...)` and `TryParse("-30", ...)` both *succeed* — so an explicit `SIMPLYSIGN_TIMEOUT_SEC=0` (or a negative value) still produces a 0-or-negative `$timeoutSec`, `(Get-Date).AddSeconds($timeoutSec)` lands at or before "now", and the deadline loop runs zero iterations — reproducing the exact bug this PR set out to fix, just via a valid-looking value instead of a malformed one.
   **Fix:** extracted the resolution logic into `dist/lib/timeout.ps1` (`Resolve-TimeoutSeconds`, mirroring the existing `totp.ps1` extraction pattern) and require `$parsed -gt 0`, not just a successful parse, before trusting it. Zero and negative values now fall back to the 180s default with the same warning as an unparsable value.

2. **[Medium] Missing test coverage for the very logic this PR fixes** — the PR added Pester tests for the TOTP/base32 math it *extracted*, but the `SIMPLYSIGN_TIMEOUT_SEC` parsing fix itself — the actual subject of the PR — had no automated regression test; the PR description only mentions manual reproduction. That's how the zero/negative gap above went unnoticed.
   **Fix:** added `dist/lib/timeout.Tests.ps1` covering unset/blank (default, no warning), a valid positive integer, a non-numeric value, zero, and a negative value.

3. **[Low] CI test path was hardcoded to a single file** — `.github/workflows/ci.yml`'s `powershell-tests` job pointed `$config.Run.Path` at `dist/lib/totp.Tests.ps1` specifically, so the new `timeout.Tests.ps1` (or any future `dist/lib/*.Tests.ps1`) would have silently never run in CI without a workflow edit.
   **Fix:** pointed `Run.Path` at the `dist/lib` directory so Pester discovers all `*.Tests.ps1` files there automatically.

## Not changed (noted, not blocking)

- `Get-Base32Secret`'s `otpauth://` regex (`[?&]secret=([^&]+)`) doesn't URL-decode the captured value, so a secret with URL-encoded padding (`%3D`) would fail `ConvertFrom-Base32`. This is pre-existing behavior carried over from #205 (not introduced by this PR), and in practice TOTP secrets are byte lengths that don't need base32 padding, so real otpauth URIs shouldn't hit this. Flagging for awareness only.
- File sizes: nothing in the diff is remotely close to the 1000-line split threshold.

CI on the new commit: PowerShell unit tests job passed on the prior commit already; re-running now against the added `timeout.Tests.ps1`.

---

PR #211, opened 2026-08-07
fix: stop members flapping online/offline

### LINKED ISSUE #208: Members flap online/offline: gateway attempts an unsupported RESUME and never backs off
## Symptom

In the right-hand member list, some users toggle offline → online repeatedly, on a cadence of a few seconds. Rendering is pure (`AccordMemberAvatar`), so this is real state churn, not a UI bug: each blink is a genuine `presence.update` pair caused by that user's socket dropping and reconnecting.

Presence is entirely socket-lifetime driven — accordserver broadcasts `online` on IDENTIFY and `offline` on socket close — so **one socket drop on a user's client is one visible flip for everyone**. Two defects in `packages/accordkit/lib/src/gateway/gateway_socket.dart` make drops both far more frequent and far longer than they should be.

## 1. Every reconnect sends a RESUME the server ignores (~30s dead window)

`_attemptReconnect` (`gateway_socket.dart:521`) selects `GatewayState.resuming` whenever `_sessionId` is non-empty, and on HELLO sends op 3 RESUME (`:610-616`).

accordserver's pre-identify loop only inspects `op == IDENTIFY` (`accordserver/src/gateway/mod.rs:76`) and silently drops everything else. `RESUME: u8 = 3` is declared in `events.rs:19` but has no handler. The socket then idles until the 30-second identify timeout fires, receives `INVALID_SESSION{resumable:false}`, and the client waits another 1–5s plus backoff before finally IDENTIFYing.

So after any transient drop the user is broadcast offline and **stays offline for 30+ seconds**. Server-side counterpart: DaccordProject/accordserver (linked below).

**Fix (client, no server change needed):** don't attempt resume against a server that doesn't support it — clear `_sessionId` before reconnecting, or gate the resume path behind a capability flag negotiated at HELLO/READY.

## 2. Backoff never escalates for a connect → READY → die loop

`_reconnectAttempts` is reset to 0 on READY (`gateway_socket.dart:670`). A socket that reaches READY and then dies a few seconds later reconnects at the base delay — `1.0 * 2^0 + random()`, i.e. **1–2 seconds** — indefinitely, never escalating and never reaching `maxReconnectAttempts`. That is exactly the few-second cadence observed, and every cycle emits an offline+online pair to every observer.

**Fix:** reset the backoff only after a session has been *stable* (e.g. 30s past READY), not on READY itself.

## Notes

Found during a connectivity audit. Related client issues: presence lookup ignoring federated IDs, and the lack of any flap smoothing on the roster.

### LINKED ISSUE #209: Presence lookup ignores federated qualified IDs — remote members always render offline
## Problem

`accordPresenceStatus` (`lib/features/events/controllers/presence.dart:87`) is an exact-key map lookup:

```dart
presences[userId]?.status ?? 'offline'
```

accordserver broadcasts `presence.update` with a **bare** `user_id` (`accordserver/src/gateway/mod.rs:384`, `:857`), while a federated member's `AccordMember.userId` is a **qualified** `id@domain`. The keys never match, so remote members read permanently offline in the member list (`lib/features/member/views/accord_member_list.dart:113`, `:158` — which also drops them into the "Offline" bucket).

`packages/accordkit/lib/src/utils/qualified_id.dart` already has `isSameUser` / `localPart` / `qualify` for exactly this, and they're used for message authors and mentions (`message_pane.dart:439`, `:517`) — presence just never got the same treatment.

## Fix

Normalise on the way in and on lookup: key the presence cache by qualified ID (qualifying bare IDs with the connection's `homeDomain`, which `handleAccordEvents` already receives as `selfDomain`), and resolve lookups through `isSameUser` semantics rather than raw `==`.

Worth doing carefully rather than with a bare `localPart()` fallback — snowflakes are only unique per home server, so stripping the domain would let `123@b.example` collide with local `123`.

## Notes

Found during a connectivity audit. Currently a static wrong-value bug, but it becomes a *flapping* bug the moment the server starts qualifying presence IDs inconsistently.

### LINKED ISSUE #210: Roster has no presence smoothing, and our own reconnect blanks it entirely
## Problem

Two gaps in `lib/features/events/controllers/presence.dart` make presence churn maximally visible.

**No debounce.** `presence.update` goes straight to `upsert` (`lib/features/events/services/accord_event_handler.dart:154`) and straight to the member list (`accord_member_list.dart:70`). A peer's momentary socket drop is rendered immediately as a full offline transition — including re-bucketing them into the "Offline" section, so rows visibly jump.

**READY blanks the whole roster.** `seed` (`presence.dart:41`) replaces the map wholesale. That's deliberate and correct on its own terms (a stale entry must not survive a reconnect), but it means *our own* reconnect shows the entire roster as offline until re-READY — which today is 30+ seconds, see #208.

**`merge` is dead code.** `PresenceController.merge` (`presence.dart:51`) has no callers anywhere in `lib/`. There is no partial-backfill path, so a missed `presence.update` is only ever corrected by a full reconnect + READY seed. Either wire it to a real backfill (e.g. after a roster fetch) or drop it.

## Fix

Hold an offline transition for ~5–10s before rendering it, so a peer's reconnect blip never reaches the UI. Cheap, purely client-side, and independent of the server work — worth having as a backstop even once #208 and the accordserver-side fixes land.

Optionally: while disconnected, keep rendering the last-known presence dimmed rather than flipping everyone to offline.

## Notes

Found during a connectivity audit.

krazyjakee:
Fixes the roster flap end to end: the gateway defects that made peers' sockets drop often and stay down for 30+ seconds, the lookup bug that pinned federated members at offline, and the lack of any smoothing that made every blip visible.

Closes #208, closes #209, closes #210.

## #208 — gateway (`packages/accordkit`)

**RESUME is capability-gated.** accordserver's pre-identify loop only inspects `op == IDENTIFY` and silently drops everything else, so the op 3 we sent on every reconnect went unanswered until the server's 30-second identify timeout fired — the user was broadcast offline that entire time. `_resumeSupported` now stays false until a HELLO or READY advertises `resumable: true` (or `"resume"` in `capabilities`); otherwise the stale session is dropped up front and the handshake goes straight to IDENTIFY. The RESUME path is kept intact behind the flag (and exposed as `GatewaySocket.resumeSupported`) for when the server implements it — no client change needed then.

**Backoff escalates on a READY → die loop.** The budget is no longer reset on READY; READY/RESUMED stamp the session start and the budget is credited on close only if the session survived `stableSessionThreshold` (30s, injectable). A session dying at 5s old now backs off 1s → 2s → 4s → 8s instead of looping at 1–2s forever. A new `now` constructor seam sits alongside the existing `sleep`/`random` ones, so the delay sequence is asserted deterministically rather than with real timers.

## #209 — presence keyed by qualified ID

The cache was an exact-key map, but the wire carries a bare `user_id` while a federated member's `AccordMember.userId` is qualified — the keys never matched. Presences now live in a `PresenceMap` that carries the connection's home domain and normalises **both** writes and lookups through `qualify()`, i.e. `isSameUser` semantics expressed as a map key. `123` and `123@a.example` resolve to the same entry; `123@b.example` stays distinct, which a bare `localPart` fallback would not. `handleAccordEvents` threads its existing `selfDomain` into `upsert`/`seed`.

## #210 — smoothing, seeding, dead code

- Offline transitions are held for a grace window (8s) before reaching state. Going non-offline is never delayed and cancels a pending hold, so a drop-and-reconnect inside the window emits **no** state change at all — asserted with a listener test.
- Users absent from a READY re-seed go through the same hold, so our own reconnect keeps rendering last-known presence instead of blanking the roster.
- `merge` is dropped — no callers, and with no presence REST endpoint in accordkit there is no backfill path to wire it to.

Not done (listed as optional in #210): rendering last-known presence *dimmed* while disconnected. That's UI work across the avatar/roster widgets, and the graced re-seed already covers the reported symptom.

## Notes for the reviewer

A genuinely flapping socket can now **exhaust** `maxReconnectAttempts`, which it previously never could. The in-app Retry buttons call `ensureConnected()`, but `AccordAuth.ensureConnectedAll()` still has no caller — there is no app-lifecycle/foreground hook wiring it up. Worth a follow-up issue.

## Testing

- `dart analyze` + `dart test` in `packages/accordkit` — clean, 151 passing, including new `re-identifies against a server that never advertised resume` and a `backoff` group asserting the escalation and the stable-session credit.
- `flutter analyze --no-fatal-infos` — no errors.
- `flutter test` — green except `test/features/voice/voice_dm_state_update_test.dart`, which fails identically on `master` (audioplayers `MissingPluginException`) and is untouched here.
- `test/features/events/presence_controller_test.dart` gains federation and smoothing groups (31 tests).

### DIFF
--- a/lib/features/events/controllers/presence.dart
+++ b/lib/features/events/controllers/presence.dart
@@ -1,14 +1,67 @@
+import 'dart:async';
+
 import 'package:accordkit/accordkit.dart';
 import 'package:bonfire/features/server/controllers/connections.dart';
+import 'package:flutter/foundation.dart';
 import 'package:flutter_riverpod/flutter_riverpod.dart';
 import 'package:riverpod_annotation/riverpod_annotation.dart';
 
 part 'presence.g.dart';
 
-/// One connection's per-user presence cache, keyed by user ID and scoped to
-/// [serverKey] (`userId@baseUrl`) — the same scoping [ReadStateController]
-/// uses, and for the same reason: snowflake IDs are minted per server, so a
-/// single global map lets two servers' users collide.
+/// One connection's presence cache, keyed by **qualified** user ID.
+///
+/// The gateway broadcasts `presence.update` with a bare `user_id`, while a
+/// member seen through a federated space carries a qualified `id@domain` — so
+/// an exact-key lookup never matched and those members read permanently offline
+/// (#209). Both sides are normalised through [qualify] here instead: bare IDs
+/// are suffixed with the connection's [homeDomain], already-qualified IDs are
+/// left alone. That is [isSameUser] semantics expressed as a map key, and it
+/// keeps a remote `123@b.example` from colliding with a local `123` — which a
+/// bare [localPart] fallback would not, since snowflakes are only unique per
+/// home server.
+@immutable
+class PresenceMap {
+  const PresenceMap({this.byUser = const {}, this.homeDomain = ''});
+
+  /// Presences by qualified user ID (or by raw ID while [homeDomain] is still
+  /// unknown — [withDomain] re-keys them as soon as it is).
+  final Map<String, AccordPresence> byUser;
+
+  /// The connection's home domain (`a.example`), empty until the first write
+  /// supplies it.
+  final String homeDomain;
+
+  bool get isEmpty => byUser.isEmpty;
+  bool get isNotEmpty => byUser.isNotEmpty;
+  int get length => byUser.length;
+
+  /// The cache key for [userId]. [qualify] is idempotent, so passing an
+  /// already-qualified ID is a no-op.
+  String keyFor(String userId) =>
+      homeDomain.isEmpty ? userId : qualify(userId, homeDomain);
+
+  /// The presence for [userId], given bare or qualified, or null when none has
+  /// been received.
+  AccordPresence? operator [](String userId) => byUser[keyFor(userId)];
+
+  /// This map re-keyed for [domain]. Returns `this` when nothing changes; an
+  /// empty [domain] means "not known yet" and never un-qualifies keys.
+  PresenceMap withDomain(String domain) {
+    if (domain.isEmpty || domain == homeDomain) return this;
+    return PresenceMap(
+      byUser: {
+        for (final entry in byUser.entries)
+          qualify(entry.key, domain): entry.value,
+      },
+      homeDomain: domain,
+    );
+  }
+}
+
+/// One connection's per-user presence cache, scoped to [serverKey]
+/// (`userId@baseUrl`) — the same scoping [ReadStateController] uses, and for the
+/// same reason: snowflake IDs are minted per server, so a single global map lets
+/// two servers' users collide.
 ///
 /// Every connection (active *or* background) seeds this from its gateway READY
 /// payload's `presences` array and keeps it current from `presence.update`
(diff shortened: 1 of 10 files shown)


krazyjakee:
## Automated review

Scoped to code quality, file structure, and test coverage (no 3D/rendering code is touched by this PR, so that check doesn't apply). CI (`Analyze & Test` on both stable and beta) is green as of this review.

### Findings

No issues that warrant a code change. Summary of what was checked:

**Code quality** — `PresenceMap`'s qualify/keyFor design correctly handles the bare-vs-qualified-ID mismatch (#209) without ever un-qualifying keys or colliding cross-server snowflakes. `GatewaySocket._handshakeState()`/`_advertisesResume()`/`_creditStableSession()` cleanly separate "can we RESUME" from "should the backoff budget reset," matching the failure modes described in #208. Naming, structure, and doc comments are consistent with the surrounding codebase's conventions (see `packages/accordkit/CLAUDE.md`).

**File size** — largest touched file is `gateway_socket.dart` at 933 lines (mostly pre-existing stream-getter boilerplate untouched by this diff); nothing crosses the 1000-line threshold, so no split is warranted.

**Test coverage** — new logic is well covered: `gateway_test.dart` gains dedicated cases for RESUME gating via `resumable`/`capabilities`, backoff escalation on a flapping session, and budget credit on a stable session; `presence_controller_test.dart` gains a federation group (bare-vs-qualified keying, no cross-domain collision, pre-domain-known re-keying) and an offline-smoothing group (grace window, blip suppression, repeated-offline non-extension, disposal cancelling pending timers). I traced a theoretical edge case — a `seed()`-scheduled offline hold (payload `null`) can win a race against a same-key `upsert()` offline hold that arrives before the first fires, since `_holdOffline` no-ops when a hold is already pending — but the two outcomes are behaviorally identical (`accordPresenceStatus` reads "offline" from a removed entry exactly like the missing status field), so it's not a real bug and not worth a test or a fix.

No commit pushed — there was nothing to fix.

---

PR #212, opened 2026-08-07
ci: dispatch an accordwebsite rebuild after a release

krazyjakee:
## Problem

`daccord.gg/chat/` was serving the **v0.2.6** web client — two releases behind — while v0.2.9 had been out since 2026-08-05.

The website image bakes in this repo's latest `daccord-web.zip` release asset and serves it under `/chat/`. accordwebsite's build workflow already listens for a `web-export-updated` `repository_dispatch`, but **nothing has ever sent one**. So `/chat/` only refreshed when an unrelated commit happened to land on accordwebsite — the last one was 2026-07-09, which is exactly where the live client was pinned.

Verified by md5-matching the live `flutter_bootstrap.js` against each release's export:

| | `flutter_bootstrap.js` |
|---|---|
| live `daccord.gg/chat/` | `dc1fbe3dc30c` |
| v0.2.6 | `dc1fbe3dc30c` ← |
| v0.2.7 | `d954feea3516` |
| v0.2.9 | `64c852fc05a1` |

## Change

Adds a `website-rebuild` job that fires the dispatch after `release:` completes on a tag push, closing the loop.

Gated on a new `ACCORDWEBSITE_DISPATCH_TOKEN` secret. Consistent with every other optional secret in this workflow, a missing token **warns and skips** rather than failing a tagged release — the website just keeps serving the previous client until someone rebuilds by hand. `GITHUB_TOKEN` can't be used here; it's scoped to this repo and can't dispatch to another one.

## Before this merges

Add repo secret `ACCORDWEBSITE_DISPATCH_TOKEN` — a PAT (or GitHub App token) with `contents: write` on `DaccordProject/accordwebsite`. Until then the job is a no-op warning.

## Out of scope

v0.2.9 has already been shipped to `/chat/` manually: accordwebsite build [31154997290](https://github.com/DaccordProject/accordwebsite/actions/runs/31154997290) published a `:latest` verified to contain the v0.2.9 export with `<base href="/chat/">`. That image still needs redeploying and a Cloudflare `/chat/*` purge.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -633,6 +633,42 @@ jobs:
           files: artifacts/*
 
   # ---------------------------------------------------------------------------
+  # Rebuild daccord.gg. The website image bakes in the `daccord-web.zip` asset
+  # from *this* repo's latest release and serves it at daccord.gg/chat/, so a new
+  # release does nothing for web users until accordwebsite rebuilds. Its workflow
+  # already listens for a `web-export-updated` repository_dispatch; nothing was
+  # sending one, so /chat/ sat on v0.2.6 for the whole v0.2.7–v0.2.9 window and
+  # only ever refreshed when someone happened to push to accordwebsite.
+  #
+  # Needs ACCORDWEBSITE_DISPATCH_TOKEN: a PAT (or GitHub App token) with
+  # `contents: write` on DaccordProject/accordwebsite. GITHUB_TOKEN can't be used
+  # — it's scoped to this repo and can't dispatch to another one.
+  # ---------------------------------------------------------------------------
+  website-rebuild:
+    name: Rebuild daccord.gg /chat/
+    needs: release
+    runs-on: ubuntu-latest
+    # As elsewhere in this workflow, a missing secret must never fail a tagged
+    # release: without the token this job is skipped, and the website just keeps
+    # serving the previous client until someone rebuilds it by hand.
+    if: ${{ success() && github.event_name == 'push' && github.repository == 'DaccordProject/daccord' }}
+    steps:
+      - name: Notify accordwebsite of the new web export
+        env:
+          GH_TOKEN: ${{ secrets.ACCORDWEBSITE_DISPATCH_TOKEN }}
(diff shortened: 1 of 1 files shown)


krazyjakee:
Reviewed this PR (single-file CI workflow change, `.github/workflows/release.yml`) against code quality, file structure, and test coverage. Findings, ranked by severity:

**1. [Fixed] Script-injection footgun — `github.ref_name` interpolated directly into `run:`**
`${{ github.ref_name }}` was spliced straight into the shell script (both the warning message and the `gh api -f 'client_payload[tag]=...'` call). The tag name is attacker-influenceable — anyone able to push a tag controls this value — and GitHub's own hardening guidance is to never splice untrusted/semi-trusted expression context directly into a `run:` block, since a maliciously crafted ref (e.g. containing `` ` `` or `$(...)`) executes as shell. Fixed in `6fd434a` by routing it through an `env: TAG: ${{ github.ref_name }}` and referencing `$TAG` in the script instead.

**2. [No action taken, flagged for awareness] `release.yml` is 1129 lines, over the 1000-line guideline**
This predates the PR (already ~1093 lines before this change) and the PR only adds a self-contained 36-line job at the end of the file. Splitting this workflow (e.g. into reusable workflows per platform) is a real option, but it's a release pipeline that only executes on a tag push — I have no safe way to exercise it in this session to verify a refactor doesn't break signing/publish. Recommend doing that as its own follow-up PR with a test tag, rather than bundling a risky structural change into this small, focused dispatch fix.

**3. No issues found**
- The new job's `needs`/`if: success() && ...` pattern matches the existing convention used by `ios-appstore`/`macos-appstore`/`android-playstore` (lines 681, 793, 932).
- Missing-secret handling (warn + `exit 0` rather than failing the release) is consistent with every other optional-secret step in this file.
- No test coverage gap to fill — this repo's test suite (`flutter test`) doesn't cover GitHub Actions workflows, and there's no 3D/visual surface touched by this change.

Pushed the fix to this PR's branch (`ci/dispatch-website-rebuild`).

---

Issue #218, opened 2026-08-15
Joining a space while connected delivers no gateway events until reconnect


krazyjakee:
## Symptom

Join a space while the app is running and the space is dead: no new messages arrive, no typing indicators, no member updates, no channel changes. Everything works after an app restart (or any gateway reconnect).

Same for a space you just *created* — the creator's own live session doesn't receive events for it either.

## Cause

The gateway session snapshots the user's space memberships once, at IDENTIFY, and never updates them for the life of the connection — `src/gateway/mod.rs`:

```rust
let space_ids: HashSet<String>;      // note: not `mut`
...
space_ids = db::spaces::list_space_ids_for_user(&state.db, &user_id).await…
```

Fan-out is then filtered against that snapshot (`classify_broadcast` → `(None, Some(sid)) => space_ids.contains(sid)`), so a space added after IDENTIFY matches nothing. `muted_channel_ids` next to it *does* get refreshed mid-session (`Delivery::RefreshMutes`); `space_ids` has no equivalent. The comment at the load site says memberships "are reloaded on RESUME too — they can change while a client is away", which covers the away case but not the joined-while-connected case.

The client doesn't compensate: `accord_discovery.dart:244` joins over REST and upserts the space into `spacesControllerProvider` without reconnecting the gateway. Same in `add_server_dialog.dart:124` and `accord_login.dart:176`.

## Evidence

Found by the new integration suite (#217). With the gateway opened *before* the join, all five gateway fan-out assertions time out at 10s; opening it after the join makes all five pass. Reproduced against a locally built accordserver, both message and non-message events (`message.create`, `message.update`, `message.delete`, `typing.start`, `channel.create`).

## Fix options

- **Server (preferred):** make `space_ids` mutable and update it when the session's user joins/leaves a space — either by having the join/leave handlers push a session-scoped refresh (the `Delivery::RefreshMutes` pattern already does exactly this for mutes), or by re-reading memberships when a `member.join`/`member.leave` broadcast names this user.
- **Client (workaround):** reconnect the gateway after a successful join. Cheap, but it drops and re-establishes the whole session for every join, and does nothing for other clients.

A regression test is committed and skipped, pointing at this issue — unskip it when fixed.

krazyjakee:
Server-side fix tracked at DaccordProject/accordserver#52 — `space_ids` needs to be mutable and refreshed on membership change, mirroring the existing `Delivery::RefreshMutes` path.

[...]

Issue #217, opened 2026-08-15
Testing: no end-to-end coverage of the client against a real accordserver


krazyjakee:
## Problem

The client has 89 Dart test files, all pure unit/widget tests with no server behind them. Every bug class that lives in the seam between the client and `accordserver` — gateway event wiring, cache mutators, controller state transitions, protocol drift — is currently uncovered, and that is where most of the bugs we hit in practice actually live.

The server side is well covered (`accordserver/tests/`: ~10.4k lines across `e2e.rs`, `security.rs`, `federation.rs`, `ws.rs`, with a `TestServer` fixture). What's missing is anything that exercises **this client against a real server**.

## Plan

Three layers, in descending bugs-per-hour order. Layer 1 carries most of the value; the server fixture it introduces is shared by all three.

### Layer 1 — headless client↔server integration tests

A new top-level `integration/` directory (excluded from the default `flutter test` run), containing a fixture that boots a real `accordserver` and tests that drive `AccordClient` and the Riverpod cache controllers directly, no widgets.

- Server fixture resolves in order: `ACCORD_TEST_SERVER_URL` (use a running server) → `ACCORD_SERVER_BIN` → auto-detected `../accordserver/target/{release,debug}/accordserver` → `ghcr.io/daccordproject/accordserver:latest` via docker → skip with a clear message.
- Fresh temp SQLite DB + CDN dir per server, ephemeral port, `ACCORD_TEST_MODE=1` (bypasses the LiveKit requirement so voice-state paths are testable without a real SFU).
- Two clients in one test process: send from A, assert B's gateway stream and the `accord_messages` cache both reflect it.

Covers: auth/session, space + channel CRUD, message send/edit/delete fanout, member join/leave, reactions, typing, presence, and REST↔gateway consistency.

### Layer 2 — UI e2e via `integration_test/`

Real widget tree on Linux desktop against the same server fixture. Catches routing, rendering, and state-not-updating bugs that layer 1 can't see. Needs xvfb in CI.

### Layer 3 — multi-instance scenario tests

Drive two real app instances through the MCP control surface the app already exposes (`lib/features/developer/services/mcp_tools.dart` — ~40 tools over loopback JSON-RPC: `navigate`, `send_message`, `join_voice_channel`, `get_current_state`, …). This is the Flutter equivalent of the Godot client's `client_test_api.gd` and the approach from #45. Best fit for voice, presence, and typing, which are unreachable from a single process.

## Known constraint

[...]

PR #219, opened 2026-08-15
test: add client↔accordserver integration suite

krazyjakee:
Layer 1 of #217: end-to-end coverage of this client against a **real accordserver**.

All 89 existing Dart test files run against mocks, so the client↔server seam — request shapes, gateway event names, and the `handleAccordEvents` → cache wiring — has had no coverage. This adds a top-level `integration/` suite (not picked up by `flutter test`, which only walks `test/`).

## What's here

- **`support/accord_test_server.dart`** — boots a real server per test file with a fresh temp SQLite DB, fresh CDN dir, and ephemeral port. Resolves one from `ACCORD_TEST_SERVER_URL` → `ACCORD_SERVER_BIN` → a sibling `../accordserver` build → the `ghcr.io` image under docker. When none are available every suite **skips with an explanation** instead of failing.
- **`support/harness.dart`** — registers real accounts, opens real gateways, and can wire a client into a `ProviderContainer` (overriding `accordAuthProvider`) so the app's own controllers and gateway event handler run against live traffic. Plus `waitForEvent` / `waitForState`, which fail with a timeout rather than hanging.
- **`accordkit_protocol_test.dart`** — protocol seam: auth, space + channel CRUD, message create/list/edit/delete, and gateway fan-out for `message.create`, `message.update`, `message.delete`, `typing.start`, `channel.create`.
- **`messaging_cache_test.dart`** — controller seam: a message sent by another user must land in this client's `accordMessagesControllerProvider` cache with no refetch, and edits/deletes must apply to it.

15 tests, ~10s wall clock against a local debug server build.

## It already found a bug

First run: all five gateway fan-out assertions timed out. Cause is #218 — the server snapshots a session's space memberships at IDENTIFY and never refreshes them, so **a space joined while connected receives no gateway events until reconnect**. The client doesn't compensate (`accord_discovery.dart:244` joins over REST without reconnecting). A regression test is committed `skip:`-ped, pointing at #218.

## CI

Runs as a separate `integration` job against the server image, `continue-on-error: true` — a registry outage or server-side regression shouldn't wedge the merge gate.

## Notes

- Registration is rate-limited server-side to 5/IP/15min and `ACCORD_TEST_MODE` doesn't bypass it, so accounts are created once in `setUpAll` and the harness throws with a clear message at the budget. A one-line accordserver change would lift this.
- `test/features/voice/voice_dm_state_update_test.dart` fails on master, unrelated to and untouched by this PR.
- Layers 2 (UI e2e via `integration_test/`) and 3 (multi-instance via the in-app MCP surface) remain open in #217.

Refs #217. Found #218.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -5,9 +5,10 @@ name: CI
 # - `analyze`  : code generation + static analysis + unit tests (the gate).
 #                Runs on every push/PR — this is what blocks merges.
 # - `integration`: client↔accordserver end-to-end tests against the ghcr.io
-#                server image, on the self-hosted runners. Non-blocking
-#                (continue-on-error) so a server-side problem can't wedge
-#                merges; see integration/README.md.
+#                server image. GitHub-hosted — it needs docker, which the
+#                self-hosted runners don't have (see the job's note).
+#                Non-blocking (continue-on-error) so a server-side problem
+#                can't wedge merges; see integration/README.md.
 # - `ui-e2e`   : the real app driven on Linux under xvfb against the same
 #                server image. GitHub-hosted (see the job's own note) and also
 #                non-blocking; see integration_test/README.md.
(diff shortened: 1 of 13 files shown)


krazyjakee:
## Review

Reviewed against the requested checklist (code quality, file size/structure, test coverage, 3D/visual — n/a here). No Flutter/Dart SDK is available in this environment, so I couldn't run `flutter analyze`/`flutter test` directly; instead I cross-checked every accordkit/app API the new files call (constructors, method signatures, stream element types, Hive box names, `handleAccordEvents`/`AccordAuth`/`ProfileStore` shapes — ~25 call sites) against the real source in `packages/accordkit` and `lib/`. All of it matches.

### Findings

**1. Medium — unawaited async cleanup in `IntegrationHarness` (fixed)**

`AccordClient.dispose()` does real async work (cancels subscriptions, closes the WebSocket with a proper close handshake, closes voice-manager stream controllers), but two call sites fired it without awaiting:
- `harness.dart`'s `dispose()` loop called `client.dispose()` unawaited, right before `await _server?.stop()` — so the server (or its docker container) could be torn down mid-cleanup, and any error from `dispose()` would become an unhandled async error instead of failing the test.
- `newAccount()`'s register-only client (`restOnly.dispose()`) had the same issue, and the client was never removed from `_clients`, so the teardown loop would later call `logout()`/`dispose()` on it a second time.

Fixed in `995a227e`: both calls are now `await`ed, and the register-only client is removed from `_clients` once manually disposed.

**2. Low — `_freePort()` bind-close-reuse has a theoretical TOCTOU race**

`accord_test_server.dart`'s `_freePort()` binds an ephemeral port, closes it, and hands the number to the spawned binary/container. Between close and the server actually binding, another process could grab the same port. This is a standard, widely-used test idiom and not worth restructuring, but worth knowing about if the integration job ever shows sporadic "did not become healthy" boot failures in CI.

### Not flagged

- File sizes: largest new file is `harness.dart` at 356 lines — nowhere near the 1000-line split threshold.
- Test coverage: the PR's entire purpose is adding coverage for a previously-untested seam; no gaps identified in what it covers. The `skip:`-ped regression test for #218 is clearly documented and linked.
- No 3D/visual code is touched by this diff.


---

krazyjakee:
Updated: this branch now carries **layers 1 and 2** of #217, plus the #220 fix it depends on.

**Layer 2 (#221) — `integration_test/`**, added in fffb15e: drives the real app shell (`MainWindow`, real router, real screens) on a real Linux device against the same server fixture. Three scenarios, all green in ~6s:

- the space shell renders for the signed-in user (channel in the sidebar, composer addressing it, own name in the member list)
- a channel's existing history renders
- a message another account sends appears live

The only new seam is `IntegrationHarness.scopeFor`, which wraps a widget in a `ProviderScope` reporting an account as signed in. Runs in CI as a third non-blocking job, on Linux under xvfb.

**The one that cost real time**, written up in `integration_test/README.md`: with the self-updater left enabled, the app reaches the real release endpoint mid-test and stages a download, and the throw lands *after* the test body finishes — so the runner blames whichever test is running. It presented as "the first test in the process fails, later ones pass" and looked exactly like a message-rendering bug. Two scenarios were nearly shipped skipped because of it. `setupHive` now seeds `autoUpdateCheck: false`.

**Also on this branch:** 995a227 (a cloud agent's fix — `AccordClient.dispose()` is async and was fire-and-forget) is merged in.

**Merge order:** #223 (the SoundManager fix) first — it's merged into this branch, so this PR's harness relies on it.

Suites: 960 unit / 15 integration (+1 skipped, pending accordserver#55) / 3 UI e2e.

PR #223, opened 2026-08-15
fix(sound): silence SoundManager under flutter test

### LINKED ISSUE #220: SoundManager isn't test-seamed: platform-channel failures leak into unit tests
## Symptom

`test/features/voice/voice_dm_state_update_test.dart` fails on master:

```
MissingPluginException(No implementation found for method getTemporaryDirectory
on channel plugins.flutter.io/path_provider)
  package:bonfire/features/notifications/services/sound.dart 79:5   SoundManager.play
  package:bonfire/features/voice/controllers/voice.dart 307:18      VoiceController.setMute
  test/features/voice/voice_dm_state_update_test.dart 134:56
```

It's also **nondeterministic**: run the file alone and both tests in the group fail; run the full suite and only one does. The runner reports "This test failed after it had already completed", so the failure arrives from an un-awaited async gap and lands on whichever test happens to be running.

## Cause

`SoundManager` isn't test-seamed. It's a top-level singleton (`final soundManager = SoundManager._()` in `lib/features/notifications/services/sound.dart`) that eagerly builds a pool of `AudioPlayer`s and reaches `path_provider` on `play()`. Any code path that chimes — `VoiceController.setMute` here, `handleAccordEvents` on an incoming message — drags a platform channel into a plain `flutter test` VM, where there's no implementation.

There's no injection point: it can't be overridden, and the only thing gating it is the `soundsEnabled` client setting read from Hive.

## Suggested fix

Give the sound layer a seam, in rough order of preference:

1. Put `soundManager` behind a Riverpod provider so tests can override it with a no-op, matching how the rest of the app is wired.
2. Failing that, have `SoundManager` no-op when `!kIsWeb && Platform.environment.containsKey('FLUTTER_TEST')`, or expose a `SoundManager.silent()` the tests install.

Either way the pool should be built lazily rather than in the field initializer.

## Notes

Found while building the integration suite (#217) — the same root cause bites there, worked around by seeding `soundsEnabled: false` into the settings box before wiring a container (`integration/support/harness.dart`). That workaround is load-bearing and should be removable once this is fixed.

krazyjakee:
Closes #220. Fixes the `voice_dm_state_update_test.dart` failure on master.

## The problem

`SoundManager` is a global singleton that reaches `audioplayers` — and the `path_provider` behind it — on every play. Neither has a platform implementation in a plain `flutter test` VM, so any code path that chimes (a mute toggle, an incoming message) drags a `MissingPluginException` into the test process.

It doesn't fail cleanly. The throw arrives from an un-awaited async gap, so the runner reports *"this test failed after it had already completed"* and pins it on whichever test is running at the time — `voice_dm_state_update_test.dart` failed **both** its tests when run alone and only **one** in a full-suite run. That's a failure that can migrate to innocent tests.

## The fix

`SoundManager.silent` defaults to true when `FLUTTER_TEST` is set, and short-circuits `play`, `startRingtone`, `init`, and `dispose`. The player pools are now `late` so a silent manager never constructs one.

Kept as a plain mutable static rather than a Riverpod migration: the singleton is deliberate and documented (no observable state, call sites just fire one-shots), the guard fixes every current *and future* call site at once, and a test that wants real audio can set it back.

`universal_io`'s `Platform` rather than `dart:io` — this file is compiled for web, where `dart:io` isn't available; universal_io reports an empty environment there, which is the right answer for this check.

## Tests

Three cases in `sound_test.dart` pinning the guard: `silent` defaults true under test, `play`/`startRingtone`/`stopRingtone` complete without touching a player, `init` no-ops.

Full unit suite: 960 passing, 0 failing (was 956 passing / 1 failing).

## Note for #219

That PR's harness works around this by seeding `soundsEnabled: false` into the settings box before wiring a container. Once this lands the workaround is dead weight; #219 already drops it and relies on this instead, so this should merge first.

### DIFF
--- a/lib/features/notifications/services/sound.dart
+++ b/lib/features/notifications/services/sound.dart
@@ -1,5 +1,8 @@
 import 'package:audioplayers/audioplayers.dart';
 import 'package:flutter/widgets.dart';
+// `dart:io` isn't available on web; universal_io's Platform reports an empty
+// environment there, which is the right answer for this check anyway.
+import 'package:universal_io/io.dart' show Platform;
 
 /// Plays short UI sound effects (message sent/received, mentions). The Dart
 /// analogue of the reference client's `SoundManager` autoload
@@ -33,14 +36,28 @@ class SoundManager {
 
   static const _poolSize = 4;
 
-  final List<AudioPlayer> _pool =
+  /// Silences every playback path, and stops any [AudioPlayer] from being
+  /// constructed at all.
+  ///
+  /// Defaults to true under `flutter test`, where `audioplayers` and the
+  /// `path_provider` it reaches for have no platform implementation: a chime
+  /// fired from code under test (a mute toggle, an incoming message) would
+  /// otherwise throw `MissingPluginException` from an un-awaited async gap and
+  /// fail whichever test happened to be running (#220). Settable so a test can
+  /// opt back in.
+  static bool silent = Platform.environment.containsKey('FLUTTER_TEST');
+
+  /// Lazy so a silent manager never builds a player. `audioplayers` reaches the
+  /// platform on first use, not construction, but building four of them plus a
+  /// ringtone player in a test process is pointless either way.
+  late final List<AudioPlayer> _pool =
       List.generate(_poolSize, (_) => AudioPlayer());
   int _next = 0;
   bool _initialized = false;
 
   /// Dedicated looping player for the incoming/outgoing call ringtone, separate
   /// from the one-shot SFX pool so it can sustain across a ring.
-  final AudioPlayer _ringPlayer = AudioPlayer();
+  late final AudioPlayer _ringPlayer = AudioPlayer();
   bool _ringing = false;
 
   AppLifecycleListener? _lifecycle;
@@ -56,7 +73,7 @@ class SoundManager {
   double volume = 1.0;
 
   void init() {
-    if (_initialized) return;
+    if (silent || _initialized) return;
     _initialized = true;
     for (final player in _pool) {
       player.setReleaseMode(ReleaseMode.stop);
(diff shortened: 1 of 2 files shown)


krazyjakee:
Automated review — code quality, structure, and test coverage.

**Note on branch:** this session's designated branch was `claude/eager-curie-478bop`, but no PR existed for it (and the branch itself no longer exists on `origin`). Its tip commit is identical to this PR's head (`4064c21`), so the fix below was pushed directly to `fix/sound-manager-test-seam`, the PR's actual branch.

## Findings

1. **Test coverage gap — `dispose()` untested while silent** (low severity, fixed)
   The PR description explicitly lists `play`, `startRingtone`, `init`, and `dispose` as the four methods `SoundManager.silent` short-circuits, but `sound_test.dart` only asserted no-op behavior for the first three. `dispose()` is exactly the risky path this PR touches: `_pool`/`_ringPlayer` became `late final`, so an unguarded `dispose()` would force their construction (and reach the `audioplayers` platform channel) the same way an unguarded `play()` used to. Added a `dispose is a no-op while silent` test mirroring the existing `init`/`play` cases in `56eb04c`.

## Reviewed, no issues found

- **Guard ordering** — `silent` is checked before any access to the `late final _pool`/`_ringPlayer` fields in `init`, `play`, and `startRingtone`, so a silent manager genuinely never constructs an `AudioPlayer`. `stopRingtone` has no explicit `silent` check, but relies on `_ringing` staying `false` (since `startRingtone` returns before setting it) — correct, though slightly implicit.
- **`FLUTTER_TEST` via `universal_io`** — correct choice given this file compiles for web (`dart:io` isn't available there); consistent with the PR's own rationale.
- **File size / structure** — `sound.dart` is 193 lines, well under the 1000-line split threshold. No structural changes needed.
- **Scope** — no 3D model/rendering code touched; visual review not applicable.

No other correctness issues found. This is a small, well-targeted fix with a clear root-cause writeup; the one coverage gap has been closed.

---

Issue #222, opened 2026-08-15
Testing layer 3: multi-instance scenario tests over the in-app MCP surface


krazyjakee:
Layer 3 of #217. Layer 1 (client↔server, headless) landed in #219.

## What

Drive **two real app instances** against one server and assert on both sides of an interaction.

Voice, presence, and typing are the features where the interesting bugs are inherently multi-process: who hears whom, who sees the speaking indicator, what happens when one side drops. A single test process can't see them, and layers 1–2 don't either.

## The control surface already exists

`lib/features/developer/services/mcp_tools.dart` (~996 lines) exposes ~40 tools over a loopback JSON-RPC/MCP server (`mcp_server_io.dart`, `HttpServer.bind`), enabled from developer settings: `navigate`, `open_dm`, `open_settings`, `open_voice_view`, `join_voice_channel`, `leave_voice`, `send`/`edit_message`, `delete_message`, `add_reaction`, `list_messages`, `list_members`, `get_current_state`, plus role/ban/kick/moderation tools.

This is the Flutter equivalent of the Godot client's `client_test_api.gd`, and it means no synthetic input is needed — no xdotool, no fake clicks. A test harness launches two app instances on different MCP ports and speaks JSON-RPC to both.

## Scope

- A launcher that starts N app instances with developer mode + MCP enabled on distinct ports, pointed at one server-fixture instance.
- A thin Dart client for the MCP surface.
- First scenarios: A joins a voice channel, B sees A in it; A mutes, B sees the mute; A types, B sees the indicator; A goes offline, B's presence updates; A sends a DM, B's DM list surfaces it.

## Notes

- Prior art: #45 (merged, Godot-era, GDScript) did multi-instance app testing with login and input simulation. The approach carries over; the code doesn't.
- Voice scenarios need either a real LiveKit (`docker-compose.yml` runs one in dev mode) or `ACCORD_TEST_MODE=1`, which relaxes the SFU requirement for the signalling paths.
- Heaviest of the three layers — worth doing after layer 2 proves the fixture holds up under a real app process.

krazyjakee:
## Scouting notes (not started)

Layers 1 (#219) and 2 (#221) are done. This one is not — recording what's now known so it doesn't need re-discovering.

[...]

PR #225, opened 2026-08-16
test(e2e): add the multi-instance suite (layer 3)

krazyjakee:
Layer 3 of #217. Completes the three-layer plan: layers 1 and 2 landed in #219.

Two real app processes against one server, driven through the MCP server the app already exposes in Developer Mode (`mcp_tools.dart`). Layers 1 and 2 both run inside a single process, so they can only check what one client believes — this asserts that what A did is what B sees. No synthetic input: no xdotool, no fake clicks.

**Six scenarios, ~9s** once the app is built: both clients come up signed in, navigation moves the real UI, a message crosses in each direction, each client sees the other in the member list, and an edit and a delete propagate. Five consecutive clean runs locally.

## How an instance is made

Throwaway `HOME`, with `accord-settings` (Developer Mode + MCP + a unique port + all tool groups — the default is `read`/`navigate` only) and `accord-session` (a token from a layer-1 account) seeded before launch. The app comes up already signed in with its MCP server listening. Isolation keeps the two instances off each other's Hive boxes and off your real profile.

## Four findings worth reading

1. **A debug bundle launched directly never runs its Dart entrypoint.** The engine starts, native plugins register, the GTK loop idles — `main()` never executes, so no Hive box opens and no MCP server starts. No error anywhere; it presents as a window that does nothing. `flutter test -d linux` works only because the tool attaches and injects the entrypoint. Release is now preferred, and this is why the CI job builds release.
2. **`XDG_DOCUMENTS_DIR` doesn't decide where app data goes.** `getApplicationDocumentsDirectory()` shells out to the **`xdg-user-dir` executable** and ignores the env var. Under a fresh `HOME` with no `user-dirs.dirs` it returns `$HOME`, not `$HOME/Documents` — so seeding the obvious path wrote where the app never looks, and it came up with default settings and no MCP. The launcher now writes a `user-dirs.dirs` and asks `xdg-user-dir` the same question the app will.
3. **`select_channel` must follow `select_space`** — it resolves against the selected space's channel list, and answers "Channel not found" for a channel that plainly exists otherwise.
4. **HTTP keep-alive flaked ~1 run in 2** ("Connection closed before full header was received") — a pooled socket the app had dropped. Each call now opens its own connection, with one retry on transport errors.

## Voice is deliberately not covered

It's the strongest reason to have this layer, and the current tool surface can't express the assertion: `get_current_state` reports only the caller's **own** `voice_channel_id`/`voice_self_mute`, and nothing reports other members' voice states. B has no way to say what it thinks A is doing.

A small addition to `mcp_tools.dart` — a `list_voice_states` in the `read` group — makes the scenarios a handful of lines each. Left out rather than faked; noted in the README and on #222.

## CI

New non-blocking `multi-instance` job on Linux under xvfb, after a release build, with the same no-op guard as `integration` (a suite that skips everything must not report success). GitHub-hosted for the same reason as `ui-e2e`: the self-hosted runners are Ubuntu 20.04 and can't build the Linux desktop app.

Refs #217, #222.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -12,6 +12,10 @@ name: CI
 # - `ui-e2e`   : the real app driven on Linux under xvfb against the same
 #                server image. GitHub-hosted (see the job's own note) and also
 #                non-blocking; see integration_test/README.md.
+# - `multi`   : two real app processes driven over the in-app MCP surface,
+#                against the same server image. Non-blocking; the most
+#                expensive suite (release build + two apps).
+#                See multi_instance/README.md.
 # - `build`    : release builds for web / Android / Linux / Windows, uploaded
 #                as artifacts. Only runs on manual `workflow_dispatch` — PRs
 #                don't need four release builds to merge, and tagged releases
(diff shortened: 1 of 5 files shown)


krazyjakee:
Reviewed this PR against the three requested areas (code quality, file size/structure, test coverage). No 3D/visual changes here, so that section doesn't apply. Pushed fixes directly to this branch (`037f466`).

**Note on branch:** this session's assigned branch was `claude/eager-curie-zg3x1s`, but that ref no longer exists on `origin` — its commit history (HEAD `893e4feb...`) is identical to this PR's head, so I pushed the fixes here instead per the "push to the PR branch, don't create a new one" instruction.

## Findings & fixes (ranked by severity)

1. **[Fixed] `resolveBinary()` silently fell back to a debug build, contradicting its own documented rationale.** `multi_instance/support/app_instance.dart`
   The doc comment (and `multi_instance/README.md`) explain in detail that a debug bundle launched directly never runs `main()` — no Hive box opens, no MCP server starts, the window just sits there. Yet `resolveBinary()` still searched `['release', 'debug']` and would happily return the debug candidate. Since `AppFleet.resolve()` calls `resolveBinary()` specifically to catch "no usable bundle" up front and turn it into a clean `skip:` reason, this meant a debug-only environment (e.g. someone who'd previously run `flutter run` but never `flutter build linux --release`) sailed past that check, then failed 60–120s later deep inside `setUpAll` with an opaque "the app's MCP server never answered on port ..." timeout — exactly the failure mode the surrounding comments were written to explain away.
   Fix: split the decision into a pure `chooseBundle({releaseExists, debugExists})` → `BundleChoice.{release,debugOnly,none}` (mirroring the existing `onboardingTrigger`/`releaseNotesTrigger` pure-decision pattern already used elsewhere in this codebase), and unit-tested it in `test/multi_instance/app_instance_test.dart`. A debug-only environment now fails immediately with an actionable message instead of a slow, confusing timeout.

2. **[Fixed] Stale job-name in `ci.yml`'s header comment.** The top-of-file job index still said `` `multi` `` while the actual job id/name is `multi-instance`. Small, but it's the kind of doc drift that makes the next person grep for the wrong string.

3. **[Fixed] Test-coverage gaps for the suite's own logic.**
   - `_memberId` (the REST-vs-MCP member-shape flattening helper in `two_clients_test.dart`) had no direct coverage — only exercised indirectly by the full two-app fleet, which needs a display, a release build, and a live server. Added a `group('_memberId', ...)` that runs unconditionally (not gated on `fleet.skipReason`), covering the flattened-id, nested-`user.id`, `user_id`-fallback, and non-`Map` branches.
   - The final test ("an edit by alice reaches bob, and so does a delete") relied on an earlier test in the same `group` having left the channel selected, rather than calling `openChannel` itself. Tests in a group execute in declaration order today, but nothing enforces that stays true, and the fix is two cheap idempotent calls — made it self-contained.

## Not flagged
- No file in the diff is anywhere near the 1000-line split threshold (largest is `app_instance.dart` at ~450 lines).
- Verified the MCP tool names/arg shapes the test drives (`select_space`, `select_channel`, `get_current_state`, `list_messages`, `list_members`, etc.) against `lib/features/developer/services/mcp_tools.dart` — all consistent, including the `select_channel`-needs-`select_space`-first behavior the README documents.
- Verified the seeded Hive settings/session keys (`mcpPort`, `mcpAllowedGroups`, `mcpToken`, the `'99.0.0'` onboarding/release-notes sentinel) all match their real consumers in `lib/features/settings` and `lib/features/updates` — the `'99.0.0'` sentinel is an existing convention already used in `test/features/updates/*`, not something new.

## Couldn't verify
No Flutter/Dart SDK is available in this execution environment, so I couldn't run `flutter analyze` or `flutter test` myself — the changes above were verified by manual cross-referencing against the actual tool/settings implementations rather than by running the suite. Worth a CI run confirming green before merge.

---

PR #227, opened 2026-08-18
Give voice channel chat full text-channel parity

krazyjakee:
## The problem

The voice channel view rendered its own slimmed-down message list — avatar, author name, content, and nothing else. A voice channel's text chat is an ordinary text channel, but from a UX perspective it had been treated as a different thing, so it silently lacked everything a text channel offers:

| | Text channel | Voice chat (before) |
|---|---|---|
| Context menu (right-click / long-press) | ✅ | ❌ |
| Reactions (pills, add, see who reacted) | ✅ | ❌ |
| Reply / thread | ✅ | ❌ |
| Edit / delete / pin / report / copy | ✅ | ❌ |
| Bulk select + delete | ✅ | ❌ |
| Attachments, embeds, image lightbox | ✅ | ❌ |
| Timestamps, edited marker, mention highlight | ✅ | ❌ |
| History paging ("load older") | ✅ | ❌ (first page only) |
| Pinned messages, channel mute | ✅ | ❌ |
| Composer: attachments, emoji picker, mentions, drafts, typing | ✅ | ❌ (plain text field) |

## The fix

`VoiceTextPanel` is now a thin wrapper around the real `MessagePane` rather than a second, slimmer implementation of it. The two surfaces can't drift apart again, and the panel keeps only the voice-specific bits it actually owns: acking the channel on open (against the connection the *call* is pinned to, which needn't be the active one) and the panel's title/close button.

`MessagePane` gains a `panel` presentation to support it:

- a compact 44px header — channel name, pinned messages, mute, close — with a divider on its leading edge, in place of the full 48px channel header;
- tighter list padding for the narrower column;
- no delegation back to `VoiceChannelView`, which would otherwise recurse, since the voice view is what embeds the pane.

Everything below the header is the pane's existing widget tree, unchanged.

## Layout fixes that fall out of the narrow column

Embedding the full pane in a ~340px column surfaced two real overflows, both fixed:

- **Hover action bar.** It reserves its full width even while invisible, which left the message itself about 40px. In a narrow row the same bar now floats over the row's top-right corner while hovered (`_FloatingActionsOverlay`) instead of sitting inline. Wide panes are untouched.
- **Author header.** A long name now ellipsizes rather than pushing the timestamp out of the row — matching what the thread view already did.

The voice side panel also goes from 300px to 340px so the pane's avatar gutter, reactions and floating action bar stay readable.

## Testing

`flutter analyze --no-fatal-infos` clean (same 23 pre-existing infos as master) and all 970 tests pass.

- New `test/features/voice/voice_text_panel_parity_test.dart` drives the voice view end to end against a mocked server and asserts the chat renders through `MessagePane`, the hover actions and overflow menu are reachable, reaction pills render and toggle, the full composer (attachments + emoji) is present, and the close button still closes the panel. It also pins the layout fix by asserting the message keeps the bulk of the row's width before hover.
- The existing voice widget tests get an in-memory settings override, since the embedded pane reads the preferences the live `SettingsController` loads from a Hive box that only `setupHive()` opens.

https://claude.ai/code/session_01R57ScM8jooyAgneZvpJmWD

---
_Generated by [Claude Code](https://claude.ai/code/session_01R57ScM8jooyAgneZvpJmWD)_

### DIFF
--- a/lib/features/messaging/views/message_pane/message_pane.dart
+++ b/lib/features/messaging/views/message_pane/message_pane.dart
@@ -74,11 +74,34 @@ class MessagePane extends ConsumerStatefulWidget {
     required this.channel,
     required this.channelId,
     required this.spaceId,
+    this.panel = false,
+    this.panelTitle,
+    this.onClosePanel,
   });
 
   final AccordChannel? channel;
   final String? channelId;
   final String? spaceId;
+
+  /// Renders the channel's *text chat only*, in the slimmer presentation used
+  /// by the voice channel view's side panel (see [VoiceTextPanel]): a compact
+  /// header with a left divider instead of the full 48px channel header, and no
+  /// delegation to [VoiceChannelView] — a voice channel shows its chat here
+  /// rather than recursing back into the voice view that embedded this pane.
+  ///
+  /// Everything below the header is the same widget tree the regular pane
+  /// builds, so a voice channel's chat keeps every text-channel affordance:
+  /// context menus, reactions, replies, threads, edit/delete/pin/report,
+  /// attachments and history paging (#210).
+  final bool panel;
+
+  /// Header title for [panel] mode, when the caller knows the channel name but
+  /// not the [AccordChannel] itself. Falls back to the channel's own name.
+  final String? panelTitle;
+
+  /// Closes the panel. When null (or outside [panel] mode) no close button is
+  /// shown.
+  final VoidCallback? onClosePanel;
 
   @override
   ConsumerState<MessagePane> createState() => _MessagePaneState();
@@ -212,7 +235,9 @@ class _MessagePaneState extends ConsumerState<MessagePane> {
       );
     }
 
-    if (channel?.type == 'voice') {
+    // In panel mode this pane *is* the voice view's chat panel — delegating
+    // back to the voice view would recurse.
+    if (!widget.panel && channel?.type == 'voice') {
       return VoiceChannelView(
         channelId: channelId,
         spaceId: spaceId,
@@ -294,14 +319,23 @@ class _MessagePaneState extends ConsumerState<MessagePane> {
       );
     }
 
+    final panel = widget.panel;
+
     return Container(
-      color: colors.background,
+      decoration: BoxDecoration(
+        color: colors.background,
+        // In panel mode the pane sits beside the video grid, so it needs a
+        // divider on its leading edge too.
+        border: panel
+            ? Border(left: BorderSide(color: colors.foreground, width: 1))
+            : null,
+      ),
       child: Column(
         children: [
           Container(
-            height: 48,
+            height: panel ? 44 : 48,
             alignment: Alignment.centerLeft,
-            padding: const EdgeInsets.only(left: 16, right: 8),
+            padding: EdgeInsets.only(left: panel ? 12 : 16, right: panel ? 4 : 8),
             decoration: BoxDecoration(
               border: Border(
                 bottom: BorderSide(color: colors.foreground, width: 1),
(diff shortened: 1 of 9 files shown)


krazyjakee:
## Automated review

Reviewed the diff (all 9 changed files) against the three areas requested: code quality, file size/structure, and test coverage. No 3D model/rendering code is touched, so that check doesn't apply.

**Result: no issues requiring a fix.** No commit pushed.

### Code quality
- The `MessagePane` → `panel` mode refactor is clean: a single boolean gates the compact header, narrower list padding, and skips the voice-view delegation that would otherwise recurse. The old bespoke `VoiceTextPanel` (300+ lines: its own message list, composer, typing line) is fully replaced by a thin wrapper, matching the "reuse Bonfire" project convention.
- Traced the new `_FloatingActionsOverlay` / `narrow` row path carefully, since floating a `Row(mainAxisSize: MainAxisSize.min)` inside a `Stack`'s `Positioned` is an easy place to get Flutter's unbounded-vs-bounded main-axis constraints wrong. The `mainAxisSize: MainAxisSize.min` addition on `_HoverActions` is the right fix — without it, the floated bar would stretch to the full row width inside the (bounded) `Positioned` slot rather than hugging its icons in the corner. This is correct.
- `hoverActions` is built for every row when not editing/selecting/mobile, then routed either inline (`!narrow`) or into the floating overlay (`narrow`) — verified the two paths are mutually exclusive so the actions never render twice.
- `ellipsizeName: true` is now unconditional (not just for `narrow`/panel rows), matching the PR description ("matching what the thread view already did") — intentional, not scoped incorrectly.

### File size/structure
All changed files are well under 1000 lines (`message_row.dart` 983, `message_pane.dart` 648, `voice_view.dart` 968, `voice_text_panel.dart` 96) — no split needed.

### Test coverage
`voice_text_panel_parity_test.dart` is a solid addition: it drives the real widget tree (not a shortcut) and asserts the previously-missing affordances — hover actions, overflow menu, reactions, full composer, close button — plus a regression assertion pinning the layout fix (message keeps >180px of the ~340px row before hover). The four other voice test files' `_FakeSettingsController` additions are necessary now that they render the embedded `MessagePane`, which reads `SettingsController`; the duplicated boilerplate across 5 files matches an existing pattern already used elsewhere in the test suite (`global_unread_test.dart`, `updates_page_test.dart`, etc.), so it's consistent with project convention rather than a new problem.

### Minor (non-blocking) observation
On hover, the floating action bar (`top: 0, right: 4`) can visually sit over the author-name/timestamp header of an ungrouped row in the ~340px panel, since it isn't offset below the header's own padding. This mirrors common floating-toolbar UX elsewhere (Discord et al.) and doesn't block interaction, so I'm not treating it as a fix — flagging only in case it's not what was intended.

CI (`Analyze & Test` on stable and beta) is green on the head commit, which corroborates the manual read — this sandboxed review environment doesn't have the Flutter SDK installed, so I read the diff and full files rather than re-running `flutter analyze`/`flutter test` myself.

If a reviewer wants a specific area re-checked (e.g. narrow-layout breakpoint behavior below 720px, which isn't covered by the new tests), let me know and I'll take another pass.


---

krazyjakee:
Both points from the automated review checked.

**Floating bar over the author header — confirmed, keeping as is.** Measured it in a widget test rather than eyeballing: on the ~340px panel the bar occupies `LTRB(1142, 567, 1182, 607)` and the header row spans `LTRB(919, 569, 1190, 590)`, so they do intersect. In practice the header's *content* is left-packed, so occlusion only shows when name + timestamp actually run past the bar's left edge.

Every alternative is worse:

- Raising the bar to straddle the row's top edge (`top: -8` + `Clip.none`) puts it outside the row's bounds, and Flutter's bounded hit-testing then makes it unclickable.
- Reserving trailing space in the header while hovered introduces layout jitter on every mouse-over.
- Going back inline is the ~40px-wide message bug this PR fixes.

The bar is opaque, appears only on hover, and un-hovering restores the header immediately — the standard floating-toolbar trade-off. Intended.

**Narrow-layout coverage gap — closed** in 4e45b88. Added a test at 400×800 (below the 720px breakpoint, where the chat takes over the body and rows render in touch mode with no hover bar at all), asserting the long-press menu carries the full set: Add reaction, Reply, Thread, Copy text, Edit, Delete. That's the parity claim for touch, which nothing was pinning before.

971 tests pass, `flutter analyze --no-fatal-infos` unchanged.

---

PR #228, opened 2026-08-19
fix(ios): clear the App Review 2.5.1 rejection and submit Apple releases from CI

krazyjakee:
iOS has never reached the App Store: 0.2.6 (build 141) was rejected on
2026-07-09 under guideline **2.5.1 Performance: Software Requirements**, and
every release since then stopped at a TestFlight build because the iOS lane
only called `upload_to_testflight`. This branch fixes both halves.

## The rejection

Apple's message was the automated one and named no API. Scanning every
framework the iOS app links found exactly one dirty binary — the
`Mpv.xcframework` vendored by `media_kit_libs_ios_video`:

```
$ llvm-nm -u Mpv.framework/Mpv | grep -E '^_(fork|execve|setsid|waitpid|kill)$'
_execve _fork _kill _setsid _waitpid
```

mpv's POSIX subprocess and TTY layer (`tcgetattr`/`tcsetattr`/`tcgetpgrp`,
`shm_open`), plus `EAGLContext` and `CVOpenGLESTextureCache*` — the OpenGL ES
stack Apple deprecated in iOS 12. `media_kit_video`'s own iOS plugin renders
through that same path and compiles with `GL_SILENCE_DEPRECATION`. WebRTC, the
FFmpeg libs, libxml2 and every plugin source came back clean.

So iOS no longer gets media_kit: `pubspec.yaml` lists the native libs
per platform instead of via the `media_kit_libs_video` umbrella, omitting the
iOS package, which makes `media_kit_video`'s iOS podspec fall back to its
no-op stub. Video attachments there play through `package:video_player`'s
AVFoundation backend. AVFoundation demuxes less than libmpv, so a container it
cannot open shows a poster rather than a player that fails on the first frame.

## Submitting from CI

Also carries `d67a743e`, previously stranded on
`claude/voice-channel-chat-parity-rogbid` — the Apple lanes now create the
version record, attach the build, deliver generated release notes and submit
for review with automatic release on approval.

## Verification

`flutter analyze` clean, 976 tests pass (5 new ones cover the container gate),
`flutter build linux` succeeds, and `media_kit_libs_ios_video` is gone from
`pubspec.lock` and from the iOS half of `.flutter-plugins-dependencies`. The
iOS archive itself is first compiled by the release job — it cannot be built on
Linux.

The stuck 0.2.6 review submission has been cancelled in App Store Connect, so
the version record is editable and `deliver` can submit 0.2.11 against it.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -45,7 +45,7 @@ on:
         type: boolean
         default: true
       deploy_mac:
-        description: "Build the Mac App Store build, upload it and submit it for review"
+        description: "Build the Mac App Store build and upload it (no review submission — see docs)"
         type: boolean
         default: true
       deploy_android:
@@ -930,11 +930,12 @@ jobs:
             --dart-define=APP_STORE=true \
             --dart-define=SENTRY_DSN="${SENTRY_DSN:-}"
 
-      - name: Generate What's New
-        # See the iOS job: Apple rejects a submission with no release notes.
-        run: ./dist/app-store-release-notes.sh
-
-      - name: Archive, sign, upload & submit for review
+      - name: Archive, sign & upload
+        # Upload only, no review submission: the macOS product page has never
+        # been filled in, so there is nothing to submit yet and deliver's
+        # metadata step hard-fails on the missing review detail. See the
+        # :appstore lane and docs/app-store-deploy.md. No release notes are
+        # generated here for the same reason — nothing delivers them.
         env:
           ASC_KEY_ID: ${{ secrets.ASC_KEY_ID }}
           ASC_ISSUER_ID: ${{ secrets.ASC_ISSUER_ID }}
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -41,11 +41,11 @@ on:
   workflow_dispatch:
     inputs:
       deploy_ios:
-        description: "Build + upload the iOS App Store build to TestFlight"
+        description: "Build the iOS App Store build, upload it and submit it for review"
         type: boolean
         default: true
       deploy_mac:
-        description: "Build + upload the Mac App Store build to App Store Connect"
+        description: "Build the Mac App Store build, upload it and submit it for review"
         type: boolean
         default: true
       deploy_android:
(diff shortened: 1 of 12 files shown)


krazyjakee:
Reviewed the diff (code quality, file size, test coverage; no 3D/rendering code touched so no visual review needed). Pushed one fix to this branch.

## Findings

**1. (High, fixed in c702f9b) `prettify()` in `dist/app-store-release-notes.sh` used GNU sed's `\U` case-conversion escape, but the jobs that run it (`ios-appstore`, `mac-appstore`) both use `runs-on: macos-15`.** macOS's default `/usr/bin/sed` is BSD sed, which doesn't support `\U`/`\L` — it passes the escape through as a literal `U` rather than erroring, so every bullet in the real App Store "What's New" text would have come out as `• Uadd x` instead of `• Add x`. This wasn't caught because the script clearly wasn't exercised against BSD sed (verified locally: GNU sed here handles `\U` fine, which is presumably how it was tested). Replaced the capitalization step with a portable `awk 'toupper(substr(...))...'` pipeline and verified it against this repo's actual commit history — output is correct. Also pinned `LC_ALL=C` for the whole script so the `${#notes}` length check and the `head -c` truncation count the same units regardless of runner locale (minor: they could previously disagree on multi-byte content, though never in a way that risked exceeding Apple's 4000-char cap — just occasionally over-conservative).

**2. (Low, not changed) `.github/workflows/release.yml` is 1156 lines**, over this repo's 1000-line split guideline. It was already 1135 lines before this PR (which adds ~21 net lines), so this isn't new. I didn't split it: GitHub Actions workflows don't support clean in-file modularization short of extracting jobs into separate reusable workflows, which is a structural change with real blast radius on the release pipeline and isn't verifiable without an actual tag push through Actions. Flagging for a deliberate follow-up rather than a speculative refactor bundled into this PR.

**3. (Info) Couldn't run `flutter analyze`/`flutter test` locally** — no Flutter SDK in this environment. I did syntax-check the Ruby `Fastfile` (`ruby -c`, OK) and the YAML workflow (parses OK), and manually traced the `InlineVideoPlayer` container-gating logic (`_playableHere`) against all 5 new tests in `test/features/messaging/inline_video_player_test.dart` — the logic and test expectations line up correctly for the cases covered (off-iOS, AVFoundation container on iOS, non-AVFoundation container on iOS, extension-less URL falling back to filename, embed-page URL). Deferring to this PR's own CI run (`Analyze & Test`) for the authoritative result.

No other correctness issues found in the `InlineVideoPlayer`/`video_player` fallback path, the `pubspec.yaml` media_kit platform split, or the `Fastfile` submission lane.

---

PR #229, opened 2026-08-19
fix(messaging): revive the voice chat panel's message actions; ban can purge message history

krazyjakee:
Two reports, both presenting as "does nothing": deleting a message in a voice channel, and reporting one. Plus the feature that came out of the follow-up — banning a member with the option to delete their message history.

## 1. The voice chat panel's "⋯" menu dropped every selection

A narrow row — the voice channel's chat panel — only builds its hover action bar *while hovered*; inline, the bar would squeeze the message into a few dozen pixels. Opening the `PopupMenuButton` pushes a modal barrier, the barrier ends the hover, `_hovered` goes false, and the button unmounts mid-menu. Flutter then discards the pick outright:

```dart
).then<void>((T? newValue) {
  if (!mounted) { return null; }   // ← selection dropped
```

So **Delete, Edit, Pin and Report were all dead** in voice chat. React / Reply / Thread worked, because those are plain `IconButton`s that fire synchronously on tap. Normal text channels were never affected — there the bar stays mounted behind `Opacity(0)`.

`PopupMenuButton.onOpened` fires *before* the barrier steals the hover, so it can flag the menu open and keep the bar mounted for the menu's lifetime.

## 2. Deleting a message failed silently

`_delete` discarded the controller's return value, and `AccordMessagesController.delete` only `debugPrint`s on failure — so any server rejection looked exactly like the bug above. It now surfaces a snackbar.

## 3. Reporting could no-op too

`_submit` has a validation guard:

```dart
if (category == null) {
  setState(() => _error = 'Choose a reason for this report.');
  return;
}
```

…that was unreachable, because the Submit button was disabled under precisely that condition. Opening the dialog and clicking "Submit report" without touching the Reason dropdown did nothing, with no explanation. Enabling the button lets the existing guard run.

## 4. Bans can now delete message history

The ban actions offered no way to clear a spammer's posts. `showBanDialog` replaces the plain confirm across all three ban paths — member popout, space reports, admin reports — with Discord's windows (none / 1h / 6h / 12h / 24h / 3d / 7d).

Backed by **DaccordProject/accordserver#58**, which adds `delete_message_seconds` to `PUT /spaces/{id}/bans/{user_id}`. Until then accordkit documented the field but the server had no such thing on `CreateBanBody`, so serde dropped it and the request silently no-opped.

**This PR is safe to merge first.** The field is omitted from the body entirely when nothing is being purged, so a normal ban is byte-identical to what it sends today. And the member popout reports the count the server actually deleted rather than assuming — which is exactly how an older server's silent ignore becomes visible (`"Banned Bob and deleted 0 messages"`) instead of looking like it worked.

## Tests

4 new widget tests for the ban dialog (default keeps history and omits the field, chosen window round-trips, cancel resolves null, target is named). All 980 tests pass; `flutter analyze --no-fatal-infos` reports no new issues.

The action-bar fix is a widget-lifecycle interaction that isn't reachable from a widget test without a full `MessagePane` harness (Riverpod auth client, members, settings), so it wants manual verification:

1. Join a voice channel, open its chat, hover a message, "⋯" → **Delete**. The confirm should appear and the message should go. Same for Edit / Pin / Report.
2. Right-click a member → **Ban**, pick "Previous 24 hours" (needs accordserver#58). Their recent messages should vanish live for everyone in the channel, and the snackbar should report the count.

## Not addressed

`bulk_delete_messages` on the server broadcasts no gateway event at all, unlike single delete — so bulk deletes don't propagate to other connected clients. Pre-existing, out of scope here.


krazyjakee:
Reviewed the diff (voice-chat action-bar fix, delete/report guard fixes, and the ban-purge feature across `ban_dialog.dart`, `accord_member_popout.dart`, `accord_reports.dart`, `admin_reports_tab.dart`, `bans_api.dart`). Logic checked out — the `PopupMenuButton`/hover lifecycle fix, the `mounted`/context-capture handling around the new `onResult` callback, and the report-button guard all matched their stated intent and the surrounding codebase's conventions. `flutter analyze --no-fatal-infos` and the full `flutter test`/`dart test` suites pass (981 + 154 tests) after the changes below.

Two things fixed and pushed (`d21a751`):

**1. File size (medium).** `message_row.dart` crossed the project's 1000-line threshold (983 → 1021) with this PR's own action-bar changes. Split the presentational leaf widgets (`_FloatingActionsOverlay`, `_MaybeTappable`, `_GutterTimestamp`, `_HoverActions`, `_MessageActions`) into a new `message_row_actions.dart` part file, following the same pattern `message_pane.dart` already uses for `attachments.dart`/`composer.dart`/`reactions.dart`/`mute_button.dart`. `message_row.dart` is now 780 lines. Pure move, no behavior change.

**2. Missing test coverage (medium).** Two code paths this PR introduced had zero coverage:
- `message_row.dart`'s new `if (!ok) showInfoSnack(context, 'Failed to delete message')` branch — added a test to `message_row_target_test.dart` that mocks a 500 on the delete request and asserts the row survives and the snackbar appears.
- `BansApi.create`'s `delete_message_seconds` field and `deleted_message_count` response parsing — the entire point of the ban-purge feature, and `BansApi` had no test coverage at all before this. Added a `BansApi` group to `packages/accordkit/test/endpoints_test.dart` (empty body by default, field forwarded when purging, count parsed back, `remove` path).

Everything else looked solid — `BanRequest.toJson()`'s omit-when-zero behavior is correctly tested in the existing `ban_dialog_test.dart`, the `mounted` guards after every `await` in the three ban call sites are consistent with the file's existing patterns, and the messenger-capture-before-pop ordering in `accord_member_popout.dart`'s `_ban()` is correct (captured before `Navigator.pop()`, used in `onResult` which runs before the pop).

No 3D/visual changes in this diff, so no render pass was needed.

---

PR #287, opened 2026-08-27
Open the channel drawer from a swipe anywhere on mobile

krazyjakee:
Swipe-to-open only worked from a thin strip at the screen edge. That
strip is Flutter's own gesture: `DrawerController` paints a translucent
drag target `Scaffold.drawerEdgeDragWidth` wide against the edge, and
the narrow layout set that to 48 logical pixels — so a swipe starting
past ~48px never reached the drawer at all.

Widening the strip to the whole screen is not the fix. That target lives
in `_ScaffoldSlot.drawer`, which is painted after the body and therefore
hit-tested before it, so a full-width strip wins the gesture arena
against every horizontal scrollable underneath: the open-tab strip stops
scrolling and the voice participant rails go dead.

Add `DrawerSwipeArea` instead and wrap the narrow body in it. Sitting
inside the body, as an ancestor of the content, it joins the arena after
the content's own recognizers, so inner horizontal drags are claimed
first and only an unclaimed one opens a drawer. A rightward swipe (or
flick) from anywhere over the message area opens the channel drawer; the
mirrored swipe opens the member list, matching the right-edge gesture
that already existed. Directionality is honoured, and the gesture is
limited to touch platforms the way `DrawerController` limits its own.

The 48px edge strip stays: there the framework's gesture drags the
drawer under the finger, which beats the snap this can offer from
mid-screen (`ScaffoldState` exposes no way to drive the animation
incrementally).

Co-Authored-By: Claude Opus 5 <noreply@anthropic.com>
Claude-Session: https://claude.ai/code/session_01CnLkwMF6a7jv6WiyXApJVt

### DIFF
--- a/docs/navigation/spaces-and-channels.md
+++ b/docs/navigation/spaces-and-channels.md
@@ -52,3 +52,5 @@ When you open a channel, it appears as a **tab** at the top of the message area.
 ## On Small Screens
 
 On narrow windows or mobile devices, the sidebar becomes a **drawer** that slides in from the left. Tap the hamburger menu button to open it, and tap outside to close it.
+
+You can also swipe it open. On touch devices, swipe right from anywhere over the message area — not just from the screen edge — and the channel drawer opens; swipe left the same way to open the member list. Swipes that start on the tab bar still scroll the tabs, and a swipe that starts at the very left edge drags the drawer along with your finger.
--- a/lib/features/spaces/views/accord_home.dart
+++ b/lib/features/spaces/views/accord_home.dart
@@ -48,6 +48,7 @@ import 'package:bonfire/features/events/controllers/connection.dart';
 import 'package:bonfire/shared/components/async_state_views.dart';
 import 'package:bonfire/shared/components/color_swatch_chip.dart';
 import 'package:bonfire/shared/components/context_menu.dart';
+import 'package:bonfire/shared/components/drawer_swipe_area.dart';
 import 'package:bonfire/shared/components/horizontal_wheel_scroll.dart';
 import 'package:bonfire/shared/components/server_unreachable.dart';
 import 'package:bonfire/shared/models/server_entity_key.dart';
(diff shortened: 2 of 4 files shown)


krazyjakee:
Reviewed this PR against the three requested areas (code quality, file size/structure, test coverage). No 3D model/rendering code is touched, so that check doesn't apply.

**Verification performed** (Flutter 3.47.2 stable, Dart 3.13.2):
- `flutter analyze --no-fatal-infos` on the touched files — no issues.
- `dart format --set-exit-if-changed` on `drawer_swipe_area.dart`, `accord_home.dart`'s diff region, and the new test file — clean. (The full `accord_home.dart` does have pre-existing formatting drift at two unrelated call sites, lines 203/519, from before this PR — not introduced here, left alone.)
- `flutter test test/features/spaces/` — all 49 tests pass, including the 14 new ones in `drawer_swipe_area_test.dart`.
- Wrote and ran an extra throwaway widget test tapping a button nested inside `DrawerSwipeArea`: the tap still fires. This was the main regression risk I wanted to rule out myself — an opaque `GestureDetector` wrapping message content could plausibly eat taps on messages/reactions/links — but Flutter's gesture arena resolves tap vs. horizontal-drag correctly here (a `TapGestureRecognizer` accepts explicitly on pointer-up when there's no significant movement, which beats a drag recognizer that never explicitly claimed victory), matching what the PR description assumes.
- Confirmed `AccordHomeScreen` is the only `Scaffold` in `lib/` with `drawer:`/`endDrawer:`, so there's no second narrow-layout site this fix should also have touched.
- Manually walked the RTL, no-end-drawer, and velocity-vs-distance tie-break logic in `drawerSwipeAction` — all correct and covered by tests.

**Findings: none.** The new `DrawerSwipeArea` widget is well-scoped (126 lines), the approach (join the gesture arena as an ancestor rather than widen the edge-drag strip) is the correct fix for the stated problem, and the test suite exercises the real regression risks (tab-strip scroll, edge strip, RTL, platform gating, threshold tuning). No code changes pushed since nothing needed fixing.

---

PR #294, opened 2026-08-30
feat(authentication): present the app's terms before signing in or registering

krazyjakee:
App Review rejected 0.2.16 under guideline 1.2 because no EULA is shown
before a user registers or logs in. The terms gate we had could not show
one: it read `tos_enabled` from the server's `GET /settings`, which
accordserver serves to authenticated users only, so a signed-out fetch
401s and the gate always resolved to "disabled" — and it was wired into
the Register tab alone, leaving sign-in with no terms at all.

Carry terms of our own instead. `app_terms.dart` holds the document
(what may not be posted, how to report and block, the GPL and warranty
note); acceptance is recorded per terms version in the device-global
`auth` box, so a revision re-prompts and a profile switch does not. The
gate replaces the whole signed-out flow — welcome, server browser and
credentials form alike — until it is accepted, and the text stays
reachable afterwards from the auth form and Settings → About.

A server's own ToS checkbox is unchanged and still shows alongside this
when the server advertises one.

Refs #289

Co-Authored-By: Claude Opus 5 <noreply@anthropic.com>
Claude-Session: https://claude.ai/code/session_01GVXDeSVn9njx9vcDv1JWfD

Issue #293, opened 2026-08-30
App Review 1.2: record the UGC-precautions screen recording and reply in App Store Connect


krazyjakee:
## App Review report

- Submission: `0c312f06-8042-4684-8684-2d3f94289cb1`
- Review date: 2026-08-30
- Reviewed version/build: `0.2.16 (160)`
- Guideline: **1.2 — Safety, User-Generated Content**

Apple asked for a reply containing a screen recording **captured on a physical device** showing all three precautions. This is the submission-side deliverable that closes 1.2 once the code work lands.

The recording must demonstrate, in order:

- [ ] The EULA / terms of use presented **before** registering or logging in (#289)
- [ ] The mechanism for users to flag objectionable content — shown in a space channel *and* in a direct message (#290)
- [ ] The mechanism to block an abusive user (already implemented: member popout "Block user", and the DM relationship actions in `lib/features/user/views/accord_direct_messages.dart`)

Then:

- [ ] Attach the recording in **App Review Information → Notes** in App Store Connect, so later submissions carry it too.
- [ ] Reply to the 1.2 message on the App Review page, pointing at the timestamps for each precaution.

Blocked by #289 and #290 — the recording can only be made once both ship in a build.

krazyjakee:
**Unblocked on the code side.** Both dependencies are merged to `master`, so all three precautions Apple asked to see are now demonstrable. What remains is the physical-device recording and the App Store Connect reply, neither of which can be done from here.

### What to record, and where it now lives

**1. EULA / terms before registering *or* logging in** — #289, `a09888e4` + `749c4b5a`.
The gate replaces the entire signed-out surface (welcome, server browser and credentials form alike) before any auth mode is chosen, so it demonstrates on a cold install without needing to reach Register. Acceptance is recorded per terms version in the device-global `auth` box — so **delete and reinstall, or clear app data, before recording**, otherwise the gate will not appear. The text stays reachable afterwards from the auth screen's "By continuing you agree to …" line and from Settings → About, both worth showing.

**2. Flagging objectionable content, in a space channel *and* in a DM** — #290, `a09888e4` + `5a47d319`.
- In a space: long-press (or hover) any message from another user → **Report message**.
- In a DM: same gesture on a message — the `spaceId != null` gate that used to remove it there is gone.
- On a user: **Report user** from the member popout, from the DM user profile, and from the 1:1 DM header's overflow button.
Show at least the space-channel message path and the DM path, since Apple named both contexts.

[...]

Issue #292, opened 2026-08-30
App Review 2.2: build judged a pre-release/trial with a limited feature set — establish what the reviewer saw


krazyjakee:
## App Review report

- Submission: `0c312f06-8042-4684-8684-2d3f94289cb1`
- Review date: 2026-08-30
- Reviewed version/build: `0.2.16 (160)`
- Device: iPad Air 11-inch (M3)
- Guideline: **2.2 — Performance, Beta Testing**

> Your app appears to be a pre-release, test, or trial version with a limited feature set … complete, remove, or fully configure any partially implemented features.

Unlike the other two findings this one has no located root cause yet — it needs a determination before a fix, and it is the finding most likely to repeat if we only address 1.2 and 2.3.10.

## What the code search found

- No `beta` / `alpha` / `preview` / "coming soon" / "not implemented" strings in `lib/`, and no obviously stubbed screens.
- The signed-out entry point is fine on paper: `WelcomeView` → **Browse Servers** → the unauthenticated master-server directory (`lib/features/spaces/views/accord_discovery.dart`), which currently returns exactly **one** listing:

  ```
  $ curl -s https://master.daccord.gg/api/v1/directory
  {"data":[{"name":"Daccord Official","server_name":"Daccord","server_url":"https://chat.daccord.gg",
            "member_count":169,"slug":"general","space_id":"312525716459618304"}],"total":1}
  ```

So the plausible readings are: the reviewer landed on a nearly-empty directory and a fresh account in one space with little to do; or the two 1.2 gaps (no terms, no reporting outside space channels) read to them as unfinished features; or something on the iPad layout specifically. The previous round's blank-screen bug on the same device class (#285, #286) is worth ruling out on 0.2.16 rather than assuming it is gone.

## Next steps

- [ ] Reproduce a first-run on an iPad Air-class simulator against `0.2.16 (160)`: install → Browse Servers → register → join → message / voice, and record what a reviewer actually sees.
- [ ] Confirm the #286 onboarding-navigator fix is present and holding on the iPad nested navigation path in this build.
- [ ] Audit for features that are reachable but non-functional against the public instance (permissions-gated admin surfaces, empty states that look broken rather than empty).
- [ ] Seed the demo account so the reviewer lands in a populated space, and expand App Review Information notes with a step-by-step walkthrough of the main features.
- [ ] Reply to App Review describing the app's completed feature set once the above is established.

[...]

Issue #289, opened 2026-08-30
App Review 1.2: no EULA/terms are ever shown before registering or signing in


krazyjakee:
## App Review report

- Submission: `0c312f06-8042-4684-8684-2d3f94289cb1`
- Review date: 2026-08-30
- Reviewed version/build: `0.2.16 (160)`
- Device: iPad Air 11-inch (M3)
- Guideline: **1.2 — Safety, User-Generated Content**

Apple requires apps with user-generated content to present an EULA / terms of use **before a user registers or logs in**, and asked for a screen recording demonstrating it. Today the app can never show one.

## Diagnosis

The terms gate is server-driven and register-only, and the server call it depends on cannot succeed while signed out:

1. `fetchTosConfig()` (`lib/features/authentication/utils/tos_gate.dart:16`) reads `tos_enabled` / `tos_url` / `tos_text` from `AccordAuth.fetchServerSettings()`.
2. `fetchServerSettings()` (`lib/features/authentication/repositories/accord_auth.dart:209`) issues `GET /settings`.
3. In accordserver that route is **authenticated-only** (`src/routes/mod.rs:518`, "Public settings (GET only, any authenticated user)"). Signed out it returns `401`, so `fetchServerSettings()` returns `null` and the gate resolves to `enabled: false`.

Confirmed against the live public instance:

```
$ curl -s https://chat.daccord.gg/api/v1/settings
{"error":{"code":"unauthorized","message":"invalid or missing authentication"}}
```

Even if that call did succeed, the gate is only fetched in register mode (`lib/features/authentication/views/accord_login.dart:111`, `:133`) and the checkbox is only rendered in register mode (`lib/features/authentication/views/auth_form.dart:124`), so **sign-in never shows terms at all** — which is explicitly part of what Apple asked for.

Net effect for the reviewer: a fresh install, "Browse Servers" → register on `chat.daccord.gg`, and no terms are presented at any point.

## Proposed fix

Stop depending on the server for the app's own EULA:

- [ ] Ship a bundled Terms of Use + Community Guidelines document in the client, including the zero-tolerance-for-objectionable-content and abusive-user clauses Apple expects.
- [ ] Present acceptance **before** the auth form is usable (first-run gate, persisted per terms version), covering both Sign In and Register — not only Register.
- [ ] Keep an always-visible "By continuing you agree to …" line with tappable links on the auth screen.
[...]

PR #299, opened 2026-09-02
fix(authentication): stop a server's ToS disappearing silently

krazyjakee:
Finishes the last open checklist item on #289 (App Review guideline 1.2). PR #294 landed the bundled terms, the pre-auth gate covering both Sign In and Register, version-keyed device-global acceptance and the Settings entry; the remaining item was *"make its fetch failure non-silent"*.

`fetchTosConfig` mapped every settings-read outcome onto a single bool, so a failed `GET /settings` was indistinguishable from a server that advertises no terms — and accordserver serves that route to authenticated users only, so the signed-out read reliably 401s. A server that genuinely requires terms therefore showed none, with nothing logged.

`AccordAuth.fetchServerSettings` now returns the settings map, or the reason and status code it couldn't be read, and logs the failure. `TosConfig` carries a four-way `TosAvailability`:

| value | when | checkbox | notice |
|---|---|---|---|
| `advertised` | `tos_enabled == true` | yes, required | no |
| `absent` | server answered, no terms | no | no |
| `refused` | 401/403 | no | no |
| `unknown` | network error, 5xx, malformed body | no | yes |

`refused` is split out deliberately. It is the expected outcome of every signed-out read today, so a user-visible "couldn't be loaded" note would sit permanently on the register screen of every instance — which reads as a half-configured feature under the open guideline 2.2 rejection (#292) and would be captured in the pre-auth terms screen recording #293 asks for. It is logged, not shown. The enum documents this so it isn't "fixed" back later; the case disappears once the accordserver follow-up exposes the ToS fields unauthenticated.

The gate never blocks: `tosRequired` is true only for `advertised`, so the app's own bundled terms remain the thing that gates auth.

## Verification

- `flutter test` — 1141 passing, 0 failing, 0 skipped.
- `scripts/codegen.sh --check` — clean; the one-line `accord_auth.g.dart` hash change is committed.
- New `test/features/authentication/server_tos_gate_test.dart` (11 tests) covers all four states, both 401 and 403 for `refused`, 0/404/500/502 for `unknown`, and drives two failure paths end to end from a settings result through `fetchTosConfig` into a pumped form.

Refs #289

### DIFF
--- a/lib/features/authentication/repositories/accord_auth.dart
+++ b/lib/features/authentication/repositories/accord_auth.dart
@@ -204,18 +204,46 @@ class AccordAuth extends _$AccordAuth {
   }
 
   /// Fetches a server's public settings (e.g. Terms-of-Service config) before
-  /// login. Returns the settings map, unwrapping a `{ data: {...} }` envelope,
-  /// or null on failure.
-  Future<Map<String, dynamic>?> fetchServerSettings(AccordServer server) async {
+  /// login. Exactly one of [settings] (the map, with a `{ data: {...} }`
+  /// envelope unwrapped) and [error] is non-null, so callers can tell "the
+  /// server said nothing is configured" from "we never got an answer".
+  ///
+  /// [statusCode] carries the HTTP status the failure came with (`0` when the
+  /// request never produced a response at all), because the *reason* matters:
+  /// accordserver serves `GET /settings` to authenticated users only, so a
+  /// signed-out read reliably 401s and callers treat that structural refusal
+  /// differently from an unexpected failure (#289).
+  Future<({Map<String, dynamic>? settings, String? error, int statusCode})>
+  fetchServerSettings(AccordServer server) async {
     final client = _restClientFor(server);
     try {
       final result = await client.rest.makeRequest('GET', '/settings');
-      if (!result.ok || result.data is! Map) return null;
+      if (!result.ok || result.data is! Map) {
+        final message =
+            result.error?.message ??
+            (result.ok
+                ? 'Unexpected settings response'
+                : 'HTTP ${result.statusCode}');
+        debugPrint(
+          'Failed to read ${server.baseUrl} settings '
+          '(${result.statusCode}): $message',
+        );
+        return (
+          settings: null,
+          error: message,
+          statusCode: result.statusCode,
+        );
+      }
       final map = Map<String, dynamic>.from(result.data as Map);
       final inner = map['data'];
-      return inner is Map ? Map<String, dynamic>.from(inner) : map;
-    } catch (_) {
-      return null;
+      return (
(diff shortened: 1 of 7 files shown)


krazyjakee:
Reviewed the diff (`accord_auth.dart`/`.g.dart`, `tos_gate.dart`, `accord_login.dart`, `auth_form.dart`, `add_server_dialog.dart`, `server_tos_gate_test.dart`) against the three areas in scope: code quality, file size/structure, and test coverage. (No 3D/rendering code is touched, so area 4 doesn't apply.)

**Code quality** — no bugs found.
- Verified `fetchServerSettings`'s new record return type has exactly one caller (`fetchTosConfig`), so the signature change is safe.
- Traced the `TosAvailability` enum through all four call sites (`accord_login.dart`, `add_server_dialog.dart`, `auth_form.dart`) — every comparison is exhaustive-in-effect (`advertised`/`unknown` checked explicitly, everything else falls through to "no gate"), no stale `tos.enabled`/`bool` references left anywhere (including `admin_settings_tab.dart`, which is an unrelated, unaffected code path for a server's own settings admin UI).
- The `result.error?.message ?? 'HTTP $statusCode'` fallback pattern matches every other call site in this file (`login`, `register`, `changePassword`, etc.) — consistent with existing convention, not a regression.
- `refused` vs `unknown` split, and the deliberate silence on `refused`, is well-reasoned and documented in the enum doc comment directly (guards against someone "fixing" it back into a visible warning later).

**File size/structure** — no action needed. Largest touched file is `accord_auth.dart` at 935 lines, under the 1000-line threshold; all others are well under.

**Test coverage** — good. The new `server_tos_gate_test.dart` covers all four `TosAvailability` states, both 401/403 for `refused`, 0/404/500/502 for `unknown`, the register-form rendering per state, and two end-to-end paths from a settings result through to the pumped widget. This matches the existing test style in this directory (e.g. `guest_session_test.dart`) — no test in this app-level suite mocks `AccordAuth`'s underlying HTTP layer directly, so not adding one here isn't a new gap.

Two very minor, non-blocking observations (not fixed — neither is a behavior bug, and both would be scope creep on a deliberately narrow fix):
- `AccordError.message` defaults to `''` rather than being nullable, so on a rare failure that has an `error` object with an empty message, the `?? 'HTTP $statusCode'` fallback won't kick in and the debug log/returned `error` string will just be empty. Cosmetic only — `fetchTosConfig` never reads `.error`, only `.statusCode`.
- The per-`baseUrl` fetch cache (`_tosFetchedServer`) means a transient `unknown` result (e.g. a network blip) sticks for the rest of that screen's lifetime with no retry short of restarting the flow. This mirrors the pre-existing caching behavior for `advertised`/`absent` and is an intentional tradeoff per the PR description, not a regression.

No fixes were needed, so no commit was pushed to this branch.

---

Issue #290, opened 2026-08-30
App Review 1.2: objectionable content can only be flagged on space messages — no reporting in DMs or on users


krazyjakee:
## App Review report

- Submission: `0c312f06-8042-4684-8684-2d3f94289cb1`
- Review date: 2026-08-30
- Reviewed version/build: `0.2.16 (160)`
- Device: iPad Air 11-inch (M3)
- Guideline: **1.2 — Safety, User-Generated Content**

Apple's required precaution is "a mechanism for users to flag objectionable content". The app has one, but only in a single place, so a reviewer exercising direct messages or a user profile finds nothing.

## Diagnosis

**Reporting exists only for messages inside a space.** `canReport` is gated on a non-null space in both entry points:

- `lib/features/messaging/views/message_pane/message_row.dart:373` (hover actions) — `canReport: !widget.isOwn && widget.spaceId != null`
- `lib/features/messaging/views/message_pane/message_row.dart:547` (long-press / right-click menu) — same gate

**Direct messages always pass a null space**, so the Report entry disappears entirely there:

- `lib/features/user/views/accord_direct_messages_conversations.dart:648` — `spaceId: null`

**There is no "report user" anywhere.** The member popout offers Block (`lib/features/member/views/accord_member_popout.dart:474`) but no report action, and it requires a non-null `spaceId` (`:24`), so it is not reachable from a DM either.

**The protocol backs only space-scoped reports.** `ReportsApi.create()` posts to `/spaces/{id}/reports` (`packages/accordkit/lib/src/rest/endpoints/reports_api.dart`), and accordserver's handler requires the reporter to be a member of that space (`src/routes/reports.rs`, "Verify user is a member of the space"). There is no account-level or DM report route (`src/routes/mod.rs:235-243`), and no admin-level report view across spaces.

Blocking, the other precaution Apple names, *is* implemented (member popout, and the DM relationship actions in `lib/features/user/views/accord_direct_messages.dart`) — reporting is the gap.

## Proposed fix

- [ ] Offer Report on every message, DMs included: drop the `spaceId != null` gate and give `showReportDialog` a null-space path.
- [ ] Add "Report user" alongside "Block user" in the member popout, and make the popout usable without a space so it works from a DM.
- [ ] Add Report / Block to the DM conversation header menu.
[...]

PR #300, opened 2026-09-02
fix(messaging): close the remaining gaps in reporting and blocking

krazyjakee:
Finishes #290 (App Review guideline 1.2 — "a mechanism for users to flag objectionable content"). PR #294 made reporting work outside a space; an audit found surfaces it still didn't reach and three ways the new dialog could mislead.

**A DM user tap was a dead end.** The member popout requires a space, so a DM falls back to the account-level profile, whose only action was Close. That profile now carries Report user and Block, reusing the same dialog and confirm flow the popout and the DM menu already use — a smaller, better-contained change than making the space-scoped popout (roles, nickname, kick/ban/timeout, permission lookups) work without a space.

**The DM report/block menu had no visible affordance** — it opened on long-press or right-click only, which is exactly the shape that reads as "there is no way to report". The 1:1 header now has an overflow button matching the group header's.

**Report reaches thread replies and pinned messages**, so no message surface is without it, and both filter the reported/blocked set the pane already filtered.

**Three dialog faults fixed:**
- It claimed *"Reported, but blocking the account failed"* even when the account-level route was missing and nothing had been filed. The copy now tracks what was actually delivered.
- That failure left the form live, so submitting again filed a second report. A `_sent` latch keeps the report from being sent twice while still allowing the block to be retried.
- The block promised *"their content is hidden"* while nothing in the client filtered it. `BlockedUsersController` holds the blocked account ids per connection — seeded from the relationship list on gateway READY, re-read on relationship events, updated locally by every block and unblock — and `MessageVisibility` applies it, with the hidden-message set, in the pane, the thread view and the pinned list.

The bottom-sheet menu became scrollable while doing this: nine entries overflowed a short viewport and clipped the last of them.

## Verification

- `flutter test` — 1139 passing, 0 failing, 0 skipped.
- `scripts/codegen.sh --check` — generated files match committed output.
- New tests: `test/features/user/dm_report_affordance_test.dart` (Report user from the DM profile, block confirmation, the visible header overflow reaching the report dialog), `test/features/messaging/pinned_messages_report_test.dart`, plus additions to `report_dialog_test.dart` (truthful copy for both delivered/not-delivered block failures, and the no-duplicate-report latch) and `dm_report_action_test.dart` (blocked-author filtering).

## Known limits

- No test for the thread-view Report entry — there is no thread-pane harness and building one is a larger lift than this issue needs; the pinned surface is covered instead.
- `MessageVisibility` matches blocked ids against `message.authorId` by exact string, so a federated author echoed back qualified would not be filtered. Same-server blocking, the case App Review exercises, works.
- Server-side work (an account-level report route, an admin surface for non-space reports) remains a separate accordserver task, per the issue.

Refs #290

### DIFF
--- a/docs/app-store-deploy.md
+++ b/docs/app-store-deploy.md
@@ -252,11 +252,19 @@ Two of those had gaps, both since fixed:
   alongside it when the server advertises one.
 - **Reporting only existed on space messages.** The action was gated on a
   non-null space, and a DM pane passes none. Report now sits on every message
-  and on user profiles and DM user menus, filing to the space's moderators
-  where there is a space and to the account-level `/reports` route where there
-  isn't.
+  surface — the pane, thread replies and the pinned list — and on user profiles
+  and DM user menus, filing to the space's moderators where there is a space and
+  to the account-level `/reports` route where there isn't. When neither route
+  took the report the confirmation says so instead of promising a review.
 
-Blocking was already in place (member popout, DM relationship actions).
+Blocking was already in place (member popout, DM relationship actions) and is
+now reachable from the two places a reviewer actually lands: the account-level
+profile a DM user tap opens, and a visible overflow button in the 1:1 DM header
+(the menu behind it was long-press/right-click only). A blocked account's
+messages are filtered out of every message surface — see `MessageVisibility`
+(`lib/features/messaging/utils/message_visibility.dart`) over
+`BlockedUsersController` — so the promise the block makes is one the client
+keeps rather than one only the server could.
 
 Do not re-gate the terms on a server response: nothing a signed-out client can
 read is guaranteed to be there, and the gate has to hold before any server is
--- a/lib/features/events/services/accord_event_handler.dart
+++ b/lib/features/events/services/accord_event_handler.dart
@@ -21,6 +21,7 @@ import 'package:bonfire/features/server/controllers/connections.dart';
 import 'package:bonfire/features/server/utils/space_cache.dart';
 import 'package:bonfire/features/settings/controllers/settings.dart';
 import 'package:bonfire/features/user/controllers/accord_users.dart';
+import 'package:bonfire/features/user/controllers/blocked_users.dart';
 import 'package:bonfire/features/voice/controllers/call.dart';
 import 'package:bonfire/features/voice/controllers/voice.dart';
 import 'package:bonfire/features/voice/controllers/voice_states.dart';
@@ -105,6 +106,14 @@ VoidCallback handleAccordEvents(
         serverKey: serverKey,
         homeDomain: selfDomain,
       );
+      // Blocked accounts drive the message filter that makes "their messages
+      // are hidden" true (#290), and READY doesn't carry relationships — so
+      // this connection's set is fetched here, for background connections too.
+      unawaited(
+        ref
+            .read(blockedUsersControllerProvider(serverKey).notifier)
+            .refresh(client),
+      );
       if (isActive()) {
         _seedVoiceStates(ref, data, serverKey: serverKey);
       }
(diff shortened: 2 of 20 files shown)


krazyjakee:
## Automated review

Reviewed against the three requested areas (code quality, file size/structure, test coverage). No 3D model/rendering code is touched by this PR, so that check doesn't apply.

I read all 18 changed files line-by-line (report dialog latch/retry logic, `BlockedUsersController`, `MessageVisibility` and its wiring into the pane/thread view/pinned list, the DM profile and header affordances, the event handler's relationship listeners, the scrollable context-menu sheet, and all four new/extended test files), then had an independent second pass done the same way rather than just skimming and agreeing with the first. No file exceeds 1000 lines (largest is `thread_view.dart` at 814), so no splitting was needed.

Two genuine issues survived that process; both are fixed in the pushed commit.

### Findings & fixes

1. **Medium — thread root bypassed the new visibility filter.** `lib/features/messaging/views/thread_view.dart`: `MessageVisibility.filter()` was applied to a thread's replies but not to the root message rendered at list index 0. So reporting a thread's root post (or blocking its author) hid it from the channel pane and pinned list as promised, but it kept showing at the top of its own already-open thread — the one message surface the filter didn't reach, contradicting both the in-app "hidden for you" copy and the PR's own "every message surface" description. Fixed by closing the thread pane when its root becomes filtered (mirroring the existing root-deleted behavior, but passing `null` rather than `ThreadResult.deleted()` so the forum board doesn't wrongly treat a hidden-for-you post as actually deleted). Added `test/features/messaging/thread_hidden_root_test.dart` covering blocked-author and reported-root cases, plus a control case that a visible root stays open.

2. **Low/Medium — race condition in `BlockedUsersController.refresh()`.** `lib/features/user/controllers/blocked_users.dart`: the gateway handler fires an unawaited `refresh()` on READY and again on every `onRelationshipAdd`/`Update`/`Remove` event, with nothing serializing them. With a burst of relationship events (or one landing right after READY), two `GET /users/@me/relationships` calls can be in flight together, and the response that resolves last wins via `sync()`'s full-set replacement — even if it was the request that was sent *first*, silently reverting a set a later, already-applied refresh had just written. Fixed with a generation counter (same pattern already used in `accord_discovery.dart`) that discards a response once a newer refresh has superseded it. Added `test/features/user/blocked_users_controller_test.dart`, which reproduces the out-of-order-response scenario with controlled `Completer`s and asserts the earlier-started, later-resolving response can't clobber state, plus baseline coverage for a normal single refresh and `sync()`'s no-op-on-unchanged-set optimization.

### Areas checked with no issues found
- Report dialog `_sent` latch / delivered-copy accuracy / retry-after-failed-block logic — correct, and the existing tests genuinely exercise the retry and failure-injection paths rather than just the happy path.
- `MessageVisibility` wiring in the message pane and pinned list — consistent and reactive.
- `context_menu.dart`'s `SingleChildScrollView` wrap for the bottom sheet — no layout/overflow regression.
- Null/empty-string edge cases (`readActiveServerKey() ?? ''`, `relationship.user?.id`) — all guarded correctly.
- `docs/app-store-deploy.md` — accurate against the diff (finding 1 above is what it inadvertently overstated).
- New/extended tests in `pinned_messages_report_test.dart`, `dm_report_affordance_test.dart`, `report_dialog_test.dart`, `dm_report_action_test.dart` — assert on real widget/request behavior, not just "dialog opens without crashing".

### Note on verification
This environment has no Flutter/Dart SDK available, so I could not run `flutter analyze`, `flutter test`, or `scripts/codegen.sh --check` myself. The fixes and new tests were validated by careful manual tracing against the actual provider/model signatures in `packages/accordkit` and by matching established test-harness and generation-counter patterns already used elsewhere in this codebase (`accord_users_controller_test.dart`, `accord_discovery.dart`), but please run the full suite before merging to confirm.

---

PR #302, opened 2026-09-02
fix(app-store): remove the pre-release surfaces a reviewer can reach

krazyjakee:
First of two parts for #292 (App Review guideline 2.2 — *"appears to be a pre-release, test, or trial version with a limited feature set … complete, remove, or fully configure any partially implemented features"*). That finding had no located root cause. An investigation ran the app against the live `chat.daccord.gg` at iPad Air viewports and found three literal partially-implemented surfaces on the reviewer's path; this fixes them. The iPad layout and empty-state findings are the second part.

**Settings → Account showed a permanent red "Failed to load connections."** `GET /users/@me/connections` returns 404 on the live server, and any non-`ok` became that error — so the first Settings screen a reviewer opened showed a red failure. There is also no "link account" affordance anywhere in `lib/`, so the section can never hold content. 404/405/501 is now treated as *unsupported* rather than *failed*: the embedded Account pane hides the card and its heading entirely, the standalone page states the capability plainly, and genuine failures (no client, network, 5xx) still show the red error.

**The GitHub self-updater was fully live in the App Store build.** `kAppStoreBuild` is correctly true for iOS (the release workflow passes `--dart-define=APP_STORE=true`), but it gated only `maybeCheckOnStartup()` and `canInstallInPlace`. The Settings → Updates tile, the "Check for updates" button, the available-release block and "View release" were all reachable. v0.2.17 was published to GitHub on 2026-08-27 and the review was 2026-08-30, so a reviewer tapping through Settings would have seen *"Update available: v0.2.17"* and a link to download the app from GitHub — which reads as "this store build is not the real distribution", and touches 3.1.1 / 2.4.5 as well as 2.2.

The whole feature is now gated, not just its UI: `check()`, `applyUpdate()`, `prepareUpdate()` and `platformAssetUrl()` are no-ops on store builds so GitHub is never queried, the banner collapses before reading any provider, and the Updates entry point is removed from both layouts. `UpdatesScreen` keeps the current version and "What's new" (release notes for the *running* build — not a release-availability check) and otherwise says updates come through the app store. **The sideload `github` flavor is unchanged**, with a regression test asserting the self-updater still works there.

**Developer Mode / the MCP server shipped enabled-able.** The Settings → Advanced toggle had no gate, and `mcp_server.dart` exports the live `dart:io` server under `if (dart.library.io)` — true on iOS — so a persisted `developerMode` flag would have opened a 127.0.0.1 listener inside the store binary. `mcp_server_controller.dart`'s doc comment claimed a desktop check that did not exist; `isDeveloperModeAvailable` (desktop **and** not a store build) now implements it, and gates the toggle in both the narrow list and the wide Advanced pane.

## Testing store-only paths

`kAppStoreBuild` stays a compile-time const, so the code is still statically dead in a real store build. Call sites read `isAppStoreBuild` = `kAppStoreBuild || (debugAppStoreBuild ?? false)` — the `||` still folds to `true` under `--dart-define=APP_STORE=true`, but tests can flip the override. This mirrors the existing `UpdateInstaller.debugInstallRootWritable` seam.

31 new tests: connections 404/405/501 → no error and card collapsed, 500/502 → error still shown, 200 → list and empty state intact; the update entry points proven to be no-ops with state untouched (a real HTTP call would flip `checkedOnce` either way); banner collapsed; Updates page stripped; Updates and Developer sections hidden in both layouts; MCP server with a positive control (binds when available) and the gate (port stays free when not).

## Verification

- `flutter analyze --no-fatal-infos` — 2 issues, both the pre-existing `unintended_html_in_doc_comment` infos in `packages/accordkit`.
- `flutter test` — 1161 passing, 0 failing.
- `scripts/codegen.sh --check` — clean.

Refs #292

### DIFF
--- a/lib/features/developer/controllers/mcp_server_controller.dart
+++ b/lib/features/developer/controllers/mcp_server_controller.dart
@@ -1,6 +1,7 @@
 import 'package:bonfire/features/developer/services/mcp_server.dart';
 import 'package:bonfire/features/developer/services/mcp_tools.dart';
 import 'package:bonfire/features/settings/controllers/settings.dart';
+import 'package:bonfire/shared/app_info.dart';
 import 'package:riverpod_annotation/riverpod_annotation.dart';
 
 part 'mcp_server_controller.g.dart';
@@ -40,8 +41,15 @@ const int _activityLogCap = 100;
 /// tool groups are read live by the server, so changing those takes effect
 /// without a restart.
 ///
+/// The desktop-only part is enforced here rather than assumed: [McpServer]
+/// resolves to the real `dart:io` implementation on *any* platform with
+/// `dart:library.io` — which includes iOS and Android — so without this gate a
+/// persisted `developerMode` flag would start a real HTTP listener on a phone or
+/// inside a store build. [isDeveloperModeAvailable] is the single source of
+/// truth, shared with the settings UI that offers the toggle.
+///
 /// On web (no `dart:io`) the [McpServer] facade is a no-op, so this controller
-/// is inert there.
+/// is inert there too.
 @Riverpod(keepAlive: true)
 class McpServerController extends _$McpServerController {
   McpServer? _server;
@@ -54,6 +62,7 @@ class McpServerController extends _$McpServerController {
   McpServerState build() {
     final settings = ref.watch(settingsControllerProvider);
     final shouldRun =
+        isDeveloperModeAvailable &&
         settings.developerMode &&
         settings.mcpEnabled &&
         settings.mcpToken.trim().isNotEmpty;
--- a/lib/features/developer/controllers/mcp_server_controller.g.dart
+++ b/lib/features/developer/controllers/mcp_server_controller.g.dart
@@ -15,8 +15,15 @@ part of 'mcp_server_controller.dart';
 /// tool groups are read live by the server, so changing those takes effect
 /// without a restart.
 ///
+/// The desktop-only part is enforced here rather than assumed: [McpServer]
+/// resolves to the real `dart:io` implementation on *any* platform with
+/// `dart:library.io` — which includes iOS and Android — so without this gate a
+/// persisted `developerMode` flag would start a real HTTP listener on a phone or
+/// inside a store build. [isDeveloperModeAvailable] is the single source of
+/// truth, shared with the settings UI that offers the toggle.
+///
 /// On web (no `dart:io`) the [McpServer] facade is a no-op, so this controller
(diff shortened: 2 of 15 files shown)


krazyjakee:
## Automated review

Reviewed against the three requested areas (code quality, file size/structure, test coverage); area 4 (3D model/visual review) doesn't apply here. Verified independently by installing Flutter beta locally and re-running the checks the PR description claims:

- `flutter analyze --no-fatal-infos` — same 2 pre-existing `unintended_html_in_doc_comment` infos in `packages/accordkit`, nothing new.
- `flutter test` — all 1161 tests pass (also ran the 121 tests in the touched suites individually first).
- `scripts/codegen.sh --check` — generated `*.g.dart` matches committed output.
- `dart format --set-exit-if-changed` on the touched files — flags pre-existing drift on files this PR didn't touch (confirmed against `origin/master`), nothing introduced by this diff.

### Findings

| # | Severity | File | Issue | Status |
|---|----------|------|-------|--------|
| 1 | Low (doc accuracy) | `lib/features/developer/controllers/mcp_server_controller.dart` | The new doc comment explaining the desktop-only gate wrote the conditional-import check as `` `dart:library.io` `` (colon). The actual syntax, used one file over in `mcp_server.dart:8`, is `if (dart.library.io)` (dot) — `dart:library.io` isn't valid Dart and would send a future reader searching for a directive that doesn't exist. | **Fixed** in 8d05d97, with the matching `.g.dart` doc copy regenerated via `build_runner` (hash unchanged — doc-only). |

No functional, correctness, or structural issues found:
- The `isAppStoreBuild` / `isDeveloperModeAvailable` gates are applied consistently at every entry point I could find (`_UpdatesSection`, `_DeveloperSection`, `UpdateBanner`, `UpdatesScreen`, `McpServerController`) — no un-gated path back to the GitHub updater or the MCP server on a store build.
- The pre-existing `kAppStoreBuild` use in `background_connection.dart` (gating the Android foreground service) is correctly left as the raw compile-time const rather than switched to `isAppStoreBuild` — that gate is about avoiding a crash from an undeclared manifest service, not an App-Review-visible surface, so it shouldn't follow the test-overridable seam.
- Connections 404/405/501-vs-500/502 handling, the store-build/no-updater UI branch, and the MCP desktop gate all have solid, well-isolated test coverage (positive + negative controls); I didn't find a risky path left uncovered.
- No file in the diff is anywhere near the 1000-line split threshold (largest touched file is 893 lines).

---

PR #303, opened 2026-09-02
fix(ui): stop an iPad reading as an unfinished build

krazyjakee:
Second of two parts for #292 (App Review guideline 2.2). [#302](https://github.com/DaccordProject/daccord/pull/302) removes the three half-built surfaces; this fixes the layout and empty states an iPad Air reviewer actually looked at. Findings come from running the app in a browser at iPad viewports against the live `chat.daccord.gg`.

**iPad portrait forced a three-pane desktop layout into an unusable message column.** `_wideLayoutBreakpoint = 720` sat just under the iPad Air's 820pt portrait width, so at 820: rail 72 + channel list ~219 + member list 240 left **~280pt for messages** — narrower than the member list. Body text wrapped to three or four words a line and the composer placeholder wrapped onto two. The phone at 393pt looked strictly better because it used the drawer layout.

Replaced with `resolveHomeLayout(width, preferredChannelListWidth)`, a pure function returning wide / medium / compact:

```
kHomeSidebarBreakpoint    = 72 + minChannelListWidth(180) + 8 + 420 = 680
kHomeMemberListBreakpoint = 680 + 240 = 920
```

Derived from the panes rather than from a device width, so the thresholds can't drift from the widths actually rendered, and panes are given up in the order the screen needs them least (roster to the end drawer first, then the channel list). A user who dragged the divider wide gets the list squeezed back toward its minimum rather than thrown into a drawer. 1180 → wide (640pt messages), **820 → medium (520pt)**, 1024 → wide, 768 → medium, 393 → compact. `kOnboardingWideBreakpoint` now aliases `kHomeSidebarBreakpoint`.

**The signed-out funnel was three near-empty screens on a 1180x820 canvas.** Welcome was a logo, one sentence and one button in ~700px of black. Above 620pt it now carries three value-prop cards (claims checked against `README.md`) and a full-weight "Run your own server" button opening a shared `showSelfHostingDialog` describing the two paths from `docs/self-hosting/overview.md`. Phone layout is untouched except that button, placed last so nothing above it moves.

**The directory shows one listing** (`master.daccord.gg` returns `"total":1`) followed by ~570px of black. Sparse unfiltered results now carry an honest "this list is short on purpose" note, and the connect-by-URL / host-your-own footer moved into `AccordDiscoveryBody` so all three hosts get it — including the discovery *dialog*, which had no manual-connect path at all. **Bug fixed:** `_error` was only painted inside the `listings.isEmpty` branch, so a failed Join on the one visible listing rendered nothing.

**A voice channel at rest was a ~60px strip.** Both presentations now render a pre-join card: channel name, participant summary, read-only mic/camera/screen-share chips, what joining does. Worth noting — the voice view's own `_sidePanelBreakpoint = 720` meant an iPad **always** got the compact strip, so fixing only the non-compact branch would have changed nothing on the rejected device. No transport logic touched.

## Verification

- `flutter analyze --no-fatal-infos` — 2 issues, both the pre-existing `unintended_html_in_doc_comment` infos in `packages/accordkit`.
- `flutter test` — 1146 passing, 0 failing (cold and warm runs identical), including 16 new tests.
- `scripts/codegen.sh --check` — clean; no `@riverpod` file touched.
- Screenshots taken at 1180x820, 820x1180, 1024x768 and 393x852 before and after each change.

## Notes

- `AccordDiscoveryBody` gained an `onManualConnect` parameter and an always-rendered footer; all three call sites are updated.
- Left alone deliberately: the Discover Servers *dialog* still reserves its full 600pt height, so a one-listing directory leaves black below the footer inside the dialog frame. Tightening it means unpicking the `Expanded`/`mainAxisSize` chain and risks the long-list case.

Refs #292

### DIFF
--- a/lib/features/authentication/views/accord_login.dart
+++ b/lib/features/authentication/views/accord_login.dart
@@ -353,7 +353,10 @@ class _AccordLoginScreenState extends ConsumerState<AccordLoginScreen> {
               ? () => context.push('/switcher')
               : null,
         ),
-        maxWidth: 480,
+        // Wide enough for the welcome screen's three-up highlights on a tablet
+        // or desktop canvas; it collapses itself back to a phone layout below
+        // `kWelcomeWideBreakpoint` (#292).
+        maxWidth: 760,
       ),
       _LoggedOutView.browse => _BrowseView(
         onBack: _goBack,
@@ -635,17 +638,12 @@ class _BrowseView extends StatelessWidget {
                 ],
               ),
             ),
+            // The connect-by-URL / host-your-own footer lives inside
+            // [AccordDiscoveryBody] so the dialog variant gets it too (#292).
             Expanded(
               child: AccordDiscoveryBody(
                 onJoinRequiresAuth: onJoinRequiresAuth,
-              ),
-            ),
-            Padding(
-              padding: const EdgeInsets.fromLTRB(16, 4, 16, 16),
-              child: TextButton.icon(
-                onPressed: onManualConnect,
-                icon: const Icon(Icons.link, size: 18),
-                label: const Text('Connect to a server by URL'),
+                onManualConnect: onManualConnect,
               ),
             ),
           ],
--- a/lib/features/authentication/views/welcome_view.dart
+++ b/lib/features/authentication/views/welcome_view.dart
@@ -1,10 +1,61 @@
+import 'package:bonfire/shared/components/self_hosting_dialog.dart';
 import 'package:bonfire/theme/theme.dart';
 import 'package:flutter/material.dart';
 
-/// First-run / signed-out landing screen: daccord branding, a short pitch, and
-/// the primary path into the public server browser. The Flutter port of the
-/// reference client's `welcome_screen`, kept deliberately compact so choosing
-/// a server remains visible on narrow devices.
+/// Width at or above which the welcome screen lays its value props out
+/// three-up. Below it they are dropped entirely so a phone keeps the
+/// "Browse Servers" button above the fold.
+const double kWelcomeWideBreakpoint = 620;
+
+/// What the app actually does, in the order a newcomer cares about. Kept as
+/// data (and public) so a test can assert the pitch stays on screen without
+/// matching prose in the widget tree.
+@immutable
+class WelcomeHighlight {
+  const WelcomeHighlight({
+    required this.icon,
+    required this.title,
+    required this.body,
+  });
+
+  final IconData icon;
+  final String title;
+  final String body;
+}
+
+/// Every claim here is a shipped feature — see the feature list in `README.md`.
+const List<WelcomeHighlight> kWelcomeHighlights = <WelcomeHighlight>[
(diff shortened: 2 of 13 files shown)


krazyjakee:
## Automated review

Reviewed the diff (12 files, +1239/-182) against code quality, file size, and test coverage. No `flutter`/`dart` toolchain was available in this environment, so this is a careful manual read rather than a run of `flutter analyze`/`flutter test` — worth a normal CI pass to confirm.

### Findings

**1. File size — `lib/features/voice/views/voice_view.dart` crossed 1000 lines (Medium, fixed)**
The pre-join card rewrite grew the file from 968 → 1099 lines. Extracted the self-contained lobby unit (`_LobbyBody` → `VoiceLobbyBody`, `_PreJoinChip`, `_LobbyAvatar`) into a new `lib/features/voice/views/voice_lobby.dart`, dropping `voice_view.dart` to 771 lines. Pure move — no behavior change, both call sites updated, unused imports (`accord_channels.dart`, `collection.dart`'s `IterableExtension`) trimmed from `voice_view.dart` since they were only needed by the extracted code. `test/features/voice/voice_lobby_test.dart` already drives this through the public `VoiceChannelView` API, so it isn't affected by the move. Pushed as `34bbb97`.

**2. Everything else checked out**
- `resolveHomeLayout`'s breakpoint math (`kHomeSidebarBreakpoint`/`kHomeMemberListBreakpoint`) is internally consistent at every boundary I traced by hand, and matches the new `home_layout_test.dart` coverage (iPad Air 820/1180, iPad 768/1024, phone 393, the pane-drop order, and the "never below `kMinMessagePaneWidth`" sweep).
- All three `AccordDiscoveryBody` call sites (`accord_login.dart`, `accord_discovery.dart`'s dialog, `add_server_dialog.dart`) now pass a correct `onManualConnect` — the dialog hands off to Add-Server instead of stacking dialogs, and Add-Server's own callback (`_tabs.animateTo(0)`) correctly targets the "Enter URL" tab (index 0 in the `TabBarView`).
- The medium-layout `Scaffold`/`endDrawer` restructuring in `accord_home.dart` (rail+channels+messages inline, roster in an end drawer) is well-formed — single `_scaffoldKey` used in only one live branch at a time, `messageArea`'s members button only shows when the roster isn't inline.
- `voice_view.dart`'s new pre-join card correctly threads `channelName` through both the full and compact (`VoiceChannelView`) call sites, and `selfMute`/`kIsWeb` usage is sound.
- Self-hosting dialog content and the welcome screen's feature claims both check out against `README.md` and `docs/self-hosting/overview.md`.
- No test coverage gaps found for the changed behavior beyond the file-size fix above — `welcome_view_test.dart` and `accord_discovery_test.dart` cover the new branches (tablet highlights + self-host CTA, sparse-directory note, failed-join error placement).
- No 3D model/rendering code is touched by this PR, so the visual-review requirement doesn't apply here.

No other issues found. Ranked: (1) file size — fixed; nothing else risked correctness or regressions.

---

Issue #306, opened 2026-09-02
Silent failures and dead controls surfaced by the guideline 2.2 audit


krazyjakee:
Split out of #292. The App Review 2.2 audit fixed everything on the reviewer's path; these are the lower-priority findings from the same sweep. None are reachable in fifteen minutes on a fresh account, so they did not hold #292 open, but each is a small isolated change and collectively they are the same class of defect Apple's "partially implemented features" wording describes.

File and line references are from 2026-09-02 and may have drifted.

## Unbounded requests

- `packages/accordkit/lib/src/rest/accord_rest.dart:45` uses a bare `http.Client()` with **no timeout anywhere**, so every spinner in the app is unbounded. A slow or black-holed server produces an indefinite loading state rather than an error. This is the root cause behind several "screen looks broken" symptoms and is probably the highest-value item here.

## Panes that cannot distinguish "loading" from "failed"

`lib/features/member/views/accord_member_list.dart:73-101` already has the right pattern — a `membersLoadFailedProvider` plus `ServerUnreachable(onRetry:)`. Three panes still overload `null` as both states, so a failed load renders as a permanent spinner:

- `lib/features/messaging/controllers/accord_messages.dart:64-70`
- `lib/features/channels/controllers/accord_channels.dart:26-32`
- `lib/features/events/services/accord_event_handler.dart:977-980`

## Optimistic actions that revert without saying why

Each of these mutates local state, fails, and silently rolls back:

- reactions — `accord_messages.dart:363-373`
- pin / unpin — `accord_messages.dart:176-200`
- channel mute — `lib/features/channels/controllers/muted_channels.dart:38-48`
- friend accept / remove / block — `lib/features/user/views/accord_direct_messages_friends.dart:44,54,66` (an `_error` field is declared **and rendered** but never assigned)
- thread delete — `lib/features/messaging/views/thread_view.dart:193`

## Dead or misleading controls

- `lib/features/spaces/views/accord_home_rail.dart:463` — `onTap: () {}` on hidden-server icons.
- `lib/features/voice/views/voice_settings_screen.dart:119-128` — the output-device dropdown renders on iOS but the write is swallowed, so the control appears functional and does nothing.
- `lib/features/settings/models/accord_settings.dart:145` — `backgroundConnection` has a reader and no setter, so the Android foreground service it controls is unreachable from the UI.

## Latent accordkit inconsistency

[...]

PR #307, opened 2026-09-02
fix(ui): bound REST requests and stop four silent failures

krazyjakee:
First half of #306 — findings from the App Review guideline 2.2 audit that didn't sit on the reviewer's path but are the same class of defect ("complete, remove, or fully configure any partially implemented features"). The other half is a separate PR.

**Every request in the app was unbounded.** `AccordRest` used a bare `http.Client()` with no timeout anywhere, so a slow or black-holed server produced an indefinite spinner rather than an error — the root cause behind several "screen looks broken" symptoms. Every attempt is now bounded by a configurable `timeout` (default 30s), threaded through `AccordClient(requestTimeout: …)` the way `httpClient`/`sleep`/`onUnauthorized` already are. A timeout surfaces as a normal `RestResult` failure, never a thrown `TimeoutException`, so it flows through every caller's existing error path.

*Per attempt, not across retries*: a 429 retry spends most of its wall clock asleep in the server-dictated `Retry-After` pause, so a shared budget would turn a healthy rate-limit wait into a bogus timeout. Per-attempt still bounds the whole call. A test asserts a second attempt gets a full budget after a slow 429.

**Three panes couldn't tell "loading" from "failed"** — they overloaded `null` as both, so a failed load rendered as a permanent spinner. The member roster already had the right pattern (`membersLoadFailedProvider` + `ServerUnreachable(onRetry:)`); it's factored into a shared `LoadFailed` family keyed `(scope, serverKey, id)`, with each feature exposing a named accessor. `membersLoadFailedProvider` keeps its exact call-site signature, so member views and tests are unchanged, and messages, channels and spaces now reuse one provider rather than three copies.

**Two optimistic actions reverted silently.** `pin`, `unpin` and `toggleReaction` now return `String?` — null on success, else the error — mirroring `_createMessage` in the same controller, and the call sites report it with `showInfoSnack`, the mechanism the adjacent delete action already uses. The optimistic update and rollback are unchanged.

**`DirectoryApi` double-prefixed the API base path.** It prepended `AccordConfig.apiBasePath` while `AccordClient` already builds its REST client with `config.apiUrl()`, so `client.directory` would have hit `/api/v1/api/v1/directory`. Latent rather than live — discovery deliberately constructed a bare `AccordRest` to dodge it — but exposed and broken for anyone who reached for it. Convention settled the way every other endpoint group already works: **endpoint paths are bare, the `/api/v1` prefix belongs to the base URL**. The discovery comment no longer documents a workaround; it now explains why `AccordClient` isn't used there (a gateway and voice manager for one anonymous GET).

## Verification

Run from the repository root after the final edit (a worktree checkout does not reach `packages/accordkit`, so root-tree numbers are the ones that count):

- `flutter test` — **1219 passing**, 0 failing (1203 baseline + 16 new).
- `packages/accordkit` — `dart test` **180 passing** (174 baseline + 6 new).
- `scripts/codegen.sh --check` — generated files match committed output, verified from a cold `.dart_tool/build`.

## Risk

The 30s timeout is the widest-reaching change: nothing in-tree relies on a request outliving it (the member loader's own guard is tighter at 20s), but it changes behaviour for every network call.

One sharp edge is inherited from the existing member-list pattern rather than introduced here: because the failure flag outlives the autoDispose cache controller, re-entering a channel or space that previously failed briefly shows the error before the fresh load clears it. Kept for parity rather than diverging.

Refs #306


krazyjakee:
Automated review of this PR (focus: code quality, file size/structure, test coverage; no 3D/rendering code is touched here). Overall this is a clean, well-scoped fix: the `LoadFailed` extraction is a good factoring of the pre-existing `MembersLoadFailed` pattern, the per-attempt timeout is the right call given the 429/`Retry-After` semantics, the `DirectoryApi` double-prefix fix is correct and well-tested, and the new tests are thorough (failure vs. empty vs. retry, for each of channels/messages/spaces).

## Findings

**1. Medium — multipart uploads inherited the 30s general-request timeout (fixed, pushed)**

`AccordRest.makeMultipartRequest` goes through the same `_executeWithRetry` as every other call, so it picked up the new 30s `timeout` too. Unlike a normal API call, a file/attachment upload's duration scales with payload size and the caller's connection speed, not server latency — there's no size cap on attachments in this repo, so a slow-link or larger upload that previously had no deadline at all could now fail with a spurious timeout well before it would have finished. The PR's own risk section calls out the 30s timeout as "the widest-reaching change" but doesn't flag uploads specifically.

Pushed a fix (commit `c013c36`) that gives multipart requests their own budget: `AccordConfig.defaultUploadTimeout` (5 min), threaded through `AccordRest.uploadTimeout` / `AccordClient(uploadTimeout: …)`, independent of `requestTimeout`. Added tests confirming an upload survives past the shorter general timeout, still times out as an ordinary `RestResult` failure once its own budget elapses, and defaults/overrides work the same way `requestTimeout` does. Also documented in the accordkit README.

**2. Low — `lib/features/events/services/accord_event_handler.dart` is over 1000 lines**

Pre-existing (1061 lines on `master`, 1087 after this PR's small addition) rather than introduced here, and the PR only adds one self-contained provider plus doc comments to it. I didn't attempt a split: this file is the single gateway dispatcher wiring ~50 event streams to feature controllers, a structural refactor carries real risk of breaking event routing, and I have no Dart/Flutter SDK in this review sandbox to validate a refactor of that size (see note below). Worth a dedicated follow-up PR rather than folding into this one.

## Verification note

This sandbox has no Dart/Flutter SDK available, so I could not run `flutter analyze` / `flutter test` / `dart test` myself. I reviewed the diff and my own change by hand (brace-balance and cross-reference checks) but the PR author's reported numbers (`flutter test`: 1219 passing; `packages/accordkit`: 180 passing) should be re-verified by CI against the new commit, particularly the 4 new/changed accordkit tests.

---

PR #308, opened 2026-09-02
fix(ui): explain the failures and drop the dead controls guideline 2.2 found

krazyjakee:
Second half of #306, completing it. [#307](https://github.com/DaccordProject/daccord/pull/307) covered the data layer; this covers the controls. Apple's 2.2 wording was "complete, remove, or fully configure any partially implemented features" — these are the leftovers of that class.

## Three optimistic actions that reverted silently

All now report through the mechanisms already used nearby (`showInfoSnack` / `showErrorSnack` / `InlineError`), keeping the optimistic update and rollback unchanged.

- **Channel mute.** `setMuted` returned a `bool` that conflated "server rejected" with "another tap is still in flight", so naively reporting failure would have raised a false alarm on a double-tap. It now returns `MuteResult { ok, failed, busy }`, and a shared `toggleChannelMute()` snacks on `failed` only. Both call sites go through it so the wording can't drift.
- **Friends accept / remove / block.** The `_error` field was declared *and rendered* but never assigned — dead reporting machinery sitting right there. A `_mutate()` helper now sets it from the server's own message where there is one, and clears it on the way in so a successful retry doesn't leave a stale complaint.
- **Thread / post delete.** `if (!mounted || !result.ok) return;` became a real error, with the thread left open.

## Three dead controls

**Hidden-servers icon (`onTap: () {}`) — affordance removed.** The only plausible meaning of a tap is "select this space", and selecting a hidden space *is* unhiding it, which the row already offers as an explicit Unhide button one widget away. `_SpaceIcon.onTap` is now nullable and the sheet passes `null`, so the icon is identification only.

**iOS audio-output picker — not offered on iOS, kept everywhere it works.** Checked the stack rather than assuming: on desktop `selectAudioOutput` is a real `setOutputDevice`; on **Android** it round-trips `AudioSwitchManager` names and genuinely switches earpiece/speaker/wired/Bluetooth; on **iOS** it degrades to `overrideOutputAudioPort`, which recognises nothing but the literal id `"Speaker"`, and `enumerateDevices` returns only the route already in use plus a synthetic Speaker entry — so the picker could never offer a device you weren't already on, and "System default" was dropped entirely. LiveKit's own `Hardware.selectAudioOutput` refuses off-desktop for the same reason. iOS owns routing via Control Centre / AirPlay / headset.

Per CLAUDE.md, voice must not be stubbed or hidden — **no capability was removed**. Input device, volumes, sensitivity, mic test, AFK, camera and screen share are untouched on every platform, and the picker still ships on desktop and Android. What's gone is a control that never worked on the one platform it was removed from.

**`backgroundConnection` — made reachable.** The service works and is wired into `MainWindow` and the `github` flavor's manifest; only the setter and the UI entry were missing. Added `setBackgroundConnection` and a "Stay connected in the background" switch, both gated on a new `isBackgroundConnectionAvailable` (Android && !store) that `BackgroundConnectionController` now uses too — so the toggle can't appear on a Play build whose manifest omits the service, and the two gates can't drift.

## Verification

Run from the repository root after the final edit:

- `flutter analyze --no-fatal-infos` — identical to master run back-to-back under the same conditions; no new issues.
- `flutter test` — **1233 passing**, 0 failing (1219 baseline + 14 new).
- `scripts/codegen.sh --check` — generated files match committed output, verified from a cold `.dart_tool/build`.

Docs updated: `docs/voice-and-video/voice-channels.md`, and `docs/complexity-audit.md`, whose B2 entry described a toggle that did not exist.

## Known gap

The hidden-servers icon has no dedicated widget test. `_SpaceIcon` and the sheet are private to the `accord_home.dart` `part` library, the repo has no harness that pumps the real rail (`mobile_pop_scope_test` and `drawer_swipe_area_test` both use hand-built replicas), and extracting the icon would ripple through five other call sites. The change removes a callback and adds no behaviour, so the only assertable property is the absence of a handler — which the nullable type enforces at the call site.

Refs #306

### DIFF
--- a/docs/complexity-audit.md
+++ b/docs/complexity-audit.md
@@ -277,6 +277,13 @@ and the full test suite (431 tests) staying green.
 - **B2 [Δ fix]** — `backgroundConnectionControllerProvider` wired into
   `MainWindow` per its own doc comment; the Android "Background connection"
   settings toggle previously did nothing.
+- **B2 follow-up (#306)** — the settings *toggle* itself was missing:
+  `AccordSettings.backgroundConnection` had a reader and no setter, so the
+  wired-up service was unreachable from the UI. Added
+  `SettingsController.setBackgroundConnection` and a Notifications-section
+  switch, both gated on the new `isBackgroundConnectionAvailable` (Android,
+  non-store) that `BackgroundConnectionController` now shares, so the switch
+  cannot appear where the service does not exist.
 - **F1** — the message-pane cluster (list/row/composer/reactions/attachments/
   mute button) moved out of `spaces/views/` into
   `messaging/views/message_pane/` as one library with `MessagePane` as its
--- a/docs/voice-and-video/voice-channels.md
+++ b/docs/voice-and-video/voice-channels.md
@@ -47,3 +47,5 @@ Click the **Camera** button to share your camera feed. Other participants see yo
 ## Voice Settings
 
 Access voice and video settings from the voice bar settings button or from **App Settings > Voice & Video**. Here you can configure your input/output devices and other audio preferences.
+
+The **Output device** picker is shown on desktop and Android, where the platform can honour an explicit choice. iOS routes call audio itself — use Control Centre, the AirPlay picker, or plug in a headset — so the app does not offer a picker that iOS would ignore.
--- a/lib/features/channels/controllers/muted_channels.dart
+++ b/lib/features/channels/controllers/muted_channels.dart
@@ -3,6 +3,24 @@ import 'package:riverpod_annotation/riverpod_annotation.dart';
 
 part 'muted_channels.g.dart';
 
+/// Outcome of [MutedChannelsController.setMuted].
+///
+/// A plain `bool` could not tell "the server rejected it" apart from "another
+/// tap for this channel is still in flight", so every caller had to treat the
+/// in-flight case as a failure — or, as they did, say nothing at all (#306).
+enum MuteResult {
+  /// The server accepted the change; the optimistic state stands.
+  ok,
+
+  /// The server rejected it (or there was no client); the state rolled back and
+  /// the caller must tell the user.
+  failed,
+
+  /// Ignored because the same channel is already being updated. Not a failure —
+  /// the in-flight request still owns the outcome.
+  busy,
+}
+
 /// Server-backed channel mutes for one connected account.
 ///
 /// Both the channel header and context menu use this cache so opening either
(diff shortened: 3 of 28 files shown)


krazyjakee:
## Automated review

Reviewed the full diff (28 files, +771/-45) against the four areas below. No fixes were needed, so nothing was pushed.

**Code quality** — Traced every changed call site end-to-end rather than just the diff hunks:
- `MutedChannelsController.setMuted` → `MuteResult {ok, failed, busy}` is correctly threaded through both call sites (`channel_context_menu.dart`, `mute_button.dart`) via the new shared `toggleChannelMute()` helper. `busy` is deliberately not reported, matching the stated intent ("another tap is still in flight" isn't a failure).
- `_FriendsTabState._mutate()` correctly clears `_error` on entry and sets it from `RestResult.errorMessageOr(...)` on failure only; `InlineError` and the field were already wired up, just never assigned before.
- `thread_view.dart`'s `_deleteRoot` now uses `showErrorSnack` (has a `RestResult` to pull the server message from) while `toggleChannelMute` uses `showInfoSnack` with a fixed string (no `RestResult` available, since `setMuted` intentionally doesn't leak one) — checked this isn't a stray inconsistency: `showInfoSnack` is already the general-purpose snack helper used for both success and failure toasts throughout the codebase (`message_row.dart`, `channel_management.dart`, etc.), so this matches convention.
- `_SpaceIcon.onTap` → `VoidCallback?`: verified the only caller passing `null` (`accord_home_rail.dart`'s hidden-servers sheet) sits next to an explicit "Unhide" `TextButton`, and `GestureDetector(onTap: null)` is a valid no-tap-handler state, not a crash risk.
- `canPickAudioOutputDevice` / `isBackgroundConnectionAvailable`: both are single sources of truth shared between the settings toggle and the platform code that acts on it (`VoiceSettingsScreen`, `BackgroundConnectionController`), so the two can't drift, per the PR's stated goal.
- All new imports resolve and are used (double-checked `toggle_channel_mute.dart`'s import into `message_pane.dart` — it's needed there because `mute_button.dart` is a `part of` that library, not because `message_pane.dart` itself calls it).
- `AccordSettings.backgroundConnection` already had `toJson`/`fromJson`/`copyWith` wiring before this PR (only the setter/UI were missing, matching the PR description) — confirmed no serialization gap.

**File size** — No changed file exceeds 1000 lines (`accord_settings_screen.dart` 907, `accord_home_rail_tiles.dart` 879, `thread_view.dart` 831 are the largest; none newly crossed a threshold that would warrant a split from this diff).

**Test coverage** — All four behavior changes have new, targeted tests: `toggle_channel_mute_test.dart` (mute failure/success/busy), `thread_delete_error_test.dart` (refused delete keeps the thread open), `friends_action_error_test.dart` (server-message vs. fallback wording), `background_connection_setting_test.dart` (setter persistence + platform gating), plus the existing `voice_settings_screen_test.dart` extended for `canPickAudioOutputDevice`. The Hive-backed persistence test follows the repo's established pattern (`settings_controller_test.dart`).

**3D model / visual changes** — N/A, this is a chat client diff with no 3D rendering involved.

**CI** — Watching this PR's run; `Vendored accordkit`, `Vendored markdown_viewer`, `Protocol integration`, and `Integration (client ↔ accordserver)` have already passed. The blocking `Analyze & Test` job (which runs `flutter test` against the ~14 new tests) was still in progress at review time — I'll follow up if it comes back red.

---

Issue #304, opened 2026-09-02
Store media: iPad screenshots still depict the phone layout — needs a seeded capture space


krazyjakee:
Follow-up from #291. The 2.3.10 rejection itself is cleared (platform badges gone, status bar cropped out, nothing half-visible), but one item on that issue's checklist is genuinely outstanding and should not be lost.

## The gap

`store-media/ios-ipad-13/` now renders at an iPad-proportioned frame, but the pixels inside are still the app's **phone** layout. After the layout work in #292, a real iPad Air renders four panes at 1180pt — space rail, channel list, ~640pt message column, 240pt member roster — which the shipped iPad set does not depict. `ipad-02` also carries a ~14% dead band where a message pane would be; real three-pane captures would remove it structurally.

## Why it wasn't done in #291

Real captures were attempted against the live `chat.daccord.gg` at 1180x820. The layout photographed well; **the content was unusable**:

- `#general` contained `"good thanks! just trying to get daccord on the ios store"` — a message about getting the app onto the App Store, inside the App Store screenshot — plus `"verity die"` and leftover agent test scaffolding.
- The member roster reads `OFFLINE — 100` above junk accounts (`aa`, `34343434`, `1423424345t5`, `aidsonaburgerbun`, and one `null`).
- Most seriously: the default landing channel is `#rules`, and the server's **rule 10 reads "Discord's-not-the-point rules still apply"** — a third-party platform name in ~40px type in the centre of the message column. Capturing it would trade one accurate-metadata problem for a strictly worse one.

Shipping that would have been worse than the phone-layout mockups. Fabricating a fake three-pane mockup was also rejected as dishonest.

## What a capture pass needs

A purpose-built space, not the public instance:

- **5–7 spaces** so the rail reads as multi-server (a fresh account sits in one, making the multi-server scene impossible).
- **Categorised channels** with one unread badge and one populated voice channel.
- **6–8 written `#general` messages** from named accounts with avatars, covering a reply, an image attachment, reactions and a mention.
- **~15 named members, mostly online**, grouped Owner / Moderators / Members, with one profile card open.
- **Ownership of the space**, so Roles and permissions are reachable.

Two constraints: a fresh account is **auto-joined to the public space** and must leave before capture; and the landing channel must not contain content naming another platform.

[...]

Issue #291, opened 2026-08-30
App Review 2.3.10: store screenshots advertise other platforms and carry a non-iOS status bar


krazyjakee:
## App Review report

- Submission: `0c312f06-8042-4684-8684-2d3f94289cb1`
- Review date: 2026-08-30
- Reviewed version/build: `0.2.16 (160)`
- Device: iPad Air 11-inch (M3)
- Guideline: **2.3.10 — Performance, Accurate Metadata**

> The app or metadata includes information about third-party platforms … Revise the app's screenshots to remove non-iOS status bar images.

Both halves of that come from the generator in `store-media/ios-generator/`, so the fix is in-repo followed by a re-upload in App Store Connect.

## Diagnosis

**Third-party platform references.** Scene 6 of `store-media/ios-generator/template.html` renders a badge row reading `iOS`, `Android`, `Desktop`, `Web` (`SCENES[6].badges`). Those are exactly the "information about third-party platforms" the guideline names, and they appear in the shipped `store-media/ios-iphone-6.5/iphone-06.png` and `store-media/ios-ipad-13/ipad-06.png`.

**Non-iOS status bar.** Every inner capture in `store-media/ios-generator/inner/` (`t-01.png` … `t-06.png`, 740×1542) has a status bar painted into the image — 9:41 plus signal/Wi-Fi/battery glyphs that are not iOS's. The template composites those straight into the device frame, so all 12 shipped screenshots show it.

**iPad screenshots are phone-shaped.** `template.html` uses one `.screen` aspect ratio (`740 / 1542`) and the same six phone captures for both devices, so the 2048×2732 iPad set is an iPhone-proportioned mockup — on a review device that is an iPad Air, that reads as inaccurate on its own.

## Proposed fix

- [ ] Drop the platform badge row from scene 6 (and audit the remaining copy for non-iOS platform mentions).
- [ ] Remove the painted status bar from the inner captures — crop it out rather than redraw it, so no status bar is shown at all.
- [ ] Give iPad its own captures and frame aspect instead of reusing the phone set.
- [ ] Re-render all 12 via `store-media/ios-generator/render.sh` and commit the refreshed `ios-iphone-6.5/` and `ios-ipad-13/` sets.
- [ ] Upload the new screenshots in App Store Connect (Previews and Screenshots → View All Sizes in Media Manager) — this is metadata, so no build is required.
- [ ] Note the constraint in `docs/app-store-deploy.md` so the next regeneration does not reintroduce it.

[...]

PR #309, opened 2026-09-02
feat(store): capture the real iPad layout for the App Store screenshots

krazyjakee:
Closes the outstanding item on #304 — carried over from #291, where the 2.3.10 rejection was cleared but the iPad set still depicted the app's **phone** layout inside an iPad-shaped frame. After #303, a real iPad renders four panes; the shipped screenshots did not.

`store-media/ios-ipad-13/` is now six genuine renders of the app's four-pane tablet layout — space rail, channel list, message column, member roster — at a 13" landscape canvas, produced **entirely offline**.

## How

A capture harness under `tool/store_capture/`:

- `capture_app.dart` — a web entry point booting the shipped widgets against an in-memory `MockClient`. No socket is opened and `login()` is never called. One scene per `?scene=N`, each driving itself through the app's own affordances (open a tab, open the member popout, open the roles dialog, fold the roster away, tap a role).
- `seeded_space.dart` — the invented fixture: 6 spaces, categorised channels, an unread badge, a populated voice channel, 15 members with presence, and conversations carrying a reply, a mention and reactions.
- `shoot_scenes.dart` — a dependency-free Chrome DevTools Protocol driver.
- `verify_store_shots.dart` — IHDR size and edge-flatness checks.
- `store-media/ios-generator/capture-inner.sh` — serves the app and shoots the set.

The template's iPad frame is now landscape over its own `tab-0N.png` set, with both `focus` overrides dropped; neither frame crops any more, so nothing can meet an edge.

**No live host was contacted at any point.** The fixture is served in-process.

## Why not a widget test

That was the first approach and it does not work: `flutter test` hardcodes `--use-test-fonts`, which resolves every *unstyled* `TextStyle` to a box font. Message bodies are unstyled — `markdown_viewer`'s renderer names no family and `RichText` doesn't inherit `DefaultTextStyle` — so message text rendered as boxes. Registering a real font under `Ahem`/`FlutterTest` doesn't displace it, and `flutter run -d flutter-tester` has no font manager without `/usr/share/fonts`. The whole trail is written into `docs/app-store-deploy.md` so nobody repeats it.

## Guard

`test/store_capture/seeded_content_test.dart` is the only piece that runs in `flutter test`: a string scan banning third-party platform names and store/review talk from the fixture. It caught a real one — a space named "Signal Garden", since renamed. This is the defect class that produced the 2.3.10 rejection and that made the live instance unusable as a capture source, so it is worth a test.

## Verification

From the repository root: 6x **2048x2732** read from PNG IHDR chunks; zero colour transitions on every capture's four edge lines; all six framed renders and all six inner captures inspected by eye. `flutter test` **1235 passing** (1233 baseline + 2). `scripts/codegen.sh --check` clean. `flutter analyze --no-fatal-infos` identical to master.

## Known limits, deliberately not papered over

- **The iPhone set is untouched** — `ios-iphone-6.5/` still holds the older illustrative phone compositions. Replacing them with real compact-layout captures is now a small extension of this harness.
- **`ipad-06`** (signed-out welcome) is the weakest of the six: a 760pt column centred in a 1366pt canvas leaves wide margins. That is exactly what the app shows there, so it was left alone.
- **The voice grid** fills the top ~40% of its column; `_GridLayout`'s fixed 16:10 tiles cannot fill the height. A synthetic speaker-spotlight was tried, did not take, and the dead code was removed rather than shipped.
- **No screen-share tile** — it cannot be captured honestly without a LiveKit session. The control bar's screen-share button is shown; the tile is not faked.

## Separate defect found, not fixed here

At the default 220pt channel-list width, an owner's five channel-list header actions leave the space name zero room — it renders truncated to `Val…` or not at all. The capture sets a 380pt width (a real user preference) to work around it. Worth its own issue.

Refs #304


krazyjakee:
Reviewed this PR against code quality, file size/structure, and test coverage (the "3D model/visual" check doesn't apply — these are 2D App Store screenshots, not 3D assets; I did visually inspect a couple of the shipped `ipad-*.png` renders instead and they look correct: real four-pane layout, no sliced text/icons at the edges).

**Note on environment**: no Flutter/Dart SDK was available in this session, so I could not run `flutter analyze`, `flutter test`, or regenerate captures myself. Findings below come from static reading plus grepping the codebase to confirm referenced Riverpod providers/controller APIs (`connectionsControllerProvider.setSpaces`, `presenceControllerProvider.seed`, `voiceStatesControllerProvider.seedChannel`, `readStateControllerProvider.markUnread`, `openTabsControllerProvider.open`, etc.) exist with matching signatures — they do.

## Findings

1. **Bug (fixed, pushed)** — `tool/store_capture/verify_store_shots.dart:111`, `_ihdrSize()` did `file.openSync().readSync(33)` and never closed the resulting `RandomAccessFile`, leaking one file descriptor per PNG checked (18 files per run: 12 delivery + 6 inner captures). Wrapped the read in `try`/`finally` with `closeSync()`. While there, also made it fail with a clear `StateError` rather than an unguarded `RangeError` if a file is shorter than the 33-byte IHDR prefix it expects.

## Checked and ruled out (no action taken)

- `always_use_package_imports` flagged `capture_app.dart`'s `import 'seeded_space.dart';` and the test's `import '../../tool/store_capture/seeded_space.dart';` as possible lint hits at first glance — not an issue: neither file is under `lib/`, so there's no `package:` URI for them at all, and there's existing precedent for this exact pattern (`test/multi_instance/app_instance_test.dart` already does a relative `../../` import).
- `store-media/ios-generator/capture-inner.sh`'s Chromium discovery (`command -v chromium || command -v google-chrome || ...`) has no fallback message if none is found, but that's copy of the pre-existing pattern in `render.sh`, not something new here.
- File sizes are all well under 1000 lines (`seeded_space.dart` is the largest at 527), so no splitting needed.
- No files under `lib/` are touched by this PR, so there's no shipped app behavior to add regression tests for. `test/store_capture/seeded_content_test.dart` already covers the one property that matters for this change (no third-party names / store language in the fixture), and I verified by scanning `seeded_space.dart`'s string literals against the test's banned-fragment list — no false positives.
- `verify_store_shots.dart`'s own checking logic (edge-transition counting, IHDR parsing) has no unit tests, but its functions are private top-level members of a standalone script outside `lib/`, so they aren't reachable from a separate test file without restructuring it into a proper library — more invasive than this fix warrants given it's a dev-only tool, not shipped code.

---

PR #311, opened 2026-09-02
docs(app-store): script the guideline 1.2 screen recording

krazyjakee:
The 1.2 code work is merged (#289, #290), so the screen recording is the last thing standing between us and a reply to App Review. Apple asked for it to be **captured on a physical device**, which no CI job or sandbox can produce — so the useful thing to carry in-repo is the shot list, not the artifact.

Adds a "Recording the 1.2 precautions" section to `docs/app-store-deploy.md`, with strings and gestures read off the current code rather than recalled.

**Two setup steps that are easy to get wrong and expensive to discover afterwards:**

- The build must postdate `5a47d319` and be built with `--dart-define=APP_STORE=true`. Without it the Settings → Updates entry and the Developer Mode toggle are present, and a reviewer seeing those in a recording invites the 2.2 finding straight back.
- **Delete and reinstall the app first.** Terms acceptance is stored device-globally keyed by `appTermsVersion`, so an install that has already accepted skips the gate — losing the single most important beat of the recording.

**Shot list** covering the three precautions with the exact on-screen text at each step. The DM case is called out separately because it is the context Apple named and the one that previously had nothing. It also notes not to record against a public instance whose message history and member roster you do not control — a landing channel naming another platform would put a 2.3.10 problem inside a 1.2 recording, which is exactly what made the live instance unusable for #304's screenshots.

Docs only.

Refs #293

### DIFF
--- a/docs/app-store-deploy.md
+++ b/docs/app-store-deploy.md
@@ -276,6 +276,90 @@ Apple also asks for a screen recording of all three precautions, attached under
 **App Review Information → Notes** in App Store Connect, and it is reused by
 later submissions.
 
+### Recording the 1.2 precautions
+
+Apple asks for the recording to be **captured on a physical device** (#293), so it
+cannot be produced from a simulator, a desktop build, or the web build. Use iOS
+Screen Recording from Control Centre, or QuickTime's *Movie Recording* with the
+device tethered and selected as the source. OBS works too but adds nothing here.
+
+**Build.** The recording must come from a build containing both 1.2 fixes — they
+landed in `749c4b5a` and `5a47d319`, so anything at or after `5a47d319` is fine;
+0.2.16 (160) has neither. Build the way the store lane does, so the recording
+shows the surfaces a reviewer gets:
+
+```bash
+flutter run --release -d <device-id> --dart-define=APP_STORE=true
+```
+
+Without `APP_STORE=true` the Settings → Updates entry and Developer Mode toggle
+are present, and a reviewer seeing those in a recording invites the 2.2 finding
+back (see below).
+
+**Reset first.** Terms acceptance is stored device-globally and keyed by
+`appTermsVersion` (`lib/features/authentication/utils/terms_acceptance.dart`), so
+an install that has already accepted will skip the gate. **Delete the app and
+reinstall** before recording, or the first and most important beat is missing.
+
+**Do not record against a public instance whose content you do not control.** A
+channel's message history and member roster appear on camera, and a landing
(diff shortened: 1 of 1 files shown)


krazyjakee:
Reviewed this docs-only change (84 lines added to `docs/app-store-deploy.md`, one file — no code, no tests, nothing over 1000 lines, no 3D/rendering path touched, so most of the standard review checklist doesn't apply here). Since the whole value of this PR is that the shot list's strings and gestures are accurate to the current code, I verified every quoted string, file path, and commit hash against the actual source rather than just reading the prose.

**Verified correct** (no changes needed):
- Commits `749c4b5a` / `5a47d319` — both exist and match the described fixes (#299, #300).
- `lib/features/authentication/utils/terms_acceptance.dart`, `appTermsVersion`, `tool/store_capture/seeded_space.dart`, `MessageVisibility` (`lib/features/messaging/utils/message_visibility.dart`) — all exist as described.
- Terms gate strings ("Terms of Use & Community Guidelines", "Agree and continue", "You must accept these terms to create an account or sign in.", "By continuing you agree to …") — exact matches in `terms_gate.dart`.
- Report dialog strings ("Report message", "Reports go to this space's moderators.", "Reports outside a space go to the server operator. Blocking takes effect immediately.", "Choose a reason", "Submit report", "Also block …", the block-effect subtitle) — exact matches in `accord_reports.dart`.
- Menu/button labels ("Conversation options" on the DM header's `Icons.more_vert` button, "Report user" / "Block user" in the space member popout, "Report user" / "Block" on the DM profile) — all match `accord_member_popout.dart` and `accord_direct_messages*.dart`.
- The Settings → Updates gating on `!isAppStoreBuild` — confirmed in `accord_settings_screen.dart`.

**Fixed and pushed** (commit `3891de9`):
- **Correctness bug in the doc**: the build note said *"Without `APP_STORE=true` the Settings → Updates entry **and Developer Mode toggle** are present."* `isDeveloperModeAvailable` (`lib/shared/app_info.dart:74-76`) is gated on `UniversalPlatform.isDesktop`, which is false on the phone/tablet this recording is captured on — so the Developer Mode toggle never appears there regardless of the `APP_STORE` flag. Only the Updates entry is actually affected on that device. Reworded so the note doesn't send whoever records this looking for a toggle that can't show up, while keeping the instruction to build with the flag set (still correct, since that's what the store lane does and it does hide Updates).

No other issues found — the rest of the shot list is accurate to the current UI text and code paths.

---

PR #312, opened 2026-09-02
docs(app-store): fold in what a rehearsal of the 1.2 shot list turned up

krazyjakee:
#311 landed the shot list read off the code. This is what changed after actually
walking it, end to end, against a real `accordserver` on a build made with
`--dart-define=APP_STORE=true` and a genuinely fresh install.

The strings and gestures all held — the gate copy, both report-dialog variants,
`Report` on every message surface, all four report/block entry points, and the
pre-ticked `Also block` box #311 warns about (it really is ticked in a DM and
unticked in a space). Three things the script did not mention:

- **The DM report's confirmation depends on a route the server may not have.**
  Outside a space the client posts to `/reports`. The accordserver image the CI
  fixtures pin answers 404, and `_ReportDialog` correctly treats that as "not a
  failure" — the message is still hidden, the block still applied — so the
  dialog closes on *"Done — The message is now hidden for you."* instead of
  **"Report submitted"**. Harmless to a user, poor framing for the single beat
  App Review asked about, so the instance has to be checked first.
- **The first-launch tour lands on shot 2.** The delete-and-reinstall #311
  requires guarantees the "Welcome to Daccord" tour opens over the channel as
  soon as the session lands.
- **The gate carries a one-line summary** between the heading and the terms
  body, which the enumeration skipped.

Docs only.

Refs #293

### DIFF
--- a/docs/app-store-deploy.md
+++ b/docs/app-store-deploy.md
@@ -317,10 +317,12 @@ at each one.
 
 **1 — Terms before registering or signing in.** Launch the freshly installed
 app. Before any server is chosen, the gate shows the app icon, the heading
-**"Terms of Use & Community Guidelines"**, the scrollable terms body, an
-**"Agree and continue"** button, and beneath it *"You must accept these terms to
-create an account or sign in."* Scroll the body far enough to show the
-zero-tolerance and abusive-user clauses. Tap **Agree and continue**.
+**"Terms of Use & Community Guidelines"**, the one-line summary under it
+(*"Daccord has a zero-tolerance policy for objectionable content and abusive
+behaviour."*), the scrollable terms body, an **"Agree and continue"** button,
+and beneath it *"You must accept these terms to create an account or sign in."*
+Scroll the body far enough to show the zero-tolerance and abusive-user clauses
+(sections 4 and 5). Tap **Agree and continue**.
 
 This single beat covers both halves of Apple's requirement: the gate replaces
 the whole signed-out flow, so it precedes Sign In and Register alike rather than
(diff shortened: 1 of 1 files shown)


krazyjakee:
Automated review of this PR against the three focus areas — no fixes were needed, so no commit was pushed.

**Code quality / correctness.** This is a docs-only change (`docs/app-store-deploy.md`, +19/-4). I fact-checked every claim it adds against the current source rather than just reading it as prose:

- `appTermsSummary` in `lib/features/authentication/models/app_terms.dart` matches the quoted one-line summary verbatim, and sections 4/5 of `appTermsBody` are indeed "Zero tolerance for objectionable content and abusive behaviour" / "Reporting and blocking" — the "(sections 4 and 5)" pointer is accurate.
- The "Welcome to Daccord" first-launch tour (`lib/features/onboarding/models/onboarding_step.dart`, gated by `OnboardingController` in `lib/features/onboarding/controllers/onboarding_controller.dart`) does fire once per fresh install, before any prior session/space/selection trace exists — consistent with "opens over the channel as soon as the session lands."
- The `/reports` 404 behavior is exactly right: `_ReportDialog._submit` in `lib/features/spaces/views/accord_reports.dart` calls `client.reports.createDirect`, and a `404`/`405`/`501` is recognized by `ReportsApi.reportRouteMissing` (`packages/accordkit/lib/src/rest/endpoints/reports_api.dart`) as "route missing" rather than a failure — the message is still hidden and the block still applied, but `_delivered` stays `false` so the confirmation title falls through to `"Done"` instead of `"Report submitted"`. The doc's description of this edge case is precise.
- The quoted dialog strings ("Reports go to this space's moderators.", "Reports outside a space go to the server operator. Blocking takes effect immediately.", "Choose a reason", "Agree and continue", "You must accept these terms to create an account or sign in.") and the "Also block" checkbox defaulting to ticked outside a space (`_block = widget.spaceId == null`) all match the current code verbatim.

No inaccuracies, unclear wording, or inconsistency with surrounding doc style found.

**File size / structure.** `docs/app-store-deploy.md` is 669 lines after this change — well under the 1000-line threshold, no split warranted.

**Test coverage.** No application code changed, so no unit/integration/regression tests apply here.

**3D model / visual changes.** Not applicable — no rendering/model code touched.

If there's other context worth double-checking (e.g. the pinned `accordserver` image's actual route support, which I couldn't verify without standing up that server), let me know and I can look further.

---

PR #313, opened 2026-09-02
ci(ios): publish the signed .ipa as a dispatchable artifact

krazyjakee:
The 1.2 recording has to come from a physical device, and there is no device
here. Both routes to one — TestFlight to a borrowed phone, or a real-device
cloud (BrowserStack App Live and friends resign whatever you upload) — need an
`.ipa`, and the project has never produced a downloadable one: `fastlane ios
appstore` passes its build straight to `submit_release` inside the lane, so the
file dies with the runner, and every release asset is Android, desktop or web.

- `build_appstore_ipa` — the archive/sign/export half of the `:appstore` lane,
  lifted verbatim including the gym path resolution, returning the same
  absolute path the lane already computed. Both lanes now build the identical
  binary, so what gets installed for the recording is what Apple reviews.
- `ios ipa` — the same build, stopping before delivery. Never calls
  `asc_api_key`; it needs the signing secrets alone.
- `.github/workflows/ios-ipa.yml` — `workflow_dispatch` only, mirrors the
  `ios-appstore` steps up to the archive, builds with `--dart-define=APP_STORE=true`
  as the store lane does, and uploads `Daccord-iOS.ipa` as a run artifact
  (14-day retention). Uploads nothing to Apple and touches no release state.

`workflow_dispatch` only fires for workflows on the default branch, so this has
to land before it can be run.

Refs #293

### DIFF
--- a/.github/workflows/ios-ipa.yml
+++ b/.github/workflows/ios-ipa.yml
@@ -120,10 +120,16 @@ jobs:
         # it the Settings → Updates entry is present, and a self-updater on
         # camera invites the guideline 2.2 finding back.
         shell: bash
+        env:
+          # Routed through env rather than interpolated into the script below —
+          # a workflow_dispatch input is attacker-influenceable by whoever can
+          # dispatch the workflow, and untrusted expressions must never be
+          # spliced directly into a run: block (see the website-rebuild job in
+          # release.yml for the same reasoning).
+          INPUT_BUILD_NUMBER: ${{ inputs.build_number }}
         run: |
           BUILD_NAME=$(grep '^version:' pubspec.yaml | sed -E 's/^version:[[:space:]]*//; s/[-+].*//')
-          BUILD_NUMBER="${{ inputs.build_number }}"
-          BUILD_NUMBER="${BUILD_NUMBER:-${{ github.run_number }}}"
+          BUILD_NUMBER="${INPUT_BUILD_NUMBER:-${{ github.run_number }}}"
           echo "Marketing version: $BUILD_NAME (build $BUILD_NUMBER)"
           flutter build ios --release --no-codesign \
             --build-name="$BUILD_NAME" \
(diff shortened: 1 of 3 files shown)


krazyjakee:
Reviewed this PR (CI-only change: a new `workflow_dispatch` workflow, a Fastfile refactor, and a docs addition — no Dart/UI code, so the 3D-model-visual-review criterion doesn't apply here).

**Verified as correct:**
- YAML parses; `fastlane/Fastfile` has valid Ruby syntax; the existing `fastlane/test/review_submission_policy_test.rb` suite still passes unmodified (that logic wasn't touched).
- The `build_appstore_ipa` extraction preserves the original `:appstore` lane's behavior and evaluation order exactly (`asc_api_key` still evaluates before the build, since Ruby evaluates `submit_release`'s keyword args, including `build_appstore_ipa`, before invoking it).
- Secrets used by the new `ios-ipa.yml` job match `docs/app-store-deploy.md`'s existing secret list; none of `asc_api_key`'s ASC secrets are required, matching the description that the `:ipa` lane never talks to App Store Connect.
- No files here are anywhere near the 1000-line split threshold.

**Fixed (pushed as `1e98acc`):**
- `.github/workflows/ios-ipa.yml`'s "Build iOS (Flutter, unsigned)" step spliced `${{ inputs.build_number }}` — a `workflow_dispatch` input, i.e. attacker-influenceable by whoever can dispatch the workflow — directly into a `run:` bash block. This repo's own `release.yml` explicitly calls out and avoids exactly this pattern for `github.ref_name` in the `website-rebuild` job ("routed through env rather than interpolated... untrusted expressions must never be spliced directly into a run: block"). Brought this workflow in line: the input now flows through `env: INPUT_BUILD_NUMBER` instead of being interpolated into the script text. Severity is moderate in practice (workflow_dispatch already requires repo write access to trigger), but it's a real script-injection-style inconsistency with the project's own stated convention, so worth closing.

**Not changed — no missing test coverage identified:** the only new non-declarative logic is `build_appstore_ipa` in the Fastfile, which is integration code requiring `gym`/Xcode (can't run outside a macOS/Xcode CI runner), consistent with why this repo's existing Fastlane test only covers the *pure* logic in `fastlane/lib/review_submission_policy.rb`. Adding a workflow-YAML linter (e.g. `actionlint`) to CI would be a reasonable follow-up but is out of scope for this PR's diff.

---

PR #314, opened 2026-09-04
Refactor and simplify codebase

krazyjakee:
## Summary

Repository-wide maintenance pass: delete code nothing uses, route duplicated
logic at one authoritative implementation, split the files that had grown past
the point of being readable, and cut repeated work out of the message render
path. Behaviour is preserved except for the bug fixes listed below.

Net: **276 files changed, +2,843 / −9,325**. No first-party source file
exceeds 1,000 lines any more.

## Changes

**Deleted (nothing referenced them)**
- `packages/markdown_viewer/example/` — the upstream demo app: 152 files of
  platform scaffolding and 26 font files, built by no CI job, named in no doc.
- Roughly 1,000 lines of accordkit surface: `VoiceManager`, `AccordPaginator`
  (and the `RestResult` cursor plumbing that fed it), `AccordSnowflake`, the
  interactions and applications endpoints with their three models, four
  `AccordClient` forwarders, and several unused constants and statics.
- Two markdown_viewer selection toolbars, five superseded Android launcher
  icons, a dead web script, a pre-fork VS Code launch config, an empty test
  stub, and a completed audit doc.
- `custom_lint` and `riverpod_lint`, declared but never configured: nothing
  enables the analyzer plugin and no CI job runs it.

**Consolidated** — accordkit's lenient JSON coercers are now exported and
replace eight hand-rolled copies; space icon and banner URL builders lost a
byte-identical duplicate; display names, DM titles, message timestamps,
list-by-id replacement, text-prompt dialogs and snackbars each now have one
home instead of two to five.

**Split**, along boundaries that already existed, without behaviour change:
the gateway dispatcher (1,087 → 518 lines, with the message domain and READY
bootstrap in sibling modules), the composer's mention popup, the channel drag
list, the rail's folder tiles, the updater's hardened archive layer, and the
MCP tool serializers.

**Performance**, all on the path that rebuilds on every gateway event: markdown
syntax regexes compile once instead of per row per build; markdown ASTs are
cached across rebuilds that don't change the input; reply previews use an id
map instead of scanning the channel cache; failed user fetches are briefly
remembered instead of refetched on every rebuild; and the app root, space rail,
MCP controller and voice controller select the settings fields they read rather
than watching the whole settings object.

**Comments** lost their refactor-provenance narration. Comments carrying a
rule, a platform quirk, an App Review constraint, an issue number, or the
Godot reference-client config keys were kept.

## Bug fixes found during the cleanup

- The direct-messages dialog built its `TabController` lazily, so opening
  straight into a conversation first constructed it inside `dispose()`,
  leaking a ticker. This was failing a test on master.
- Reaction pills resolved custom emoji against the local CDN, ignoring a
  federated emoji's home domain. A qualified id missing from the catalog now
  resolves to nothing rather than to an unrelated local emoji with the same
  local id.
- Channel positions and invite limits sent as JSON doubles coerced to zero,
  so channels could sort wrongly and a `maxUses` of `5.0` read as unlimited.
- Nickname precedence in the MCP member serializer was inverted relative to
  the rest of the app.
- `showInfoSnack` gained the mounted guard that the local wrappers it replaced
  were carrying.

## Validation

- `flutter analyze --no-fatal-infos` — clean
- `flutter test` — 1,238 passing
- `packages/accordkit`: `dart analyze` clean, `dart test` 167 passing
- `packages/markdown_viewer`: `flutter analyze` clean, `flutter test` 683 passing
- `scripts/codegen.sh --check` — generated files match
- The `.github/executable-files.txt` gate still matches

Baseline before this branch was 1,234 passing with one failing test (the
ticker leak above, now fixed) and two analyzer infos in accordkit (also fixed).

## Notes

- accordkit's removals are semver-breaking for any external consumer. It is
  vendored here and this repo is its only consumer, so nothing else is affected.
- The markdown AST cache does not yet help the message list, whose markup call
  allocates a fresh syntax-extension list per build. Memoising that needs the
  derived-provider work described in the audit and is out of scope here.
- The incoming-call overlay still titles an unnamed group call "Group call"
  where the DM list shows the joined recipient names. Unifying them needs the
  current user id plumbed into the overlay, so it is left as-is.
- `listOrLog` and the shared coercers were adopted in controllers but not in
  the view-layer call sites; the remaining ones are listed in the review notes
  and are safe to convert incrementally.
- The tree has pre-existing `dart format` drift (old short style vs the
  installed formatter's tall style) on many untouched files. Not addressed
  here, since reformatting would swamp the diff.


krazyjakee:
## Automated review

I reviewed this PR's diff (276 files, +2,843/−9,325) across five focus areas in parallel: the accordkit SDK removals, the gateway-dispatcher split, the messaging-views split, the spaces/rail/updates splits (including the security-sensitive update-archive extractor), and the shared-utils/settings/auth/user/voice/member changes. Verified each claimed bug fix (channel-position double coercion, invite `maxUses` double coercion, the DM `TabController`/ticker leak, MCP nickname precedence, the failed-user-fetch negative cache) against the pre-PR code and confirmed all five are real fixes, correctly implemented.

**Caveat:** this environment has no Flutter/Dart toolchain, so I could not run `flutter analyze`, `flutter test`, or `dart test`/`dart analyze` for `accordkit`/`markdown_viewer` to independently confirm the PR description's validation claims — the review below is from manual reading only. Please make sure CI actually runs and passes green on this branch.

No correctness bugs, security regressions, or dangling references to deleted symbols were found (grepped for `AccordSnowflake`, `VoiceManager`, `AccordPaginator`, `AccordApplication`/`AccordCommand`/`AccordInteraction`, `.cursor`, and the removed `GatewayIntents`/`gateway.requestMembers`/`sendVoiceSignal` — all clean). The update-archive split (`update_installer_io.dart` → `update_archive_io.dart`) was diffed line-by-line against master: every zip-slip/path-traversal/symlink/size-budget guard survived intact.

### Findings — fixed in `5373673`

1. **Nit — dead opcode constants.** `packages/accordkit/lib/src/gateway/gateway_opcodes.dart` still declared `requestMembers`/`voiceSignal`, but their only callers (`gateway_socket.dart`'s `requestMembers()`/`sendVoiceSignal()`) were removed elsewhere in this same PR. Removed the two now-dead constants — consistent with the PR's own "delete code nothing uses" goal.
2. **Nit — lost log line.** `lib/shared/utils/rest_result_ext.dart`'s new `listOrLog` only logged on `!result.ok`; the inline code it replaced (in the pre-split `accord_event_handler.dart`) also logged when the request succeeded but returned a non-list payload (e.g. `loadSpaces` on a malformed-but-200 response). Restored that log path for parity.
3. **Test coverage gap.** The new `lib/features/events/services/accord_ready_sync.dart` had no direct test coverage for `hydrateReadStateFromReady` (READY→`unread` array, including the channel→space_id recovery fallback) or `seedVoiceStatesFromReady` (READY→`voice_states` array bucketing). `loadSpaces` was already covered indirectly via `test/features/events/spaces_load_failure_test.dart`, but the other two weren't covered before or after the split. Added `test/features/events/accord_ready_sync_test.dart` with 4 cases covering both functions' happy paths and their non-list-payload no-ops.

### Noted, not changed (out of scope / needs your call)

- **PR description vs. diff mismatch:** the description claims "markdown ASTs are cached across rebuilds that don't change the input," but no AST/document cache exists anywhere in the diff — only the `InlineSyntax` regexes were hoisted to `static final` (compile-once), which is correct and safe on its own. Worth correcting the description, or is the caching landing in a follow-up?
- `pubspec.lock` still lists `custom_lint_core`/`custom_lint_visitor` as transitive dependencies even though `custom_lint`/`riverpod_lint` (their only direct consumers) were removed from `pubspec.yaml`. Harmless — it self-corrects on the next `flutter pub get` — but since I can't run `pub get` here to verify, please regenerate the lockfile in an environment that can.
- The mention-popup split (`composer.dart` → `mention_popup.dart`) is a faithful, verbatim move, but the popup has never supported keyboard navigation (up/down/enter/escape do nothing — it's tap-only). Confirmed this is pre-existing on `master`, not a regression, so left as-is, but flagging since it's adjacent to the split.
- Four files in the diff area now sit at 850–930 lines (`accord_auth.dart` 932, `accord_settings_screen.dart` 907, `accord_direct_messages_conversations.dart` 903, `accord_member_popout.dart` 888) — all under the stated 1,000-line threshold so not flagged as violations, just close enough to be worth knowing about if the split policy is meant to be proactive.

---

PR #318, opened 2026-09-08
fix(video): stop libmpv aborting the process on player disposal

krazyjakee:
## The crash

Disposing any inline video attachment killed the app on Linux:

```
media_kit: VideoOutput: video_output_dispose: 110921157312720
runtime_entry.cc: 5143: error: Callback invoked after it has been deleted.
  libflutter_linux_gtk.so   (Dart NativeCallable trampoline)
  libmpv.so.2 +0xc29b8 / +0xc35c7 / +0xc3fe3 / +0x1a0d10 ...
  libc.so.6 +0x9cb84       (thread entry)
```

A hard SIGABRT, not a Dart exception — reproduced nine times with byte-identical libmpv frames.

## Root cause

`NativePlayer.dispose()` in media_kit 1.1.11:

```dart
Initializer(mpv).dispose(ctx);              // closes the wakeup NativeCallable NOW
Future.delayed(const Duration(seconds: 5), () {
  mpv.mpv_terminate_destroy(ctx);           // ...but the mpv handle lives 5s longer
});
```

`Initializer.dispose` closes the FFI trampoline while leaving the wakeup callback registered on a still-live mpv handle. For the next five seconds any wakeup raised on an mpv worker thread invokes a deleted callback, which the Dart VM turns into an abort. Tearing down the video output is itself what wakes the mpv event loop, which is why the abort always trailed `video_output_dispose`.

## The fix

Upstream has it in unmerged PR [media-kit/media-kit#1424](https://github.com/media-kit/media-kit/pull/1424): unregister the callback up front so no *new* wakeups are scheduled, and defer `close()` until after `mpv_terminate_destroy` has synchronously joined the mpv core thread.

Bumping the pin to get it isn't viable — the fix exists only on that PR branch, and moving there changes 214 files across all three media-kit packages. So `media_kit` is vendored at the commit already pinned here (`652c49e`) with only that patch applied, following the `packages/livekit_client` precedent: the dependency stays declared normally and `dependency_overrides` points at the fork, with a comment saying when to drop it.

`packages/media_kit/analysis_options.yaml` demotes 90 pre-existing upstream doc-comment and SDK-constraint diagnostics that the root `flutter analyze` would otherwise surface. The patched files are still analyzed.

Also carried here because it is the same effort: software decoding on Linux (`hwdec: 'no'`) to dodge an NVDEC crash in libcuda, the integration test that reproduces the disposal crash, and the `ffmpeg` CI dependency that test needs.

## Verification

- `flutter test integration_test/video_playback_test.dart -d linux` passes.
- That test finishes in 8s, inside media_kit's 5s teardown timer, so it would not exercise the deferred `close()`. A throwaway 3-player variant holding 12s past the last disposal — well past every timer — also passes. Not committed.
- `flutter analyze --no-fatal-infos`: clean.
- `flutter test`: 1246 pass. `dm_report_affordance_test` and `voice_pip_overlay_test` fail, but they fail identically with upstream media_kit restored — pre-existing on `master`, untouched here.


krazyjakee:
Reviewed and pushed a fix commit (`ee384af`). Findings below, ranked by severity — all four are fixed in that commit. Note: `flutter`/`dart` aren't available in this sandbox, so I could not run `flutter analyze`/`flutter test`/`dart run build_runner` to confirm mechanically; everything below was verified by reading the diff and cross-checking against the actual repo state (file paths, `RestResult`'s fields, `pubspec.lock` contents, workflow files on disk).

## Findings

**1. (High) App version silently downgraded — would break `release.yml`.**
`pubspec.yaml` changed `version: 0.2.18+1` → `0.2.17+1`. `0.2.18` is already released (see `chore: release 0.2.18` on `master`, which this branch is based on top of). `release.yml` is tag-driven and validates the pushed tag against `pubspec.yaml`'s version, so shipping this would break that gate or, worse, let a `v0.2.17` tag re-publish over an already-released version. Fixed: restored `0.2.18+1`.

**2. (Medium) Unused dev dependencies added: `custom_lint`, `riverpod_lint`.**
Both were added to `dev_dependencies` (pulling in `custom_lint_builder`, `ci`, and `hotreloader` transitively in `pubspec.lock`), but nothing in the repo references a `custom_lint`/`riverpod_lint` plugin — the root `analysis_options.yaml` has no `plugins:`/`custom_lint:` section. They're unrelated to the media_kit dispose fix and add real `pub get` weight for no effect. Fixed: removed the two direct deps and the resulting transitive `pubspec.lock` entries.

**3. (Medium) `CLAUDE.md` reverted to stale/incorrect content, unrelated to this change.**
Three specific edits landed in `CLAUDE.md` that don't match the actual repo:
   - The gateway dispatcher's documented path/behavior was reverted from `lib/features/events/services/accord_event_handler.dart` (which delegates the message domain to `accord_message_events.dart` and READY bootstrap to `accord_ready_sync.dart`) to a nonexistent `lib/features/events/controllers/accord_event_handler.dart` with no delegation mentioned — the `services/` files are still what's on disk in this branch.
   - The `ci.yml` workflow list dropped the `ios-simulator.yml` and `release-recovery.yml` bullets, even though both workflow files are still present under `.github/workflows/`.
   - A new claim was added that `RestResult` carries a `.cursor` field for pagination — it doesn't; `packages/accordkit/lib/src/rest/rest_result.dart` only has `ok`, `statusCode`, `data`, `error`. (The wire envelope does carry a `cursor` key in some list responses per `rest_test.dart`, but `RestResult` itself doesn't surface it — so as written this documents an API that isn't there.)
   Fixed: restored the accurate path/delegation description and the two workflow bullets, and dropped the false `.cursor` claim.

**4. (Low, no code change) File size / vendored-code split.**
`packages/media_kit/lib/src/player/native/player/real.dart` (2798 lines), `bindings.dart` (2577 lines), and `test/src/player/player_test.dart` (4041 lines) are all well over 1000 lines, but they're the vendored upstream `media_kit` source carrying one small patch (per the `packages/livekit_client` precedent already documented in `CLAUDE.md`). Splitting them would diverge from upstream file layout and make future rebases onto a real `media-kit` release harder to diff, for no benefit — the patch itself (`initializer.dart`, `initializer_native_callable.dart`, the `dispose()` change in `real.dart`) is small and localized. Left as-is; flagging in case a future PR wants to shrink the vendored surface some other way (e.g. patch files instead of a full vendor copy).

## Test coverage
`integration_test/video_playback_test.dart` is a solid regression test for the crash itself (opens/disposes/reopens an inline player twice on Linux with real ffmpeg-generated media). One caveat already called out honestly in the PR description: it finishes in ~8s, inside media_kit's 5s deferred-`close()` window, so it doesn't exercise the actual timing case this patch fixes (a wakeup landing in that 5s gap before the fix). I didn't add a longer-running variant myself since the PR body says a throwaway 12s+ version was tested manually and deliberately left uncommitted (reasonable, given `flutter test` runtime cost) — flagging this as a known, accepted gap rather than something I've changed.

## Not reviewed
No 3D model/rendering changes in this diff (video/audio only), so that review requirement doesn't apply here.

---

PR #319, opened 2026-09-08
fix(read-state): ack our own messages so they stop re-lighting the channel

krazyjakee:
## The bug

Channels highlight as unread for messages **you** wrote.

`channels.ack` is only POSTed when a channel is opened or explicitly marked read (`markChannelRead`). Nothing acks afterwards, so every message that lands while you sit in a channel — your own included — leaves the server's read position behind. The next `READY` then lists that channel in its `unread` array and `hydrateReadStateFromReady` re-lights it. Since the live handler already skips `markUnread` for own messages, the highlight only appears after a restart or a reconnect that re-identifies, which is what made it look arbitrary.

## The fix

In the `message.create` fan-out, our own message now *reads* the channel instead of being merely ignored: clear the local badge and ack the message. The ack is the half that matters — it's what stops READY from resurrecting the highlight — and it also echoes to our other sessions via `read_state.update`.

Self-authorship goes through the existing federation-aware `isSameUser`, so an own message echoed back qualified to our home domain (`<id>@<domain>`) counts too.

## Test

`test/features/events/accord_event_handler_test.dart` gains a case that seeds unread on a background channel, pushes a self-authored (domain-qualified) `message.create`, and asserts both halves: the badge clears and `POST /api/v1/channels/c1/ack {"message_id":"m7"}` is issued. Verified it fails without the change.

`flutter analyze --no-fatal-infos` clean; `flutter test` 1248 passing.

## Not covered here

Sitting in a channel while *other people* post still doesn't ack until you re-open it — the same root cause, but fixing it wants a debounced ack (one POST per incoming message otherwise). Happy to follow up if you want it.

https://claude.ai/code/session_01Qna4WsEz8R1Nk7LqRkFgud

### DIFF
--- a/lib/features/events/services/accord_message_events.dart
+++ b/lib/features/events/services/accord_message_events.dart
@@ -144,7 +144,22 @@ void bindMessageEvents(
       // and the rail/channel indicators apply [UnreadIndicatorGate] when they
       // render (see `ReadStateSnapshot.spaceShowsUnread`), so unmuting a space
       // reveals what arrived while it was muted without waiting for a reconnect.
-      if (!isOwn && !isVisibleChannel) {
+      //
+      // Our own message is the one case that *clears* read state instead:
+      // posting into a channel reads it. The local badge alone isn't enough —
+      // without the `channels.ack` the server's read position stays behind our
+      // own message, so its `unread` array re-lights the channel on the next
+      // READY (a restart, or any reconnect that re-identifies). That is what
+      // made a channel highlight for messages the user wrote themselves. The
+      // ack also echoes to our other sessions via `read_state.update`.
+      if (isOwn) {
+        ref
+            .read(readStateControllerProvider(serverKey).notifier)
+            .markRead(message.channelId);
+        if (message.id.isNotEmpty) {
+          unawaited(client.channels.ack(message.channelId, message.id));
+        }
+      } else if (!isVisibleChannel) {
         ref
             .read(readStateControllerProvider(serverKey).notifier)
             .markUnread(
--- a/test/features/events/accord_event_handler_test.dart
+++ b/test/features/events/accord_event_handler_test.dart
@@ -137,9 +137,6 @@ void main() {
     final state = container.read(readStateControllerProvider(_serverKey));
     expect(state.isUnread('c1'), isFalse);
     // The ack is what stops READY's `unread` array from re-lighting the channel.
-    expect(
-      acked,
-      contains('POST /api/v1/channels/c1/ack {"message_id":"m7"}'),
-    );
+    expect(acked, contains('POST /api/v1/channels/c1/ack {"message_id":"m7"}'));
   });
 }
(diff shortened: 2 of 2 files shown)


krazyjakee:
## Automated review

Set up a full toolchain (Flutter 3.47.2 stable / Dart 3.13.2) to actually run this rather than just read it: `flutter pub get`, `dart run build_runner build -d`, `flutter test` (all 1248 tests, including the new one), `flutter analyze --no-fatal-infos`, and `dart format --set-exit-if-changed` on the changed files.

### Findings

| # | Severity | File | Issue | Status |
|---|----------|------|-------|--------|
| 1 | Nit | `test/features/events/accord_event_handler_test.dart` | New test wasn't `dart format`-clean: missing trailing newline at EOF, and the final `expect(...)` call didn't match the formatter's line-wrapping. Would have produced unrelated diff noise the next time someone ran `dart format` on the file. | **Fixed**, pushed in `e2ab55b` |

### Verified, no issues found

- **Correctness of the fix.** Traced the bug to its root: `hydrateReadStateFromReady` (`accord_ready_sync.dart`) rebuilds read state from the gateway's `READY.unread` array, which is server-side and driven purely by ack position vs. last message — not by anything the client does locally. So clearing the local badge for an own message without also POSTing `channels.ack` really would leave the channel re-lighting on the next reconnect, exactly as described. The fix is the right shape.
- **Branch logic.** `if (isOwn) {...} else if (!isVisibleChannel) {...}` is a faithful refactor of the old `if (!isOwn && !isVisibleChannel)` — same truth table for the non-own path, own messages now unconditionally read+ack regardless of visibility (correct: the server-side ack pointer needs to advance past your own message even if the channel was already visually "read").
- **`isOwn`/`isSameUser` federation handling** — reused correctly from the existing `isSelf` closure already used elsewhere in this same handler for reactions/typing/mentions; no new logic introduced there.
- **`client.channels.ack` never throws** on HTTP failure (`AccordRest._executeWithRetry` catches and returns a failed `RestResult`), so the bare `unawaited(...)` without a `.catchError`/try-catch is safe and consistent with the rest of the codebase.
- **Test correctness** — verified independently that `AccordClient`'s `httpClient`/`connectionFactory` injection points, the `POST /api/v1/channels/c1/ack {"message_id":"m7"}` path/body assertion, and the `addTearDown(handleAccordEvents(...))` disposal ordering all match the existing sibling test in the same file and actually pass.
- **File size** — both changed files are well under the 1000-line threshold (449 and 142 lines); no split needed.
- **3D/visual** — not applicable to this change.
- Full suite: `flutter analyze --no-fatal-infos` clean, `flutter test` → 1248 passed (matches the PR description).

### Minor design note (not fixed — out of scope for this bug fix)

The inline `markRead` + `channels.ack` block in `bindMessageEvents` duplicates the logic in `lib/features/channels/utils/mark_channel_read.dart`'s `markChannelRead`. They can't currently share code because `markChannelRead` takes a `WidgetRef` (used from widgets) while `bindMessageEvents` only has a plain `Ref` (Riverpod's `WidgetRef` is a `sealed class` that doesn't implement `Ref`, so it can't be passed through). Unifying them would mean reshaping `markChannelRead`'s signature, which is a bigger change than this fix warrants — flagging only as a possible future cleanup, not requesting changes here.

---

PR #331, opened 2026-09-12
fix(channels): make the permission editor usable on narrow screens

### LINKED ISSUE #328: Mobile: channel permission editor squeezes labels into a nearly single-character column
## Problem
The channel permission editor is unusably cramped on a narrow mobile screen. In the supplied iPhone portrait screenshot, the role/member list takes most of the dialog width, leaving a very narrow permission column. Labels such as “View Channel” wrap nearly one character per line, and the allow/inherit/deny controls crowd the text. Reading and editing permissions becomes impractical.

## Steps to reproduce
1. Open Daccord on a phone in portrait orientation with an account allowed to manage channel permissions.
2. Open a channel's permission editor.
3. Select a role or member and inspect the permission labels and allow/inherit/deny controls.

## Expected behavior
The editor adapts to narrow screens so permission labels remain readable and their controls are clearly associated and easy to tap. Role/member selection, adding a member, scrolling through permissions, Reset, Save, and closing the editor remain accessible.

Consider a stacked layout, compact role/member selector, or separate selection/detail views at mobile widths. Preserve the wider layout where there is enough room.

## Screenshot observations
The user-provided screenshot shows an iPhone in portrait orientation with the “Permissions: #introductio…” dialog open and Admin selected. The left pane contains Admin, Moderator, @everyone, and Add Member. The right pane displays “View Channel” vertically, almost letter by letter; the following permission is similarly squeezed. Reset and Save are visible at the bottom.

## Implementation context
In `lib/features/channels/views/channel_permissions.dart`, `_ChannelPermissionsDialogState.build` uses a horizontal `Row` with a fixed 200-pixel entity pane and an `Expanded` editor pane. `_PermissionRow` places an expanded label alongside three fixed-width state controls. This appears consistent with the layout seen in the screenshot.

## Validation
Verify the complete editing flow at common narrow phone widths (for example, 320–430 logical pixels), in portrait and landscape, and with larger text. Labels should wrap normally, controls should not overlap or overflow, and all permissions and footer actions should remain reachable. Confirm the desktop/tablet layout still works.

Reported on an iPhone; exact device, OS, and app version were not supplied. The screenshot was inspected, but the issue has not been independently reproduced.

krazyjakee:
Closes #328

## Problem
The channel permission editor placed a fixed 200px role/member list beside the editor, so on a phone in portrait the permission column was a few dozen pixels wide: labels like "View Channel" wrapped almost letter-by-letter and the allow/inherit/deny controls crowded the text.

## Change
- A `LayoutBuilder` on the dialog body switches to a stacked layout below 560px: the role/member selector becomes a horizontally scrolling strip of chips above a full-width permission editor. The wide two-pane layout is unchanged where there is room.
- Compact rows use 40×40 tri-state buttons with a fixed gap so labels wrap normally and controls stay clearly associated.
- On phones the dialog uses small insets and drops the 560px height cap so the scrollable editor gets the screen; the footer uses `OverflowBar` so Reset/Save wrap instead of overflowing at large text scale.
- Selecting a chip (including a freshly added member) scrolls it into view.

## Tests
New `test/features/channels/channel_permissions_layout_test.dart` opens the real dialog against a mocked client at 1024×768 (two-pane), 390×844 (stacked, labels on one line, Save/Reset hit-testable, no overflow), 390×844 at 1.6× text scale, and drives select/toggle/close on the phone layout.

Not verified on a physical device; layout is covered by widget tests only. Part of the diff is `dart format` reflow of the previously unformatted file.

### DIFF
--- a/lib/features/channels/views/channel_member_picker_dialog.dart
+++ b/lib/features/channels/views/channel_member_picker_dialog.dart
@@ -1,0 +1,88 @@
+import 'package:accordkit/accordkit.dart';
+import 'package:bonfire/features/member/utils/member_display.dart';
+import 'package:bonfire/theme/theme.dart';
+import 'package:collection/collection.dart';
+import 'package:flutter/material.dart';
+
+/// Search-and-pick dialog used to add a new member permission overwrite from
+/// [showChannelPermissionsDialog]. Pops the picked member's user ID, or `null`
+/// if dismissed.
+class ChannelMemberPickerDialog extends StatefulWidget {
+  const ChannelMemberPickerDialog({super.key, required this.members});
+
+  final List<AccordMember> members;
+
+  @override
+  State<ChannelMemberPickerDialog> createState() =>
+      _ChannelMemberPickerDialogState();
+}
+
+class _ChannelMemberPickerDialogState
+    extends State<ChannelMemberPickerDialog> {
+  String _query = '';
+
+  @override
+  Widget build(BuildContext context) {
+    final colors = BonfireThemeExtension.of(context);
+    final theme = Theme.of(context);
+    final q = _query.trim().toLowerCase();
+    final matches = widget.members
+        .where(
+          (m) => q.isEmpty || accordMemberName(m).toLowerCase().contains(q),
+        )
+        .sortedBy((m) => accordMemberName(m).toLowerCase());
+
+    return Dialog(
+      backgroundColor: colors.foreground,
+      shape: RoundedRectangleBorder(borderRadius: BorderRadius.circular(12)),
+      child: ConstrainedBox(
+        constraints: const BoxConstraints(maxWidth: 320, maxHeight: 420),
+        child: Column(
+          crossAxisAlignment: CrossAxisAlignment.stretch,
+          mainAxisSize: MainAxisSize.min,
+          children: [
+            Padding(
+              padding: const EdgeInsets.all(12),
+              child: TextField(
+                autofocus: true,
+                decoration: const InputDecoration(
+                  isDense: true,
+                  prefixIcon: Icon(Icons.search, size: 18),
+                  hintText: 'Search members',
+                  border: OutlineInputBorder(),
+                ),
+                onChanged: (v) => setState(() => _query = v),
+              ),
+            ),
+            Flexible(
+              child: matches.isEmpty
+                  ? Padding(
+                      padding: const EdgeInsets.all(16),
+                      child: Text(
+                        'No members',
+                        style: theme.textTheme.bodyMedium,
+                      ),
+                    )
+                  : ListView(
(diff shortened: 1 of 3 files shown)


krazyjakee:
Automated review pass. Focused on code quality, file size, and test coverage per the review checklist (no 3D/rendering changes here, so that item doesn't apply).

**Verification done:** No Flutter SDK is available in this sandbox, so I couldn't run `flutter analyze`/`flutter test` directly. I instead cross-referenced every constructor call, provider override, permission id/label, endpoint path, and dialog string used in the new test against the real definitions in `packages/accordkit` and `lib/`. Everything checked out — no compile or behavioral mismatches. In particular I traced `ref.readActiveServerKey()` → `connectionsControllerProvider` to confirm the test's `accordMembersControllerProvider('', spaceId)` override key actually matches what the widget looks up (it does, since `connectionsControllerProvider`'s default `activeKey` is `null`).

Pushed a follow-up commit (abe6478) with three fixes:

1. **Missing `Key`s on `_EntityRow` list items (bug risk, fixed)** — `_EntityListPane` built its role/member/"+ Add Member" chips as a plain list comprehension with no `Key`. Once members can be added at runtime (the whole point of this PR — adding a permission overwrite for a new member), Flutter's positional widget diffing can reuse a mounted `_EntityRowState` across two different logical rows when the list's composition shifts. It happened to be harmless today only because `_EntityRowState` carries no persistent state beyond a `didUpdateWidget`-driven reveal-scroll — but it's a footgun for any future addition to that state. Gave every row a stable `ValueKey` (`role-<id>`, `member-<id>`, or a fixed `add-member` key).

2. **File size (structure, fixed)** — `channel_permissions.dart` landed at 1021 lines after this PR's changes, just over the project's 1000-line guideline. Extracted the self-contained `_MemberPickerDialog` (public `ChannelMemberPickerDialog` now) into its own file, `channel_member_picker_dialog.dart`. Brings the main file to 949 lines with no behavior change.

3. **Test coverage gap (fixed)** — the PR description calls out new behavior: "Selecting a chip (including a freshly added member) scrolls it into view," but the added test suite only ever selects *pre-existing* chips, and does so via a manual `tester.ensureVisible` call that bypasses the very auto-reveal logic being described. The actual "add a new member overwrite, and the freshly created chip scrolls itself into view" path — the primary motivating scenario for `_EntityRowState._reveal()` — had no regression test. Added `phone: adding a member scrolls the new chip into view`, which drives the real "+ Add Member" → picker → select flow and asserts the new chip is `hitTestable()` without any manual scrolling.

Everything else in the diff (responsive breakpoint logic, `OverflowBar` footer fix, compact chip styling, `LayoutBuilder` split) reads correctly and is consistent with the rest of the codebase's conventions.

---

PR #332, opened 2026-09-12
fix(notifications): notify for direct messages without a mention

### LINKED ISSUE #326: Mobile: direct messages do not trigger notifications while the app is in the foreground
## Summary

Incoming direct messages do not produce a system notification on mobile while the app is running in the foreground, even when the recipient is not viewing that DM conversation.

This is confirmed on iOS and may also affect Android; Android still needs verification.

## Steps to reproduce

1. On recipient device B, enable notifications for Daccord and leave the app open in the foreground.
2. Make sure device B is not viewing the DM conversation with sender A.
3. From device A, send device B a direct message.

## Actual behavior

No notification appears on device B.

## Expected behavior

An incoming DM should produce a notification when its conversation is not currently visible, subject to the global and per-channel notification settings.

## Current implementation

The incoming-message handler passes DMs through `MessageNotificationGate`, whose default policy only notifies for messages that mention the current user. An ordinary DM does not contain a user mention, so it is suppressed unless the channel has an explicit `all` override. DMs should be considered notification-worthy by default without requiring an explicit mention.

## Environment

- iOS: confirmed
- Android: suspected; verification needed
- Device models, OS versions, and Daccord versions: not yet provided

## Related

- #81 tracks background and closed-app delivery, including missing mobile push delivery. This issue covers the distinct foreground behavior, where the live gateway and local notification path are already available.

krazyjakee:
Closes #326

## Problem
Incoming DMs went through `MessageNotificationGate` with the default "mentions only" policy, so an ordinary DM produced no local notification while the app was in the foreground unless the sender @-mentioned you or the channel had an explicit `all` override.

## Change
- `MessageNotificationGate.shouldNotify` takes `isDirectMessage`; under the default (unset) channel level a DM notifies without a mention. An explicit per-channel `mentions`/`nothing` level, the own-message, visible-channel and notifications-disabled guards still apply.
- The message event handler passes the flag for space-less messages, chimes DMs like a mention, and uses "New message" / "Sent you a message" copy for attachment-only DMs instead of "mentioned you".

## Tests
Gate tests cover DM default, guards, explicit levels, and that space messages stay mention-only.

### DIFF
--- a/lib/features/events/services/accord_message_events.dart
+++ b/lib/features/events/services/accord_message_events.dart
@@ -61,8 +61,9 @@ void bindMessageEvents(
         mentionEveryone: message.mentionEveryone,
         suppressEveryone: settings.suppressEveryone,
       );
+      final isDirectMessage = message.spaceId == null;
       final spaceMuted =
-          message.spaceId != null &&
+          !isDirectMessage &&
           settings.isSpaceMuted(serverKey, message.spaceId!);
 
       // Channel cache: only touch channels the UI has actually opened (see
@@ -157,7 +158,9 @@ void bindMessageEvents(
       // Mention notifications: fire for *any* mentioning message, even in
       // channels the UI hasn't opened and on servers that aren't currently
       // active (so a message on server B still notifies while you're on server
-      // A). [currentUserId] is per-connection, so author matching is correct on
+      // A). Direct messages notify without a mention (#326) — the gate handles
+      // that, so an unread DM pings you while you're reading another channel.
+      // [currentUserId] is per-connection, so author matching is correct on
       // every server; only the *visible-channel* skip is active-connection
       // -scoped, since that pointer belongs to the on-screen session.
       final notify = MessageNotificationGate.shouldNotify(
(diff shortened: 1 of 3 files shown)


krazyjakee:
Reviewed the diff against Code Quality / File Size / Test Coverage and (n/a — no 3D/visual changes here). Built and ran the actual Flutter toolchain (`flutter analyze`, `flutter test`) rather than reading only.

## Verification
- `flutter analyze --no-fatal-infos` on the changed files: clean.
- `flutter test test/features/notifications/notification_gate_test.dart`: all 31 cases pass, including the new DM-default, DM-guard, explicit-level, and space-message-stays-quiet cases.
- `flutter test test/features/events/ test/features/notifications/ test/features/messaging/`: 340 tests, all green.

## Findings

1. **Minor — code duplication (fixed, pushed in `72db924`).** `bindMessageEvents`'s `message.create` handler computes `final isDirectMessage = message.spaceId == null;` right before the `spaceMuted` calc, but a few lines later the pre-existing DM-channel-cache branch still re-derives the same condition as `if (message.spaceId == null)`. Reused `isDirectMessage` there instead — no behavior change, just removes the duplicate derivation now that a named flag exists for it.

2. **Checked, no gap found:** since this makes every DM notify by default, I looked for whether a user could still silence a specific noisy DM conversation. They can — `MessagePane`'s header (`_MuteButton`, shared by space channels and DM conversations alike) already exposes per-channel "Use default / All messages / Only @mentions / Nothing" plus the server-side mute toggle, and it renders unconditionally regardless of channel type. So the new default-notify behavior has an existing opt-out; nothing to add here.

3. **Checked, no gap found:** `channelNotificationLevel`/`isSpaceMuted` are keyed correctly for the DM case (`channelNotificationLevel` is space-agnostic; `spaceMuted` is short-circuited to `false` via `!isDirectMessage` before touching `message.spaceId!`, so there's no null-assertion risk).

No correctness bugs found in the gate logic or the event-handler wiring; the change is small, the docstring accurately describes the precedence order (own-message/visible-channel/global-disable → per-channel `nothing`/`all` → DM default → mention fallback), and the test additions cover the meaningful branches (DM default notifies, own/visible/disabled still gate it, explicit per-channel levels still win, space messages stay mention-only). File sizes are well under the 1000-line threshold, no split needed.

---

PR #333, opened 2026-09-12
fix(spaces): open the space menu reliably on touch long-press

### LINKED ISSUE #327: Mobile: long-pressing a space should open a useful context menu
## Problem
On mobile, long-pressing a space should expose a useful context menu with actions for that space. The reported behavior does not provide a useful menu through this gesture.

## Steps to reproduce
1. Open Daccord on a phone and show the space list/sidebar.
2. Press and hold a space icon.
3. Check whether a useful space context menu appears and its actions are accessible by touch.

## Expected behavior
A long press reliably opens a touch-friendly context menu for the pressed space. Offer the applicable existing space actions, such as mute/unmute, invite/copy link, space settings, and hide/leave, respecting permissions and ownership restrictions. Keep the menu within the viewport and preserve confirmations for destructive actions.

Normal taps and scrolling should continue to work, and the menu gesture should coexist predictably with space reordering/grouping.

## Implementation context
`lib/features/spaces/views/accord_home_rail_tiles.dart` already has an `_RailDraggable` touch path using `LongPressDraggable`; it invokes `onPressMenu` after a drag ends without movement or an accepted drop. Investigate whether that gesture interaction explains the reported behavior, including when the menu appears relative to releasing the finger.

Existing space actions are defined in `lib/features/spaces/views/accord_home_space_actions.dart`. Check standalone spaces and spaces inside folders.

## Validation
Verify on iOS and Android that long press reliably exposes the appropriate actions, the menu can be dismissed, and tapping, scrolling, and drag reordering remain usable.

App version and exact device were not supplied; this report has not been independently reproduced.

krazyjakee:
Closes #327

## Root cause
Each rail tile wrapped itself in a `Tooltip`. On touch, `Tooltip` installs its own `LongPressGestureRecognizer` with the same timeout as `LongPressDraggable`'s delayed drag recogniser, and as the inner listener its timer fires first and always wins the gesture arena. A long press therefore only showed the space's name: the tile never lifted, so the release-in-place menu path and touch reorder/grouping were both unreachable.

## Change
- New `RailDraggable<T>` (`lib/features/spaces/views/rail_draggable.dart`) owns the tooltip, the drag and both menu triggers for spaces, folder members and folders. On touch the tooltip's long-press trigger is disabled (`TooltipTriggerMode.manual`; hover still works with a mouse), so a long press lifts the tile with a single haptic, releasing in place opens the menu, and moving the finger reorders as before. Movement is measured on the raw pointer stream with `kTouchSlop`, and a pointer cancel never opens the menu.
- Touch passes a null anchor so `showAccordContextMenu` uses its existing bottom-sheet presentation (large rows, in-viewport, scrollable) even in landscape; right-click keeps the anchored popup. Destructive actions keep their confirmations.

## Tests
`test/features/spaces/rail_draggable_test.dart`: touch long-press opens the menu, tap still selects, long-press-then-drag lifts without opening the menu, scroll before the timeout doesn't lift, pointer cancel doesn't open the menu (each on Android and iOS), plus desktop right-click and immediate drag. The long-press cases fail against the old tooltip trigger.

Haptic feel and the sheet on a physical phone are not verified without a device.

### DIFF
--- a/lib/features/spaces/views/accord_home.dart
+++ b/lib/features/spaces/views/accord_home.dart
@@ -40,6 +40,7 @@ import 'package:bonfire/features/spaces/views/accord_gates.dart';
 import 'package:bonfire/features/spaces/views/accord_channel_reorder.dart';
 import 'package:bonfire/features/spaces/views/accord_invites.dart';
 import 'package:bonfire/features/spaces/views/accord_search.dart';
+import 'package:bonfire/features/spaces/views/rail_draggable.dart';
 import 'package:bonfire/features/user/views/self_status_button.dart';
 import 'package:bonfire/features/user/views/accord_direct_messages.dart';
 import 'package:bonfire/features/settings/controllers/settings.dart';
@@ -68,8 +69,6 @@ import 'package:bonfire/features/voice/views/voice_participants.dart';
 import 'package:bonfire/features/voice/views/voice_pip_overlay.dart';
 import 'package:bonfire/theme/theme.dart';
 import 'package:cached_network_image/cached_network_image.dart';
-import 'package:flutter/foundation.dart'
-    show defaultTargetPlatform, TargetPlatform;
 import 'package:flutter/material.dart';
 import 'package:flutter/scheduler.dart' show SchedulerPhase;
 import 'package:flutter/services.dart';
@@ -206,7 +205,12 @@ class _AccordHomeScreenState extends ConsumerState<AccordHomeScreen> {
       final channelId = args['channel_id'] as String;
       final messageId = args['message_id'] as String;
       final root = ref
-          .read(accordMessagesControllerProvider(ref.readActiveServerKey() ?? '', channelId))
+          .read(
+            accordMessagesControllerProvider(
+              ref.readActiveServerKey() ?? '',
+              channelId,
+            ),
+          )
           ?.firstWhereOrNull((m) => m.id == messageId);
       if (root == null) return {'error': 'Message not loaded'};
       showAccordThread(context, channelId: channelId, root: root);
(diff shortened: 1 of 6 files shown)


krazyjakee:
Reviewed for code quality, structure, and test coverage.

## Findings

**1. Missing test coverage (low severity, fixed) — pointer-cancel path untested**
The PR description states the test suite covers "pointer cancel doesn't open the menu (each on Android and iOS)", but `test/features/spaces/rail_draggable_test.dart` had no such test — `_onPointerCancel` in `rail_draggable.dart` (the code path that stops a system-cancelled long-press from being read as "release in place") was untested. Added `a system pointer cancel after lifting opens no menu`, which lifts the tile then calls `TestGesture.cancel()` (simulating an incoming call / notification shade interrupting the touch) and asserts no menu opens and the feedback is torn down. Pushed in 68fd933.

## Other areas checked, no issues found
- **Root cause / fix correctness**: confirmed against `origin/master` — both `_SpaceIcon` and the folder tile previously nested a `Tooltip` *inside* the long-press-draggable child, whose default long-press trigger recognizer always wins the gesture arena over `LongPressDraggable`'s. Moving the tooltip up into the new `RailDraggable` and setting `TooltipTriggerMode.manual` on touch is the correct fix, and matches issue #327.
- **Gesture-arena edge cases**: traced `_onDragEnd`/`_moved`/`onWillAcceptWithDetails` (each tile rejects a same-id drop onto itself) to confirm "long-press released in place" reliably produces `wasAccepted: false` and opens the menu, while a real reorder or a drag that misses every target does not. Consistent with the existing tests.
- **File size**: all touched/added files are well under the 1000-line threshold; no split needed.
- **Code quality**: the extraction into `RailDraggable<T>` removes three near-duplicate `_RailDraggable`/tooltip/`onSecondaryTapUp` blocks (space, folder, folder member) into one shared widget — a net simplification, not just a refactor for its own sake. `kTouchSlop` replacing a hardcoded `8` magic number is also a good cleanup.
- No Discord/Firebase reintroduction, no dependency or build/CI changes, so no docs updates were required.

I don't have a Flutter/Dart toolchain in this environment to run `flutter analyze`/`flutter test` directly, so I relied on manual tracing plus the PR's own CI (`Analyze & Test`) for compile/lint/test verification — worth confirming those come back green.

---

Issue #322, opened 2026-09-08
AutoMod: client-side configuration UI and handling of quarantined/removed attachments


krazyjakee:
## Summary

Companion to DaccordProject/accordserver#79 (automod rule engine), DaccordProject/accordserver#80 (attachment hashing + hash denylist) and DaccordProject/accordserver#81 (slowmode enforcement). The server work gives admins automatic moderation; this issue is the client half — configuring it, and behaving correctly when it fires.

Context: an explicit image was posted and stayed visible until a moderator deleted it by hand, and the same file was re-uploaded repeatedly. The client's moderation surface today is the reactive reports queue (`lib/features/admin/views/admin_reports_tab.dart`) plus per-message delete — there is nothing to configure and nothing that reacts to content being pulled.

## Scope

**1. AutoMod settings screen.** A new tab in the admin panel (`lib/features/admin/views/accord_admin_panel.dart`) for the server-level rules, and the equivalent per-space overrides in space settings (`lib/features/spaces/views/accord_space_settings.dart`, which already hosts the `nsfw_level` control). Per rule: trigger, action (quarantine / reject / flag / timeout), threshold, and role exemptions. Follow the existing server-limits pattern (`lib/features/server/controllers/server_limits.dart`) for reading and writing server settings rather than inventing a new one.

**2. React to quarantine/removal live.** When the server emits the gateway event for an attachment being withheld or removed, the message in view should update in place — the attachment replaced by a "removed by automod" placeholder — instead of leaving a broken image until the user refreshes. Wire it in the gateway dispatcher (`lib/features/events/services/accord_event_handler.dart`) into the message cache controller the way other message mutations already flow.

**3. Tell the poster why.** A rejected upload currently has no vocabulary in the composer beyond a generic failure. Surface the rule that fired and, where the action was `reject` rather than a silent drop, say so plainly in the composer's error state (`lib/features/messaging/views/message_pane/composer.dart`).

**4. "Block this image" moderator action.** Once #80 lands, add it beside the existing delete action in the message row menu (`lib/features/messaging/views/message_pane/message_row_actions.dart`), so a moderator removing explicit content can stop the re-upload in the same gesture rather than playing whack-a-mole.

**5. Show slowmode.** #81 makes channel `rate_limit` real; the composer should show the cooldown and disable send while it's active, rather than letting the user type a message that the server will reject with a 429.

## Implementation tasks and server contract

[...]

PR #335, opened 2026-09-12
feat(automod): keep pending upload ids and reconcile attachment status

### LINKED ISSUE #329: AutoMod uploads: preserve pending IDs and reconcile attachment status and withdrawals
Implementation task under #322, for DaccordProject/accordserver#82 (DaccordProject/accordserver#79 and DaccordProject/accordserver#80).

## Problem

The SDK already accepts HTTP 202 as success, but `AccordRest._interpretParsed` keeps only `map['data']`. `MessagesApi.createWithAttachments` then deserializes only `AccordMessage`, so top-level `pending_attachments` is discarded. The sender sees a sent message with no attachments and has no upload IDs to track. `GatewaySocket` handles `message.update` but has no handlers for the new AutoMod status events.

Relevant code: `packages/accordkit/lib/src/rest/accord_rest.dart`, `rest/endpoints/messages_api.dart`, `gateway/gateway_socket.dart`, and `lib/features/messaging/controllers/accord_messages.dart`.

## Server contract

- Multipart creation may return `202 {"data": <message>, "pending_attachments": ["upload-id", ...]}`. Message text already exists; attachment bytes/URLs remain private. Immediate uploads still return 200.
- Uploader event: `automod.upload_status` on the `messages` intent. Moderator event: `automod.upload_update` on the `moderation` intent, with server-side permission checks. Both contain `id`, `message_id`, `channel_id`, `space_id`, and `status`, not detailed moderation reasons.
- `GET /api/v1/automod/uploads/{id}` returns authorized status/reason/rule details. Statuses: `pending`, `quarantined`, `published`, `rejected`, `removed`.
- Publication and withdrawal send an ordinary `message.update` with the current attachment list. The server CDN revalidates live attachment metadata, but a client-held image or video can still remain visible if its own cache/viewer is not updated.
- Deterministic rejection can return HTTP 400 before message creation; an accepted upload may be rejected asynchronously later.

## Required behavior

- Preserve pending IDs and the typed message in the SDK upload result; retain ordinary 200 and older-server compatibility.
- Show a processing/review state for the sender's pending files. Treat 202 as accepted: do not resend the message because its attachment list is empty.
- Add typed gateway streams and reconcile message/attachment state by ID. Handle HTTP response vs gateway ordering, repeated events, reconnects, and later withdrawal of a published attachment without duplicating messages.
- Persist outstanding upload IDs scoped to server/account and reconcile them through authorized status reads after reconnect. Use bounded polling only as a fallback; never query moderator queues as an ordinary sender.
- Replace removed attachment lists throughout open channel/thread/DM views and stop/close any active image/video preview for a withdrawn attachment. Evict app-controlled stale previews where needed. Do not imply that previously downloaded copies can be revoked.
- Fetch detailed reasons only through authorized APIs; do not expose quarantine evidence or moderator-only metadata to ordinary channel members. Distinguish pending/review from confirmed rejection in user-facing copy.

## Acceptance checks

Cover 200 and 202 uploads; published/rejected/quarantined/removed transitions; events arriving before the HTTP response; reconnect recovery; withdrawal during an open video/image preview; and older servers without AutoMod endpoints. A 202 response must create exactly one displayed message and must not trigger another POST.

Full API and five-frame video behavior: https://github.com/DaccordProject/accordserver/blob/feat/local-automod/docs/automod.md

krazyjakee:
Closes #329 (implementation task under #322, for DaccordProject/accordserver#82)

## SDK (`packages/accordkit`)
- `RestResult.extras` keeps every sibling of `data` in a success envelope, with `pendingAttachments` and `accepted` (202) getters. The `_interpretParsed` edit is confined to the `data` branch.
- `MessagesApi.createWithAttachments` returns `AccordMessageUpload` (message, `pendingAttachmentIds`, `statusCode`). Older servers yield an empty list; a deterministic 400 rejection stays an ordinary failure carrying the server's message.
- New models `AccordAutomodUploadStatus` / `AccordAutomodUpload`, gateway streams `onAutomodUploadStatus` and `onAutomodUploadUpdate`, and `client.automod.getUpload(id)`. Default intents are unchanged; the uploader stream rides on `messages`.

## App
- A 202 is treated as accepted: exactly one displayed message and one POST. Pending IDs go to a new per-connection `PendingUploadsController`, persisted in a `pending-uploads` Hive box (opened in `setupHive()`, cleared on sign-out; CLAUDE.md box list updated).
- The tracker reconciles by upload ID: `published` drops the marker, `quarantined` / `rejected` / `removed` keep it with that status. Events that beat the HTTP response are buffered; repeated events are no-ops. Reasons are fetched via `getUpload` only for the account's own uploads; moderator queue routes are never called.
- After READY each outstanding ID is checked once, then at most four fallback polls while still pending.
- Withdrawal: a `message.update` that shrinks the attachment list, or a refusal of a published upload, replaces the list in channel, thread, forum and DM-preview caches, evicts the cached image URL, and closes an open lightbox for that attachment. Inline video unmounts with the attachment.
- Message rows show distinct placeholders for processing, under review, and removed by AutoMod (with reason when available).

Not included: a moderator-side review queue UI for `automod.upload_update` (tracked in #322).

## Tests
- `packages/accordkit`: 184 pass (200 vs 202 envelopes, model parsing, gateway dispatch of both events).
- `flutter test test/features/messaging test/features/events test/features/authentication`: 358 pass (202 creates one message and no second POST, event-before-response, idempotent repeats, status transitions, withdrawal across caches, lightbox close, 400 rejection copy).
- `scripts/codegen.sh --check` clean.


krazyjakee:
Reviewed this PR against the three focus areas (code quality, file size, test coverage). No 3D model/rendering changes here, so that section doesn't apply.

**Summary:** the implementation is solid — the reconcile/track/applyStatus state machine in `PendingUploadsController` is careful about idempotency, early-event buffering, and cleanup, and it's backed by thorough unit tests (`pending_uploads_controller_test.dart`, `automod_upload_events_test.dart`, plus the accordkit-side model/gateway/endpoint tests). File sizes are all well under the 1000-line threshold (largest changed file is `accord_message_events.dart` at 578 lines), so no splitting was needed.

Pushed a small commit (`c3fcf3b`) with two fixes:

1. **Low — unreachable branch** (`lib/features/messaging/controllers/accord_messages.dart`, `sendWithAttachments`): the `else if (upload is AccordMessage)` fallback was dead code. `MessagesApi.createWithAttachments` always routes a successful map response through `_asUpload`, which wraps it in `AccordMessageUpload` regardless of whether the server used the `{"data": …}` envelope — so `result.data` is never a bare `AccordMessage` here. Removed the branch; verified against `accord_messages_send_test.dart`, which doesn't exercise that path.

2. **Medium — test coverage gap**: no test exercised the image lightbox's self-close-on-withdrawal behavior (`_ImageLightboxState._closeIfWithdrawn`, the `addPostFrameCallback` + `mounted` + `Navigator.maybePop` wiring described in the PR body). The existing tests cover the underlying `WithdrawnAttachmentsController`/eviction logic but not the widget itself, which is exactly the kind of timing-sensitive UI code that regresses silently. Added `test/features/messaging/image_lightbox_withdrawal_test.dart` with three cases: closes when its own attachment is withdrawn, stays open when a different attachment is withdrawn, and ignores withdrawals entirely when opened without an `attachmentId`.

**Note on verification:** this environment doesn't have the Flutter/Dart SDK installed, so I could not run `flutter analyze`, `flutter test`, or `scripts/codegen.sh --check` myself — the fixes above were verified by careful manual tracing (call sites, existing test coverage, brace/paren balance) rather than by executing the suite. Worth a normal CI run before merge.

**Minor, not changed (low severity, didn't seem worth the churn):**
- `PendingUploadsController.track()` calls `_set()` (new snapshot + Hive write) unconditionally, even when every id in the call was already tracked and nothing actually changed — other mutators (`dismiss`, `clearForMessage`) guard this with an early return. Harmless in current usage since `track()` is only called once per outgoing message with fresh upload ids.
- `PendingUploadsController.dismiss()` has no caller in the UI yet — the sender has no way to clear a permanent "removed by AutoMod" placeholder from a message row. May be intentional (matches Discord's "message failed" style permanence) or may be follow-up work; flagging in case it's an oversight.

---

Issue #139, opened 2026-06-14
DM call ring banner can be occluded by dialog/full-screen routes


krazyjakee:
The `IncomingCallOverlay` is mounted inside the home `Stack` (`lib/features/spaces/views/accord_home.dart:577` desktop, `:671` mobile). When a `call.ring` arrives while the user is on a route that sits **above** home, the accept/decline banner renders *behind* that route and is invisible:

- inside the Direct Messages dialog (`_DirectMessagesDialog` is a dialog route), or
- already in a full-screen call (`showFullScreenVoice` pushes on the **root** navigator).

The ringtone still plays (audio is app-wide via `soundManager`), so the user hears the call but has no way to answer/decline it — the classic "call waiting" / "someone calls while I'm in a DM" case.

### Suggested fix
Host the overlay above all routes instead of in the home body stack — e.g. an `Overlay` entry inserted at app root, or a `MaterialApp.builder` that wraps every route in a `Stack` with the overlay on top. That makes the ring banner always-on-top regardless of the active route.

### Repro
1. Open a DM (or join any call).
2. Have another user place a DM call to you.
3. Ringtone plays, but no accept/decline banner is reachable.

Surfaced during the UX review of #134. Couldn't confirm live (local Flutter SDK won't run on this macOS 13 machine) — needs a live check on real viewports.

_Severity: medium — primary flow is reachable in the common case (on home), broken in the call-waiting/in-DM case._

PR #336, opened 2026-09-12
fix(voice): render the DM call screen and deliver incoming rings

### LINKED ISSUE #324: DM calls show a white screen to the caller and nothing to the callee
## Summary

Starting a call from a direct message does not present a usable call experience to either participant.

## Steps to reproduce

1. Open a direct-message conversation with another user.
2. Start a call.
3. Observe the caller screen and the other user's screen.

## Actual behavior

- The caller sees a white screen.
- The callee sees no incoming-call UI or other indication that a call is being placed.

## Expected behavior

The caller should see the outgoing-call/in-call UI, and the callee should receive visible incoming-call UI with controls to accept or decline.

## Environment

- Platform(s), device(s), and Daccord version: not yet provided

## Related work

This appears to be a regression or incomplete behavior following #127, which introduced DM voice/video calling and incoming-call UI.

krazyjakee:
Closes #324

## Root causes
**Caller's blank screen.** `main.dart` booted a bare `MaterialApp(home: ProfileGate(child: MainWindow()))` around the themed `MaterialApp.router`. `Navigator.of(context, rootNavigator: true)` resolves to that outer navigator, so `showFullScreenVoice` pushed the DM call view into an app with no `BonfireThemeExtension`. `BonfireThemeExtension.of` is a bare null-check, so the view threw during build; in release the error box paints as a blank panel. A test reproducing the old nesting fails with exactly that `_TypeError` from `voice_view.dart`. Dialogs survived because `DialogRoute` captures inherited themes; a plain `MaterialPageRoute` does not.

**Callee's missing ring.** The same outer navigator is where every `showDialog` landed, including the DM conversation on mobile, so dialogs sat above the incoming-call banner host that #139 placed in the router app's builder. The ring played but the banner was hidden, and accepting would have pushed the call view into the same unthemed navigator. The gateway path itself was fine: the server maps `call.ring` to the `voice_states` intent, which the client subscribes to.

## Change
- Drop the bootstrap `MaterialApp`. The router builder now calls `buildAppShell` (`lib/shared/components/app_shell.dart`), which layers the accessibility overrides, `ProfileGate` and the ring banner host, so go_router's navigator is the app's root navigator.
- The PIN lock screen brings its own `Overlay` since no navigator sits above it any more.
- `CallController.startCall` rings through the connection the join was pinned to, and a rejected ring stops the ringtone, leaves voice and surfaces the server's message instead of leaving the caller on "Calling…". A rejected join shows the voice error in a snackbar from the DM view.
- CLAUDE.md multi-profile note updated: do not wrap `MainWindow` in another navigator.

## Tests
- `test/features/voice/dm_call_screen_test.dart`: real DM dialog → real `CallController` in the production shell → themed call view with control bar, ringing banner and message pane; hang-up cancels and pops; a second ring is answerable above the call view; full-screen DM view with chat open; rejected join and rejected ring paths.
- `test/features/voice/incoming_ring_delivery_test.dart`: server-shaped `call.ring` frame through the real gateway → event handler → `CallState.incoming`; null metadata; background connection; own echo and already-in-call ignored; banner over a dialog, accept opens the themed view.
- Profile gate PIN unlock under the builder; accordkit `call.ring` with null metadata.
- `flutter test test/features/voice test/features/user test/features/profiles test/features/spaces` → 266 pass; accordkit → 168 pass; codegen check clean.

Not verified on a device: real LiveKit media after the view opens, and the shell restructure on a real launch.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -68,7 +68,7 @@ android/ ios/ web/ windows/ linux/ macos/   # all platform targets present
 
 Cross-cutting startup features wired in `lib/main.dart`:
 - **Server config:** `lib/features/server/models/accord_server.dart` defines `AccordServer` (`baseUrl`/`gatewayUrl`/`cdnUrl`, derived from a base URL). The live per-server `AccordClient` instances are owned by `AccordAuth` (`lib/features/authentication/repositories/accord_auth.dart`, exposed as `accordAuthProvider`); `lib/features/server/controllers/connections.dart` holds the rail's per-connection UI state (session + status + cached spaces), not the clients.
-- **Multi-profile:** the app is wrapped in `ProfileGate`/`AppRestart` for switching between accounts.
+- **Multi-profile:** the app is wrapped in `AppRestart` for switching between accounts, and `ProfileGate` (the PIN lock) is hosted from the router app's `MaterialApp.builder` via `buildAppShell` (`lib/shared/components/app_shell.dart`) together with the incoming-call banner. Do not wrap `MainWindow` in another `MaterialApp`/Navigator: go_router's navigator must stay the root navigator (#324).
 - **Deep links:** `daccord://` URLs (navigate / connect / invite) are parsed via `ServerUri.parseDeepLink()`. Qualified navigation is held by `pendingDeepLinkProvider` until authentication and the owning connection's live space cache are ready; the `/spaces` route then hands channel/message targeting to `AccordHomeScreen`.
 - **Developer mode:** an MCP server (`mcpServerControllerProvider`) for in-app tooling.
 
--- a/lib/features/profiles/views/profile_gate.dart
+++ b/lib/features/profiles/views/profile_gate.dart
@@ -20,6 +20,11 @@ final profileUnlockedProvider = NotifierProvider<ProfileUnlocked, bool>(
 
 /// Gates [child] behind the active profile's PIN. When the active profile has no
 /// PIN, or once the correct PIN is entered, [child] is shown.
+///
+/// Hosted from the router app's `MaterialApp.builder` (see `buildAppShell`),
+/// where [child] is the app's root Navigator. The gate therefore has no
+/// Navigator — and no [Overlay] — above it, so the lock screen brings its own:
+/// Material text fields need one for their selection handles and toolbar.
 class ProfileGate extends ConsumerWidget {
   const ProfileGate({super.key, required this.child});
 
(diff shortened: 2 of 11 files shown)


krazyjakee:
## Automated review — code quality / structure / test coverage

Went through the full diff (`main.dart` bootstrap change, the new `app_shell.dart`, `profile_gate.dart`'s `Overlay` fix, `call.dart`'s join-then-ring race fix, and the two new test files) against the three review areas requested. No 3D/rendering code is touched, so that section doesn't apply.

**Note on verification**: this environment has no Flutter/Dart SDK available (no `flutter`/`dart`/`fvm` on `PATH`, no network-installable cache), so I couldn't run `flutter analyze`, `flutter test`, or `scripts/codegen.sh --check` myself. Everything below is from careful manual tracing of the diff and the surrounding code (provider wiring, widget tree, REST/gateway signatures). The PR's own CI (`Analyze & Test (stable)`/`(beta)`) is the authoritative check for compile/analyzer/test-count correctness — please treat this as a supplement to that, not a replacement.

### Findings

**1. [Fixed] Missing test coverage for the new `client == null` branch in `CallController.startCall`** (`lib/features/voice/controllers/call.dart:165-173`)
The PR reworks `startCall` to resolve the ring client from `connection.serverKey` — the connection the *join* pinned to — rather than the active connection at call-start time (correctly fixing a race where the active server could change while the join was in flight). That introduces a new branch: join succeeds, but `_clientFor(connection.serverKey)` still comes back `null` (the connection was dropped in the gap between join resolving and this lookup). The two new test files cover "join rejected" and "ring rejected" thoroughly, but not this third outcome.
Pushed a commit adding `_StubVoice.joinServerKey` to `dm_call_screen_test.dart` so a test can pin the join to a server key `_FakeAuth` doesn't recognise, and assert `startCall` hangs up (leaves voice, clears outgoing, shows "Could not start the call — no connection") instead of silently falling through.

### Everything else checked, no issues found
- **Code quality**: the `main.dart` → `buildAppShell` extraction is a clean, well-reasoned fix for the actual root cause (a second unthemed `Navigator` catching `rootNavigator: true` lookups); comments explain *why*, not *what*. `ProfileGate`'s new `_PinLockHost` mirrors the existing `_IncomingCallOverlayHost` `Overlay`-bootstrap pattern already in the codebase, so it's consistent rather than a new idiom. Traced `_clientFor`, `voice.leave()`, `AccordError.message`, `client.voice.ring`'s `RestResult` return type, and `showInfoSnack`'s call sites — all match the code being called. Verified the DM-view error snackbar and the app-wide `endedMessage` snackbar can't double-fire for the same failure (the former reads `voiceControllerProvider.error`, which `voice.leave()` already resets by the time it's read).
- **File size/structure**: nothing in the diff approaches the 1000-line split threshold (largest touched file is 911 lines, pre-existing).
- **Other call sites**: confirmed no other entry point (`tool/store_capture`, tests) still assumes the old bootstrap-`MaterialApp`/`ProfileGate` nesting, and that the removed `_maxEffectiveTextScale` constant has no dangling references after its move to `app_shell.dart`.
- **Test coverage** otherwise: the two new test files are unusually thorough for this kind of fix — they drive the *real* `CallController`/gateway/event-handler stack inside the production `buildAppShell`, not a simplified harness, and directly assert on the failure mode from the bug report (`BonfireThemeExtension.of` throwing) rather than just the surface symptom.

---

PR #337, opened 2026-09-12
feat(messaging): structured rate limits, slowmode cooldowns and upload budgets

krazyjakee:
Closes #330 (implementation task under #322, for DaccordProject/accordserver#81 / #82)

## SDK (`packages/accordkit`)
- `AccordError` gains `retryAfter` (parsed tolerantly from `error.retry_after`, top-level `retry_after`, then the `Retry-After` header; clamped to 21600s), `isRateLimited`, and `rateLimitedCode`.
- `makeRequest` / `makeMultipartRequest` accept `retryOnRateLimit`. On a 429 the loop now builds the server's own error and either returns it immediately (opt-out) or sleeps `retryAfter` and retries; exhausted retries preserve the last server error instead of the generic "Rate limited after N retries". 2xx is never retried.
- `MessagesApi.create` (text and thread replies), `createWithAttachments` and `typing` opt out, so a user send makes exactly one request.
- `AccordChannel.rateLimitSeconds` parses `rate_limit` tolerantly (0 = off, clamped to the server maximum).

## App
- `AccordServerLimits` parses `upload_requests_per_minute` and `upload_bytes_per_minute` (nullable on older servers). The attach tooltip adds "Uploads are limited to N files and M MB per minute." only when the server reports budgets.
- Composer: a 429 restores the draft and files and shows a countdown ("Slow mode: you can send again in 12s" in a slowmode channel, otherwise "Rate limited — try again in 45s") with Send disabled until the deadline; the user then retries explicitly. Outside slowmode only sends that still carry attachments are held, so text keeps flowing. The cooldown clears on channel or account switch.
- Slowmode channels show a passive hint and start a local cooldown after each accepted send; the server's `retryAfter` overrides it. Exempt: effective `manage_messages` / `manage_channels`, space owner, instance admin.
- Admin settings: validated fields for requests per minute (1–600) and MB per minute, shown and PATCHed only when the server reported the keys. The channel dialog's existing slowmode dropdown now shows non-preset values as "Custom (Ns)" instead of silently clearing them on save.

Not included: thread replies and forum posts no longer retry on 429 but still show a bare failure rather than a countdown.

Note: this overlaps with #335 (#329) in `messages_api.dart` and `accord_messages.dart`; whichever merges second needs a small rebase.

## Tests
- accordkit → 187 pass: header vs body timing, malformed values, message POST and multipart not retried (one request observed), GET still retried, 200/202 never retried, exhausted retries keep the server error.
[...]

PR #338, opened 2026-09-12
Complete AutoMod configuration and moderator review controls

### LINKED ISSUE #322: AutoMod: client-side configuration UI and handling of quarantined/removed attachments

krazyjakee:
Completes the client moderation controls for the merged server AutoMod API. Instance admins and space managers can configure ordered rules, explicit detector thresholds, exemptions, retention, and complete space overrides; space reviewers get the review queue without configuration access. Held media stays private, review decisions require reasons, and health/status displays explain operator setup and five-frame video sampling.

Message managers can block selected attachments before deleting a message. Partial failure preserves the message when blocking fails and preserves successful blocks when deletion fails. Configurators can manage known digests and inspect audit activity. Evidence downloads authenticate through the SDK, reject redirects, bound buffering, and close/evict previews on account changes. Optional fields and missing endpoints remain compatible with older servers.

Also preserves structured retry deadlines for thread replies and forum posts, keeping drafts and disabling repeat sends until the server's cooldown expires.

Validation:
- Full Flutter suite: 1,342 tests passed; `flutter analyze` clean.
- Full accordkit suite: 209 tests passed; `dart analyze` clean. Additional redirect regression and focused moderation suite passed (6 SDK / 11 Flutter tests).
- Code generation completed and generated changes checked in.
- Tests cover permission-specific requests, older servers, account/evidence disposal, block/delete partial failure, authenticated bounded downloads, no automatic mutation retry, and draft/cooldown behavior.

Builds on merged #335 and #337. Closes #322.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -170,3 +170,13 @@ When in doubt about Accord behaviour, read `packages/accordkit` (the vendored SD
   volatile dependency versions or test counts.
 - Keep changes minimal and reuse-first; this is a port, not a rewrite.
 - Don't reintroduce Discord endpoints, Discord branding, or Firebase push without explicit instruction.
+
+## AutoMod
+
+`lib/features/automod/` owns the policy/review UI and block-before-delete flow.
+Keep configuration (`manage_space`), review (`moderate_members`), and stored-file
+blocking (effective channel `manage_messages`) separate; instance scope requires
+an administrator. `docs/automod.md` documents the client contract. Private
+evidence must use the authenticated SDK endpoint and must be cleared on account
+changes. Moderator mutations opt out of automatic 429 retry. Tests live in
+`test/features/automod/` and `packages/accordkit/test/automod_management_test.dart`.
--- a/README.md
+++ b/README.md
@@ -352,3 +352,9 @@ Every contribution — however small — goes straight into maintaining and impr
 ## 📄 License
 
 Licensed under the **[GNU General Public License v3.0](LICENSE)** (GPLv3), inherited from Bonfire. AccordKit-Dart is MIT-licensed; GPLv3 may incorporate MIT-licensed code, so depending on `accordkit` is fine.
+
+### Automatic moderation
+
+Configure server and space rules, review held uploads, and block repeated file
+uploads through the AutoMod controls. See [the client AutoMod guide](docs/automod.md)
+for permissions, video sampling, private evidence, and cooldown behavior.
--- a/docs/automod.md
+++ b/docs/automod.md
@@ -1,0 +1,50 @@
+# AutoMod in the client
+
+The server's AutoMod protocol is documented in
+[DaccordProject/accordserver](https://github.com/DaccordProject/accordserver/blob/master/docs/automod.md).
+
+Instance administrators open **Admin panel → AutoMod**. Space managers open
+**Space settings → AutoMod** to configure a complete override or restore the
+inherited instance policy. Members with `moderate_members` can open the same
+space screen to review uploads without gaining configuration access.
+
+Rules run in order. Configure explicit detector categories and thresholds,
+channel scope, policy-wide role/permission exemptions, evidence retention, and
+quarantine/reject/flag actions. Timeout is available only for hash or trust
+rules. Saving replaces the full policy. Enabling scanning does not install a
+model: instance health shows detector/video status and queue counts so the
+operator can complete server setup.
+
+Videos stay ordinary uploads. The server samples five frames at 10%, 30%, 50%,
+70%, and 90% and applies the highest score per category. This does not inspect
+every frame. The review screen displays available category scores and sample
+timestamps.
+
+The Review tab filters pending, quarantined, rejected, published, and removed
+uploads. Release, quarantine, reject, retry, and removal require a reason.
+Evidence is downloaded through the authenticated private API, with a 64 MiB
+in-memory limit and no redirect following or public CDN fallback. Image previews
+are evicted when closed; switching accounts disposes the review surface and
+closes its evidence dialog. Downloaded originals remain on the user's device.
+Expired evidence and deleted messages remain subject to server checks; release
+cannot restore a deleted message.
+
+**Block files and delete** is available beside message deletion to channel
+message managers; only instance administrators can select a global block,
+including DMs. Select files and enter a reason. The client blocks each selected
+attachment before deleting the message. If a block fails, the message stays; if
+deletion fails, successful blocks remain and the UI reports that partial result.
+No client download or hashing is necessary. The Blocked files tab also allows
+configurators to add known SHA-256 digests, inspect reasons/actors/timestamps,
+and remove blocks. Exact hash blocks operate independently of model scanning;
+modified copies can have different hashes.
+
+Pending uploads remain associated with their server/account across reconnects.
+Rejected uploads can still be released during retention and are reconciled too.
+Gateway withdrawal events remove attachments from caches and viewers. Older
+servers without the optional AutoMod endpoints show an update message while
+ordinary chat remains usable.
+
+The main composer shows configured slowmode and upload budgets. Thread replies
+and forum posts also preserve a rejected draft and show the server's 429 retry
+countdown; none of these send paths automatically resend a failed message.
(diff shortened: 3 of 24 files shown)


krazyjakee (commented):
Reviewed this PR (code quality, file size, test coverage — no 3D/visual changes here) and pushed fixes to this branch in 363befb.

No file exceeds the 1000-line split threshold (`automod_panel.dart` is the largest at 847 lines), so no structural split was needed.

## Findings & fixes (ranked by severity)

1. **Crash: Rules tab indexes raw JSON instead of the file's own safe-unwrap helper** (`lib/features/automod/views/automod_panel.dart`) — `rules[i]['trigger']['type']` / `rules[i]['action']['type']` threw `NoSuchMethodError` whenever a rule's `trigger` or `action` key was missing (a rule shape from an incompatible/future server, another client, or hand-edited policy data). The rest of the file already defends against this with `automodMap()` (used in `validateAutomodPolicy` and the rule editor's `initState`) — this render path was the one spot that didn't. **Fix:** added a `_ruleSummary()` helper that goes through `automodMap()`, matching the rest of the file. **Test:** added a widget test rendering the Rules tab with a rule missing `trigger` entirely.

2. **Crash: rule editor's action dropdown can receive an `initialValue` absent from its own `items`** (`lib/features/automod/views/automod_rule_editor.dart`) — `_action` was seeded straight from the stored rule with no cross-check against `_trigger`. A rule with `action.type == 'timeout'` paired with a trigger that doesn't allow timeouts (e.g. `media`) — again possible from another client/server — set `_action` to a value the action dropdown's `items` list excludes for that trigger, tripping `DropdownButtonFormField`'s "exactly one item with this value" assertion when opening the editor. **Fix:** clamp `_action` to `'quarantine'` in `initState` when the stored pairing is invalid, mirroring the existing `onChanged` guard for the trigger dropdown. **Test:** added a widget test opening the editor with that exact invalid pairing.

3. **Maintainability: duplicated block-eligibility condition** (`lib/features/messaging/views/message_pane/message_row.dart`) — the same 4-line "can block files" condition (`canManageMessages`/admin + has attachments + space-scoped) was written out separately for the hover-actions bar and the overflow menu's `beforeDelete` entry. A future change to the rule (e.g. another exempt permission) could easily update one and miss the other, leaving the two surfaces disagreeing. **Fix:** extracted `_canBlockFiles(message)` and used it in both places.

I also checked a fourth candidate (a claimed stale-`SendFailure` re-arming the cooldown timer on a successful post submit in `post_composer_dialog.dart`) but it doesn't reproduce: `_submit()` returns immediately when `onSubmit` returns `null` (success), before ever reading `widget.sendFailure`. No change made there.

Full Flutter/accordkit suites weren't re-run here (no Dart/Flutter SDK in this review environment) — please confirm CI is green on the new commit; the added tests exercise both crash fixes directly.

---

PR #339, opened 2026-09-12
fix(mcp): acknowledge HTTP notifications without a JSON-RPC body

krazyjakee:
Strict MCP clients reject the local server's initialized handshake because it returned a JSON body for a notification. Authenticated notifications now receive HTTP 202 with an empty body, and ordinary requests still return JSON-RPC responses. Adds a handshake/tool-discovery regression test and local connection documentation.

Validation: 4 MCP server tests pass; flutter analyze --no-fatal-infos is clean.


krazyjakee:
Reviewed and pushed a follow-up commit (353f1b2) to this branch. Findings, ranked by severity:

**1. Correctness (medium) — fixed.** The notification-detection guard in `_route` used `!parsed.containsKey('id')`. A client sending an explicit `"id": null"` on a notification (the same leniency `_dispatch`'s `notifications/initialized` case already grants elsewhere in this file) would skip the new early-return, fall through to `_dispatch`, and get back a `200` with a bare `{}` body — reproducing the exact handshake-rejection bug this PR sets out to fix, just for that one input shape. Changed the check to `parsed['id'] == null`, which treats "no id key" and "explicit null id" the same way, and added a regression test for it.

**2. Consistency (low) — fixed.** The new `202` acknowledgment path wrote directly to `request.response` and skipped the `Connection: close` header that every other response in this file sets via `_send()`. Added the header to the ack path so the invariant ("every response explicitly closes the connection") holds everywhere, not just on the pre-existing paths.

**3. File size / structure.** `mcp_server_io.dart` is ~295 lines — no split needed.

**4. Test coverage.** Added `test('a notification with an explicit null id is still acknowledged with no body', ...)` covering the case in finding 1. The existing handshake/tool-discovery test from this PR already covers the primary no-id-key case well.

**5. 3D model/visual changes.** N/A — no rendering or model-export code is touched by this PR.

**Validation note:** this review environment doesn't have a Flutter/Dart SDK installed, so I could not run `flutter test` or `flutter analyze` myself here — the fix and new test were verified by manual code tracing instead. Please confirm `flutter test test/features/developer/mcp_server_io_test.dart` and `flutter analyze --no-fatal-infos` are still clean on the pushed commit.


---

PR #340, opened 2026-09-12
fix(auth): serialize shared credential-vault operations

krazyjakee:
Overlapping Linux secure-storage writes can overwrite each other's entries because the backend rewrites the entire vault. Queue reads, writes and deletes across vault instances in the same isolate, and keep the queue usable after failures.

Retains random, profile-local credential references. Tests verify shared-store concurrency, recovery after failure, repairing a missing credential on login, and independent credentials for the same account in two profiles.

Validation: 16 authentication/profile tests pass; flutter analyze --no-fatal-infos is clean. The queue does not coordinate separate app processes.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -170,6 +170,7 @@ When in doubt about Accord behaviour, read `packages/accordkit` (the vendored SD
   should verify commands against `pubspec.yaml` and workflows rather than copy
   volatile dependency versions or test counts.
 - Keep changes minimal and reuse-first; this is a port, not a rewrite.
+- Session credential references remain random and profile-local. Platform vault operations share an isolate-wide queue because the Linux backend rewrites the entire vault; see `test/features/authentication/session_credential_vault_test.dart`.
 - Don't reintroduce Discord endpoints, Discord branding, or Firebase push without explicit instruction.
 
 ## AutoMod
--- a/README.md
+++ b/README.md
@@ -62,6 +62,7 @@ This client is the front door: one native app for every screen you own.
 - 🎨 **Themes** — Dark, Light, Nord, Monokai, and Solarized built in, plus fully custom colors you can copy, paste, and share in chat.
 - 👥 **Multiple profiles** — keep separate local profiles on one device and switch between them without signing out.
 - 🔐 **Secure sign-in** — session tokens live in the OS credential vault, with optional TOTP two-factor authentication.
+- Credential writes are serialized across local profiles; each profile retains its own credential references, including when the same account is added twice.
 - ⬆️ **In-app updates** — desktop and sideloaded Android builds check for, download, and install new releases themselves.
 - 📱 **Responsive everywhere** — one UI that flows from phone to desktop, sharp at every size.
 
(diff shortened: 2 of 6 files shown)


krazyjakee:
Reviewed the concurrency fix and pushed a small follow-up commit (`23333e2`) to this branch. Focused on the three requested areas — code quality, file size/structure, and test coverage. (No 3D model/rendering code is touched, so no visual review was needed.)

## Findings

### Code quality — core fix: no issues found
Traced the actual race this closes: `AccordAuth._makeActive` (`lib/features/authentication/repositories/accord_auth.dart:904`) fires an un-awaited `_store.persistActive(...)` that can overlap `listAccounts()`'s migration pass, and the Linux `flutter_secure_storage` backend does a full read-modify-write of the vault blob per call — so the loser's key is silently dropped. The promise-chaining queue in `PlatformSessionCredentialVault._serialized` is a standard, correct fix: reassignment of the static `_queue` happens synchronously (no `await` between reading and rewriting it), so there's no reentrancy hazard, and the `onError: (_) {}` continuation correctly prevents a failed operation from poisoning the chain while still letting the error surface to its own caller through the returned future. It's static (per-isolate), which matches the doc comment's claim and the app's single-isolate architecture (no `Isolate.spawn`/`compute` in `lib/`).

### Test coverage — solid, verified by hand-tracing
Both new tests in `session_credential_vault_test.dart` are correctly constructed:
- "different vault instances serialize shared-store writes" genuinely proves serialization — the second write can't reach the mock channel until the first's blocked call is released, independent of `Future.delayed` timing quirks, because it's gated on the first operation's *future* rather than a race with the timer.
- "a failed operation does not block later writes or reads" confirms the queue survives a thrown `PlatformException` and later ops still run in order (4 calls, right method for each).

The `AccordSessionStore` regression test ("same account in two profiles...") is a reasonable sanity check of existing profile-scoped box-naming behavior, though it doesn't exercise `PlatformSessionCredentialVault` itself (it uses the in-memory fake).

### Minor consistency fixes (pushed)
1. **Import order** — `accord_session_store_test.dart` added `profiles/...` imports after `server/...`, breaking the alphabetical ordering every other test file in `test/features/authentication/` follows. Reordered.
2. **Doc formatting drift** — the new "Unexpected Sign-out" section in `docs/troubleshooting/common-issues.md` was a plain paragraph, while every other section in that file uses bulleted, bold-led items. Reformatted to match (and rebased past your own `bb56e93` move-to-end-of-file commit).
3. **README bullet style** — the added credential-serialization line had no emoji/bold lead-in like its sibling bullets, and read as an internal implementation detail rather than a user-facing feature. Folded the user-relevant part (per-profile credential isolation) into the existing "🔐 Secure sign-in" bullet and dropped the internal "serialized writes" detail from user-facing docs.

### File size/structure
No files in this diff approach the 1000-line threshold; no splitting needed.

## Not verified
This environment has no Flutter/Dart SDK installed, so I could not run `flutter analyze` or `flutter test` myself — review is based on careful manual tracing of the promise-chain logic and the mocked-channel test scenarios. The PR description's own validation note (16 auth/profile tests passing, `flutter analyze --no-fatal-infos` clean) covers this; worth a maintainer double-check after the follow-up commit, particularly that the rebase over `bb56e93` didn't reintroduce the duplicate section.

---

PR #343, opened 2026-09-12
fix(read-state): synchronize read cursors and suppress repeated notifications

krazyjakee:
Read acknowledgements can be lost or arrive out of order, causing channels and notifications to light up again after reconnecting or reading on another device. Add per-connection read cursors and serialized, coalesced acknowledgement retries; acknowledge focused visible panes after history loads and on resume; suppress replayed alerts and dismiss delivered banners through the acknowledged position.

Recover the existing read-synchronization work onto current master, adapting it to the split message/READY event handlers. Preserve the merged own-message acknowledgements and direct-message notification behavior. Global badges now use server-qualified mute settings.

Validation: all 1,410 Flutter tests pass, including own-message, DM call/PiP, inactive-app, replay, partial remote-read, retry, and notification-delivery race coverage. Static analysis and deterministic generated-code checks pass.

Pending acknowledgements and notification associations remain in memory; process termination can lose an unsent acknowledgement. Physical multi-device/native-notification checks and the historical server-side mention-count follow-up are documented in docs/notification-read-state-audit.md.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -172,6 +172,13 @@ When in doubt about Accord behaviour, read `packages/accordkit` (the vendored SD
 - Keep changes minimal and reuse-first; this is a port, not a rewrite.
 - Don't reintroduce Discord endpoints, Discord branding, or Firebase push without explicit instruction.
 
+## Read state
+
+Read-state updates live in `accord_message_events.dart`; READY cursors are
+parsed in `accord_ready_sync.dart`. Use `ReadStateController.acknowledge` for
+local reads so retries, monotonic cursors and notification dismissal stay
+consistent. Details: `docs/notification-read-state-audit.md`.
+
 ## AutoMod
 
 `lib/features/automod/` owns the policy/review UI and block-before-delete flow.
--- a/README.md
+++ b/README.md
@@ -358,3 +358,7 @@ Licensed under the **[GNU General Public License v3.0](LICENSE)** (GPLv3), inher
 Configure server and space rules, review held uploads, and block repeated file
 uploads through the AutoMod controls. See [the client AutoMod guide](docs/automod.md)
 for permissions, video sampling, private evidence, and cooldown behavior.
+
+Read positions synchronize across connected devices, with retries for failed
+acknowledgements and replay suppression for notifications. See the
+[read-state audit](docs/notification-read-state-audit.md) for behavior and limits.
--- a/docs/notification-read-state-audit.md
+++ b/docs/notification-read-state-audit.md
@@ -1,0 +1,54 @@
+# Notification and read-state audit
+
+Originally audited 2026-09-07; reconciled with current master on 2026-09-12. Scope: channel and DM unread highlighting, mention counts,
+desktop/app badges, local notification banners and sounds, gateway replay,
+reconnect hydration, and read acknowledgements between devices.
+
+## Client findings and fixes
+
+| Finding | Resulting change |
+| --- | --- |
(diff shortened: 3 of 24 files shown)


krazyjakee:
## Automated review — code quality, structure, tests

Reviewed the full diff (23 files, ~1,050/-144) against the areas requested: code quality, file size/structure, and test coverage. This is a well-scoped, well-documented change; the read-state/notification bookkeeping in `read_state.dart`, the multi-source ack-position selection in `mark_channel_read.dart`, and the server-qualified mute fix in `global_unread.dart` all check out against their stated invariants and existing regression tests (traced through `receiveMessage`/`markUnread`/`markRead`/`applyRemoteRead`/`acknowledge` call paths and the READY hydrate/replay/partial-remote-read tests).

No 3D model/rendering code is touched, so that check doesn't apply here.

### Findings & fixes (pushed to this branch, commit `73ae215`)

1. **Low — `ReadStateController.clear()` left `_sending` uncleared** (`lib/features/channels/controllers/read_state.dart`). The method was expanded in this PR to reset every position/queue map (`_retries`, `_pending`, `_latest`, `_received`, `_readThrough`, `_syncedThrough`, `_mentionIds`) but missed `_sending`, the set that gates re-entrant `_flush()` calls. Not currently reachable in production (callers use `ref.invalidate(...)` instead of `clear()`, which drops the whole instance), but the doc comment says "Drops every entry" and a future caller relying on that contract could see a channel wedged as "still flushing" after a wipe. Fixed by adding `_sending.clear();`.

2. **Low, test coverage — no regression test for `markChannelRead`'s ack-position priority** (`lib/features/channels/utils/mark_channel_read.dart`). The rewritten logic picks the *newest* of the cached-message list, `fallbackMessageId`, and the tracker's own `latestMessageId` via `compareMessageIds` — this is exactly the fix this PR makes over the old "cached-history-or-fallback-only" logic, per the updated doc comment ("Cached history may be stale... it must never take precedence over a newer known position"). The existing `mark_channel_read_test.dart` only exercised the fallback-only path (no client, no cached messages, no tracker position ahead of fallback), so a regression re-introducing "fallback wins over a newer tracker position" would have passed CI silently. Added a test that seeds the tracker ahead of a stale fallback and asserts the REST `ack` call uses the newer id.

### Not changed (documented limitations / out of scope)
- `docs/notification-read-state-audit.md` already discloses that a server-side `ack_channel` bug (`accordserver`) unconditionally zeroes `mention_count` on any acknowledgement, so a partial remote read can't fully reconstruct a correct local mention count from a stale READY snapshot — this is called out as a server follow-up, not a client bug, and the client already preserves the best information it has (`_mentionIds` bookkeeping in `markRead`).
- The in-memory `_messageNotifications` map in `notification.dart` only shrinks via `dismissReadNotifications`; it's bounded in practice (entries are removed on every local/remote read and on READY reconciliation) but has no hard cap. Not flagging as a fix — just noting it in case very long-lived sessions with disabled read-acknowledgement ever come up.

### Evidence needed / not verified
This sandbox has no Flutter/Dart SDK installed (`flutter`/`dart` binaries absent, no cached SDK), so I could not run `flutter analyze`, `flutter test`, or `scripts/codegen.sh --check` myself to confirm the added test compiles/passes and that `*.g.dart` outputs stay in sync — I relied on manual tracing against the existing test patterns in this file (`message_pane_read_test.dart`'s `_Auth` override) and the `accordkit` `channels.ack` endpoint contract (`POST /channels/:id/ack`, `message_id` body key). Please confirm CI is green on the new commit before merging; if the added test fails to compile I'm happy to fix it up.


---

Issue #358, opened 2026-09-13
Add deterministic event replay and long-session performance benchmarks


krazyjakee:
Add a repeatable workload harness that exercises Daccord's actual message/event controllers and cache lifecycle. Functional tests alone do not establish memory stability or responsiveness during busy, long-running sessions.

Acceptance criteria:
- Replay a fixed synthetic workload including large message bursts (e.g. 100,000 events), edits, deletions, duplicate events, pagination, channel switching, reconnect and account changes.
- Include delayed/out-of-order REST completions and colliding channel IDs across servers.
- Assert final state correctness, bounded retained caches and cleanup after logout; ensure the harness itself does not retain an ever-growing event/sample log.
- Provide a short deterministic CI mode plus a configurable soak mode.
- Report reproducible environment/workload metadata, event processing timings and cache estimates. Measure Flutter frame timing and process memory separately rather than presenting reducer timings as UI performance.
- Establish baselines before introducing timing regression thresholds; avoid fragile absolute timing assertions on shared CI runners.
- Document one-command execution and interpretation of results.

Build on existing integration/test fixtures where useful. Coordinate cache-budget assertions with the bounded timeline work.

Reference: [Serein replay and soak harness](https://github.com/ViceVerse-cz/Serein/tree/baec1df/tools/replay-bench/src).

## Flutter ecosystem candidates

Flutter/Dart candidates, researched 2026-09-13. Prefer tools already used by the repository; separate functional simulation from real-time performance measurements.

| Candidate | Role and limitations |
| --- | --- |
| Existing `flutter_test` + Riverpod `ProviderContainer` overrides | Exercise actual controllers with controlled REST/gateway fixtures. Reuse patterns from `integration/support/harness.dart`, but make the deterministic replay independent of a running server. [Provider testing](https://riverpod.dev/docs/how_to/testing). |
| `fake_async` 1.3.3 | Deterministic timers/microtasks for retry, typing expiry and cache TTL. Already resolved transitively; make it a direct dev dependency if imported. Do not time performance under a fake clock. [Package](https://pub.dev/packages/fake_async). |
| `benchmark_harness` 2.4.0 | Official Dart microbenchmark harness for extracted pure Dart state/reconciliation routines. Current Flutter-dependent controllers cannot simply run in a standalone `dart run` benchmark. [Package](https://pub.dev/packages/benchmark_harness). |
[...]

Issue #357, opened 2026-09-13
Bound message history caches and retain recent channels for immediate previews


krazyjakee:
Daccord disposes unused message controllers, but a watched timeline can grow as older pages and live messages accumulate. Returning to a disposed channel also requires loading its history again.

Add bounded timeline windows and a small cache of recently visited conversations. Show an eligible cached preview immediately, then revalidate against the server. Choose row, estimated-byte and aggregate limits from measurements; Serein's 500-row/4 MiB per timeline and two dormant windows are reference points, not prescribed defaults.

Acceptance criteria:
- Enforce per-window and aggregate memory budgets, including retained reconciliation metadata; distinguish estimated bytes from process memory measurements.
- Eviction preserves the visible reading position and supports fetching history in either direction after trimming.
- Recent-channel previews are keyed by local profile, server/account and channel, with explicit freshness/loading state.
- Invalidate or reconcile previews on edits, deletions, access removal, reconnect/resync, account replacement and logout. Recheck access before reuse.
- Drafts and pending sends survive ordinary cache eviction.
- Tests demonstrate bounded retention during prolonged scrolling/live traffic, correct scroll behavior, revalidation and isolation across servers with colliding IDs.
- Measure channel-switch latency and memory before/after.

Depends on robust stale-request reconciliation; coordinate with the message history correctness issue.

References: [Serein timeline](https://github.com/ViceVerse-cz/Serein/blob/baec1df/crates/session-cache/src/lib.rs), [retention policy](https://github.com/ViceVerse-cz/Serein/blob/baec1df/docs/storage-policy.md).

## Flutter ecosystem candidates

Flutter/Dart candidates, researched 2026-09-13. Candidate suitability still needs measurement on our actual timeline.

| Candidate | Fit and limitations |
| --- | --- |
| Riverpod `ref.keepAlive()` / `KeepAliveLink.close()` | Existing dependency; a small coordinator can retain only the recently visited provider instances and release them on eviction. Riverpod also documents time-based retention, but TTL alone is not an aggregate byte/count budget. [Official lifecycle documentation](https://riverpod.dev/docs/concepts2/auto_dispose). |
| Dart `LinkedHashMap` plus explicit weight accounting | A small owned cache can track recency and estimated retained bytes without adding a generic persistence framework. Recency requires moving an accessed entry to the end; insertion order alone is not LRU. [SDK API](https://api.dart.dev/dart-collection/LinkedHashMap-class.html). |
[...]

PR #363, opened 2026-09-14
Protect message history from stale requests and gateway races

### LINKED ISSUE #356: Protect message history from stale requests and concurrent gateway mutations
Message history requests can complete after a reload, session change, or newer gateway mutation. In `lib/features/messaging/controllers/accord_messages.dart`, the initial load checks client identity, but `loadOlder()` only checks that its provider remains mounted; `reload()` clears the loading boolean without invalidating the older request. This leaves ordering-sensitive paths worth fixing and reproducing with tests.

Introduce explicit request/session generations and reconcile REST snapshots with gateway creates, edits and deletions received while the request was in flight.

Acceptance criteria:
- Delayed results from superseded initial loads, pagination, reloads and replaced clients cannot mutate current state, pagination flags or failure indicators.
- An old request's cleanup cannot clear a newer request's loading state.
- Concurrent gateway creates/edits survive a stale history response; deleted messages cannot be resurrected by it.
- Server/account isolation and pagination deduplication remain intact.
- Deterministic tests cover out-of-order completions, reload during pagination, deletion/edit during fetch, reconnect and account replacement. Audit equivalent thread/forum paths and apply the same protection where needed.

Reference: [Serein core](https://github.com/ViceVerse-cz/Serein/blob/baec1df/crates/client-core/src/lib.rs) and [timeline reconciliation](https://github.com/ViceVerse-cz/Serein/blob/baec1df/crates/session-cache/src/lib.rs). Borrow the approach for Riverpod/Accord rather than porting its Discord transport.

## Flutter ecosystem candidates

Flutter/Dart candidates, researched 2026-09-13. These are source/documentation findings, not a compiled compatibility or runtime validation.

| Candidate | Fit and limitations |
| --- | --- |
| Existing Riverpod 3 lifecycle APIs | Keep the current Notifier architecture. Use `ref.onDispose` for cleanup and explicit monotonically increasing request generations for supersession within a still-live provider. Disposal alone cannot distinguish two requests in the same provider. [Official cancellation guide](https://riverpod.dev/docs/how_to/cancel). |
| Existing `package:http` 1.6.0: `AbortableRequest` | Already resolved in our lockfile. Can cancel an individual request when the client supports it; thread the abort signal through accordkit. Current `accord_rest.dart` uses ordinary requests/convenience GETs. Never close the shared account HTTP client just to cancel one channel fetch. [API](https://pub.dev/documentation/http/1.6.0/http/AbortableRequest-class.html). |
| `package:async` `CancelableOperation` | Optional orchestration helper, not the correctness mechanism. Canceling a wrapper does not undo mutations already made inside the wrapped operation or automatically abort its underlying transport. [API](https://pub.dev/documentation/async/latest/async/CancelableOperation-class.html). |
| Dart `Completer` + `fake_async` | Control completion order explicitly; fake timers for retries/expiry. `fake_async` does not replace real network I/O. [Package](https://pub.dev/packages/fake_async). |

Recommendation: implement generation checks and gateway reconciliation first, then add per-request HTTP cancellation to reduce wasted work. No Dio/state-management migration is warranted for this issue.

Prototype gate: start pagination A, supersede it with reload B, complete B then A, and verify A cannot alter messages or loading/error/pagination flags. Repeat with a deletion arriving between request and response. Feed these cases into #358; #357 relies on this correctness.

krazyjakee:
A delayed history page could overwrite a reconnect reload, erase concurrent gateway edits/creates, resurrect deleted messages, or clear a newer pagination spinner. Channel history, thread replies and forum posts now reject superseded responses and reconcile the accepted snapshot with live mutations received during its fetch.

- Add provider-session and request generations, checked alongside current client identity before publishing results, failure flags, pagination flags or cleanup. Reloads supersede initial loads and pagination; pagination waits for the replacement history fetch.
- Overlay live creates/edits and deletion tombstones on REST snapshots, including edits/deletes for messages that have not reached the cache yet. Preserve display ordering and deduplicate overlapping pages and repeated response IDs. Release the mutation journal on completion, supersession and disposal.
- Apply the same guards to thread/forum refreshes and preserve their visible rows when refresh fails.
- Start reconnect history refreshes and record READY immediately, before awaiting space synchronization. A delayed space request can no longer postpone supersession or make a second READY look like the first.

Validation: `flutter test --no-pub` passed **1,594 tests, 1 skipped**; `flutter analyze --no-pub --no-fatal-infos` passed with 17 existing style notices; `scripts/codegen.sh --check` passed.

The 37 deterministic cases in `test/features/messaging/history_requests_test.dart` use HTTP completers and gateway frames. They cover out-of-order initial/reload successes and failures; pagination A → reload B → pagination C with A completing late; unknown deletes/edits during fetch; page deduplication; replaced clients, accounts and servers with colliding channel IDs; disposal; and READY while space synchronization is still pending. These are reusable regression cases for #358 and establish the history correctness needed by #357.

Fixes #356.


krazyjakee:
## Automated review

Set up a full local toolchain (Flutter 3.47.4 stable) to validate this beyond static reading, since the change is concurrency-sensitive:

- `flutter pub get` — clean
- `flutter analyze --no-pub --no-fatal-infos` — **17 pre-existing info-level notices, 0 new** (matches the PR description exactly)
- `flutter test --no-pub` (full suite) — **1593 passed, 2 skipped, 0 failed**
- `flutter test --no-pub test/features/messaging/history_requests_test.dart` — **all 37 new cases pass**
- `scripts/codegen.sh --check` — generated `*.g.dart` output matches what's committed

I also hand-traced `HistoryRequests.reconcile` against several of the trickier new test cases (stale-page dedup, tombstone-wins-over-late-edit, the READY-before-space-sync race) by hand against the committed logic — the generation/session bookkeeping in `history_request.dart` and its three call sites (`accord_messages.dart`, `forum_posts.dart`, `thread_replies.dart`) behaves as documented. The `accord_event_handler.dart` reordering (marking `hadReady`/dispatching reconnect reloads before `await`ing `loadSpaces`) is correctly sequenced and is exercised by the new "gateway READY supersedes history before delayed space sync" test.

No file in this diff is anywhere near the 1000-line split threshold (largest touched file is 582 lines). No 3D/visual code is touched.

### Findings

1. **Low — minor log regression** (`lib/features/messaging/controllers/thread_replies.dart`): the refactor changed the failed-fetch log description from `'thread $rootId'` to `'replies for $channelId'`, dropping the thread's root id. In a channel with several open threads, a failed reply fetch would only log the channel, not which thread failed. **Fixed** in `1a9d70e`: restored to `'replies for $channelId thread $rootId'`. Re-ran `flutter analyze` and the full `history_requests_test.dart` suite (all 37 pass) after the change.

2. **Informational, not fixed** — `_owns(HistoryRequest, AccordClient)` is duplicated verbatim across all three controllers (`accord_messages.dart`, `forum_posts.dart`, `thread_replies.dart`). Each is only 4 lines and each generated `Ref` type is distinct per Riverpod provider, so a shared extraction would add more indirection/risk than it saves — flagging for awareness only, not proposing a change.

Nothing else stood out as incorrect. Pushed the one fix directly to this branch (`fix/message-history-generations`, commit `1a9d70e`).

---

krazyjakee:
`Analyze & Test (stable)` failed on `1a9d70e`: `scripts/codegen.sh --check` caught that my log-message fix in `thread_replies.dart` shifted the `riverpod_generator` source hash without regenerating `thread_replies.g.dart` — my fault, I ran `flutter analyze`/`flutter test` after that change but skipped the codegen check.

Fixed in `3a0f940`: regenerated and committed the updated hash. Verified locally before pushing: `scripts/codegen.sh --check` now reports "Generated Dart files match the committed output", `flutter analyze --no-pub --no-fatal-infos` still shows only the same 17 pre-existing notices, and `test/features/messaging/history_requests_test.dart` still passes all 37 cases.

---

krazyjakee:
## Review: changes requested (2 correctness issues)

The overall design is sound. Request and session generations are checked together with client identity before every write, including in `finally`. Reloads supersede pagination, and moving `loadSpaces` after `hadReady` fixes the READY race. The 37 deterministic tests are strong. Locally, `flutter test`: **1,594 passed, 1 skipped**, and `flutter analyze` shows only the 17 existing infos.

I reproduced two problems in `HistoryRequests.reconcile` with small tests built on the PR's own harness (`_Harness`):

### 1. An edited or reacted message from an older loaded page moves to the bottom after a reconnect reload
`history_request.dart` `reconcile`: live rows that are missing from the snapshot are appended after it. The code assumes "not in the snapshot" means "created during the fetch". In `AccordMessagesController` it can also mean "an older message the user paginated into, beyond the 50-message reload page".

Repro: seed page `new`, `loadOlder` → page `old`, `reload()`, then `updateMessage(old30)` while the reload is in flight, then complete with page `new`.
Result: `[..., new49, new50, old30]`. `old30` shows below the newest message, and `loadOlder`'s dedupe keeps it there for good. The same happens for pins and reactions on older messages.

Fix: for a messages reload, carry over only rows whose mutation is a *create* seen during this fetch, or rows newer than the snapshot's newest ID. Alternatively, drop pre-existing rows that are missing from the snapshot and let pagination bring them back. Please add this case to the tests.

### 2. A reaction or pin during a reload restores the stale cached copy and loses offline edits
`addMessage`, `_setPinned`, `applyReaction`, `clearReactionEmoji` and `clearReactions` all call `_history.update(<cached object>)`. `reconcile` then puts the **whole stale cached message** in place of the snapshot row. That undoes content edits the reload exists to recover (#356's reconnect case).

Repro: seed `m` with content "before offline edit", `reload()`, `clearReactions('m')` during the fetch, then complete with `m` = "edited while offline".
Result: content is `before offline edit`.

Fix: journal field-level patches, such as reactions or pinned, and apply them to the snapshot row instead of replacing it. Or record only full-message sources (gateway `MESSAGE_UPDATE`/`CREATE` payloads), not in-place mutations of cached objects. `addMessage` recording `existing` has the same problem.

### Minor
- `_load` sets `isLoadingOlder = false` without bumping state. If the superseding reload then fails (state unchanged), the "loading older" spinner can stay visible until the next rebuild.

The other paths look correct: stale and failed responses, client, account and server replacement, disposal, unknown deletes, and the forum/thread refresh keeping its rows on failure.

krazyjakee:
## Review: requesting changes

The approach works. There's one owner per request (`HistoryRequests`), a mutation journal that's cleared on begin, finish, reset and dispose, and ownership checks that cover failure flags and `finally` blocks. It fixes the races described in #356. Moving `hadReady = true` in front of `await loadSpaces` also closes the gap where a second READY arriving during space loading skipped the reload. The new tests are thorough.

**Verified locally:** `flutter test test/features/messaging` passes (400), `test/features/events` passes (53), and `flutter analyze` reports nothing in the changed files. CI is green.

### Blocking: a reaction or pin during a reconnect reload brings back the pre-disconnect copy of the message

`HistoryRequests.update()` stores the **whole cached `AccordMessage`**, and `reconcile()` puts it in place of the snapshot row. `applyReaction`, `clearReactionEmoji`, `clearReactions` and `_setPinned` (messages and forum) all change the *cached* object in place and then journal it. The same happens when `addMessage`/`addPost`/`addReply` receive a duplicate and journal `existing`. So if a reaction or pin event arrives while `reload()` is in flight, the cached copy from before the disconnect replaces the fresh server row. Anything that changed while offline is lost: content edits, embeds, attachments. Recovering those changes is the whole point of `reload()`.

Repro, using the harness in `history_requests_test.dart`:
```dart
final h = _Harness(_Kind.messages);
await h.seed([_message('x', 'before disconnect')]);
final loading = h.reload();
final pending = await h.account.next();
h.messages.applyReaction('x', 'thumbsup', added: true, isOwn: false);
pending.complete([_message('x', 'edited while offline')]);
await loading; await h.flush();
// content == 'before disconnect'  (expected 'edited while offline')
```

Suggested fix: journal the operation rather than the object. For example, store a list of `AccordMessage Function(AccordMessage)` patches per id (replace for edits and creates, a reaction or pin delta for the in-place mutators) and replay them onto the snapshot row. A narrower alternative: only whole-row events (create/edit) replace the row, and reaction/pin deltas are re-applied to the snapshot copy. Please add the repro above as a test.

### Minor
- When `reload()` supersedes an in-flight `loadOlder`, `_load` sets `isLoadingOlder = false` without changing `state`. If the reload then fails, `state` is never reassigned, so the widget that watches the spinner doesn't rebuild and the spinner stays up until something else changes the list. Consider `state = [...?state]` (or equivalent) on that path.
- `HistoryRequests._session` is effectively redundant: `reset()` already sets `_active` to null, so `owns()` fails for old requests anyway. Either rely on it explicitly or drop it.
- `thread_replies.dart` passes `prependLive: false` explicitly, which is the default. The forum log label also changed from `'forum posts for …'` to `'posts for …'`, which makes the log less clear.

krazyjakee:
Pushed 278b39c to fix the blocker from my review. Pins and reactions are now recorded as field changes and replayed onto the fresh server copy, so they no longer bring back the old cached message. Duplicate create echoes are no longer recorded. A live edit can no longer bring back a row that was deleted while offline. I added regression tests (reaction, clear reactions, pin for messages and posts, duplicate echo, deleted row) and regenerated the provider hashes. Locally, `flutter test test/features/messaging test/features/events` passes (463). The minor points from the review are not addressed.

krazyjakee (commented):
Independent re-review of **278b39c8b9a2b9b19d9399574da5974d01d1b788**: **changes still needed for reaction reconciliation**. The content-overwrite and older-message ordering problems are fixed, but the reaction-list trade-off hides real server state. See the inline finding.

1. **Offline content edits survive.** The new field patches preserve the fresh content for reactions, clear-all and pins, and duplicate creates no longer journal the cached row (messages, posts and replies). The ten added regressions pass. Their coverage is narrower than their wording: the reaction-add test starts with no reactions and returns a snapshot with no reactions, so it cannot detect lost offline reaction state; there is no clear-one-emoji regression. Content-only assertions also do not directly cover attachments/embeds, although these field patches no longer replace them.
2. **Older messages are dropped, not moved to the bottom.** I independently tested `seed(new50..new1) → loadOlder(old50..old1) → reload → mutate(old30)` for edit, pin, reaction and duplicate create. After the reload the list is exactly `new1..new50`; `hasMoreOlder` is true, the next pagination cursor is `before=new1`, and pagination restores `old1..old50,new1..new50` without duplication. This is acceptable for the documented newest-page replacement behavior and matches the first review's proposed alternative. It discards the paginated cache window, not server history. The added “deleted while offline” test only proves absence from a snapshot; absence can also mean an older page, so keep a real pagination regression.
3. **Keeping the entire cached reaction list is not an acceptable reconciliation trade-off.** It discards unrelated emojis, counts and `includesMe` recovered from REST. This is not just ambiguity about whether the snapshot already includes the concurrent event. Two independent regressions fail: add thumbs-up during reload loses seven offline hearts; clear thumbs-up during reload also loses those hearts. Clear-all legitimately replaces the list; add/remove/clear-one do not.
4. **Minor spinner issue remains.** A provider observer matching `_OlderHistoryHeader` sees `true` at `loadOlder` start and receives no list notification when a superseding reload fails. The notifier is false but the last rendered value is true, even after the old request completes. Confirmed by a failing observer test; the header only watches the message list, and the load-failed provider is watched only on the empty-history path.
5. **Validation.** On the exact commit, `flutter test --no-pub test/features/messaging test/features/events` passed **463 tests, 1 skipped** (Flutter 3.44.1 / Dart 3.12.1). Seven additional independent tests yielded four passing older-row cases and three expected failures (reaction add, clear-one emoji, spinner notification). No production files were changed. CI is now green on this exact SHA in both the [push run](https://github.com/DaccordProject/daccord/actions/runs/35446616464) and [PR run](https://github.com/DaccordProject/daccord/actions/runs/35446618365): stable/beta analysis/tests, Android native tests and all other enabled gates passed. Builds and opt-in SFU are skipped.

This is an independent agent review, posted as a comment review because the authenticated GitHub account also owns the PR. No merge performed.

lib/features/messaging/controllers/accord_messages.dart
```diff
@@ -514,6 +535,7 @@ class AccordMessagesController extends _$AccordMessagesController {
     }
 
     message.reactions = reactions;
+    _history.patch(messageId, (m) => m.reactions = reactions);
```
krazyjakee: **[P2] Preserve fresh reaction state outside the event's scope**

This closure captures the *entire pre-reconnect reaction list* after changing one emoji, then replaces the fresh REST list with it. For example, cache `x` has no reactions; seven hearts arrive while offline; a thumbs-up arrives during reload; REST returns the edited content plus those hearts. The content now survives, but the hearts disappear and only thumbs-up remains. `clearReactionEmoji` at lines 549–553 has the same problem (and can turn the full list into `[]`). Subsequent deltas cannot recover the missed counts or own-reaction state without another successful refresh.

Failing repro using this file's `_Harness` / `_message` helpers, suitable for `history_requests_test.dart`:

```dart
final h = _Harness(_Kind.messages);
await h.seed([_message('x', 'cached')]);
final reload = h.reload();
final response = await h.account.next();
h.messages.applyReaction('x', 'thumbsup', added: true, isOwn: false);
response.complete([
  _message('x', 'offline edit')..reactions = [
    AccordReaction(emoji: {'name': 'heart'}, count: 7, includesMe: true),
  ],
]);
await reload;
expect(h.state!.single.content, 'offline edit'); // passes
expect(h.state!.single.reactions!.map((r) => r.emoji['name']),
    contains('heart')); // fails: actual ['👍']
```

Please preserve unrelated snapshot reactions and scope clear-one to its emoji. Add/remove on the same emoji also needs an explicit freshness/deduplication strategy or an authoritative follow-up fetch: blindly replaying a numeric delta can double-count an event already present in the snapshot. Add regression cases with offline reaction state in the response, including counts and `includesMe`.

lib/features/messaging/controllers/accord_messages.dart
```diff
@@ -75,25 +80,36 @@ class AccordMessagesController extends _$AccordMessagesController {
...
-    // mutating another during initialization.
-    if (!ref.mounted) return;
-    if (!ref.isCurrentAccordClient(serverKey, client)) return;
-    final list = result.listOrLog<AccordMessage>('messages for $channelId');
-    if (list == null) {
-      _setLoadFailed(true);
-      return;
+    if (!ref.mounted || !ref.isCurrentAccordClient(serverKey, client)) return;
+    final request = _history.begin();
+    isLoadingOlder = false;
```
krazyjakee, line 117: **[P3, non-blocking] Notify the list observer when superseding pagination**

This clears `isLoadingOlder` without notifying the list provider. On a failed reload, `_setLoadFailed(true)` is the only published change; `_OlderHistoryHeader` watches the message list, not that failure provider. The old pagination request is correctly denied cleanup, so nothing removes the rendered spinner. I confirmed this with a provider observer: loadOlder notifies `true`, reload fails, stale pagination completes, notifier is `false` but observer's last value stays `true`. Publish a list/loading-state notification when a reload supersedes pagination (while respecting provider-build restrictions), and add a failure-path regression.

lib/features/messaging/controllers/accord_messages.dart
krazyjakee: **[P2] Preserve server reactions outside the live mutation**

This still journals the entire locally derived reaction list. Repro using the PR's _Harness: seed x with no reactions; start reload; applyReaction('x', '👍', added: true, isOwn: false); complete the pending REST snapshot with x containing content='offline edit' and reactions=[❤️ ×4]. Content is correctly 'offline edit', but the final reactions are only [👍 ×1]: the four server hearts disappear.

clearReactionEmoji has the same problem at line 553: seed [👍 ×1], start reload, clear 👍, return server [👍 ×1, ❤️ ×4]; actual [] instead of [❤️ ×4]. That case has no add/remove double-count ambiguity.

Please restrict clear-one patches to the named emoji and reconcile add/remove events without replacing unrelated server aggregates. For the touched emoji, use an authoritative refresh or an explicit snapshot/event-ordering strategy rather than blindly replaying a count delta (the snapshot may already include it). Add regressions with reactions acquired while offline, including server counts for the touched emoji. The current empty-reaction test cannot catch this loss.

lib/features/messaging/controllers/accord_messages.dart
krazyjakee, line 117: **[P3] Notify list watchers when superseding pagination clears its loading flag**

This reset does not publish state. With 50 cached messages, start loadOlder and let the header render its spinner; start reload, fail that request, then complete the superseded older request. isLoadingOlder is false, but neither request notifies the message-list watcher. _OlderHistoryHeader watches only the list, and the loaded MessagePane does not watch the load-failed flag. A minimal Consumer widget with those same dependencies keeps CircularProgressIndicator visible.

Notify when transitioning away from pagination loading (respecting build-time provider-write constraints), or expose loading state through a watched provider. This remains a minor issue, independently reproduced after 278b39c.

krazyjakee (commented):
Independent re-review of 278b39c8b9a2b9b19d9399574da5974d01d1b788: **changes still needed before approval**, due to reaction-state loss described inline.

1. **Offline content edits: fixed for the covered paths.** The new reaction, clear-all, message/forum pin, and duplicate-create tests pass. Field patches apply to the fresh row, and duplicate creates no longer replace it. The reaction test starts with no cached/server reactions; it therefore proves content preservation, not correct reaction reconciliation. Clear-one-emoji is not covered by the new tests.
2. **Older-page ordering: fixed by dropping the older row.** I independently reproduced seed newest 50 → loadOlder old 50 → reload → edit/pin/react old30 while pending → return newest 50, separately for all three mutations. The result is exactly new1…new50, with no old30 appended. A subsequent loadOlder uses before=new1 and restores old1…old50 before new1…new50, including the fresh server mutation. This is acceptable under the existing documented newest-page replacement semantics: it evicts the old page from the cache, not the server, and pagination restores it. It does not preserve the expanded history/reading position. Please retain these explicit pagination regressions; the new deleted-row test alone does not exercise this case.
3. **Reaction-list trade-off: not acceptable as implemented.** A reaction on one emoji replaces the entire authoritative server list with the pre-disconnect list plus the live change. It can hide unrelated offline reactions indefinitely until another successful refresh. The independent add and clear-one-emoji repros both fail; see inline.
4. **Spinner: still reproducible, minor.** A widget test using the same watch/flag access as _OlderHistoryHeader still renders CircularProgressIndicator after the superseding reload fails and the obsolete older request completes, despite isLoadingOlder=false. See inline.
5. **Validation:** unmodified messaging/events suite: **463 passed, 1 skipped** on Flutter 3.44.1. Six independent focused cases: three older-page cases pass; two reaction-preservation cases and the spinner case fail at their behavioral assertions. CI check-runs attached to this exact SHA: **20 success, 4 skipped, 0 failed/pending** across the push and PR workflows (stable/beta analysis/tests and integration checks passed; build and opt-in real SFU skipped).

Posting a COMMENT review because the available GitHub identity is also the PR author; the verdict above is not an approval. No merge performed.

krazyjakee:
Re: [review 5256612996](https://github.com/DaccordProject/daccord/pull/363#pullrequestreview-5256612996). Both points are addressed in 209c4017 and 08649d1b.

**[P2] Reactions (blocking).** The message no longer keeps the whole local reaction list. Reaction changes that arrive while a history page is being fetched are recorded **per user** and replayed onto the server copy in the order they arrived (`reaction_journal.dart`):
- **Clears:** `clear_emoji` removes only the emoji it names, and `clear` removes all of them. Reactions added after a clear are replayed on top.
- **Own reactions:** keyed on the signed-in user, including a qualified gateway echo. They are applied only if the server's `includesMe` doesn't already reflect them, so there is no double count.
- **Other users' reactions:** a bare count can't show whether the snapshot already includes the change. So that emoji keeps the **server count**, and it's then settled from the list of reacting users (`reactions.listUsers`), fetched after the change arrived. The recorded changes, plus any that arrive while that list is loading, are replayed onto it as set operations keyed on user id. Adds and removes are therefore idempotent whether or not the list already includes them. A failed or truncated list (100 or more users) leaves the server count.
- **Untouched emojis** keep their server counts. Reaction changes on messages that are only in the snapshot are now recorded too.

**[P3] Spinner (minor).** `reload()` now publishes the list when it clears a superseded `loadOlder` flag, so a header that watches only the list rebuilds without the spinner even if the reload then fails.

**New tests** (`history_requests_test.dart`), each run against a snapshot that includes or predates the live change:
- Own add plus its echo, with a server-only emoji surviving.
- Another user's add and removal, with a server-only emoji surviving. Before the list settles the emoji the count stays at the server value (never 3), then settles to the correct count.
- A change that arrives while the reactor list is loading, with and without the list including it.
- A truncated list.
- Clear-one keeps other emojis. Reactions after a clear-all.
- A reaction on a message that is only in the snapshot.
- The failed-reload spinner case from your repro.

I also added the older-page regressions you asked me to keep (live edit, pin and react on `old30` during a reload): the result is exactly `new1…new50`, and the next `loadOlder` (`before=new1`) restores `old1…old50` with the server copy (commit 08649d1b).

Validation: `flutter test test/features/messaging test/features/events` passes (478, 1 skipped). `scripts/codegen.sh --check` is clean after regeneration. `flutter analyze` reports nothing in the changed files. Not merged.

krazyjakee (commented):
Re-review of `209c4017` and `08649d1b`: the original per-user reconciliation and spinner findings are addressed, but I found one new stale-write race in the reactor refresh path that still needs fixing.

**[P2] An older reactor-list request can overwrite a newer reload.** `_refreshReactor` only checks the provider/client session before calling `_setReactors`; `_reactorSession` is not advanced when a newer history generation starts or publishes. Repro with the existing harness:

1. Start reload A, receive another-user `emo_a` add, publish snapshot A with count 1, and hold its reactor list (`['u1']`).
2. Start and finish reload B with a newer server snapshot whose `emo_a` count is 5.
3. Complete A's held reactor list.

The count changes from 5 back to 1. Concurrent HTTP responses can finish out of order, so the older listing must not mutate the row after a newer history snapshot has won. Please associate each refresh with the history/list generation it is allowed to settle (or otherwise invalidate/restart refreshes when a newer snapshot publishes) and retain this sequence as a regression test.

The rest of the requested re-review looks good: untouched server emojis survive; own/federated-self operations are idempotent via `includesMe`; other-user operations and clear→add ordering replay as sets; list failure/truncation is conservative; unicode/custom tokens are constructed correctly; refreshes dedupe per `(message, emoji)`; client replacement, disposal, and provider-session guards are present; and the list notification fixes the failed-reload spinner without introducing a build-time write. The temporary lag at the server count while a reactor list is pending is an acceptable safety trade-off.

Validation on `08649d1b`: `flutter test --no-pub test/features/messaging test/features/events` passed **478 tests, 1 skipped**; `scripts/codegen.sh --check` passed; `flutter analyze --no-pub --no-fatal-infos` reported only the same 17 unrelated info-level notices. The PR-triggered CI run is green. No merge performed.

krazyjakee:
Re: [review 5256806236](https://github.com/DaccordProject/daccord/pull/363#pullrequestreview-5256806236). The P2 race is fixed in 4a984836.

**Fix.** Each reactor-list refresh now records the snapshot generation it was started for. `_snapshotGeneration` goes up each time `_load` publishes a newest-page snapshot. A listing's result is applied only if both of these still hold:
- no newer snapshot has been published since it started, and
- it is still the current refresh for that `(message, emoji)`.

A refresh started for a newer snapshot replaces an in-flight one for the same emoji; it doesn't dedupe into it. Cleanup only removes the refresh's own entry.

Dropping the older result loses nothing. `begin()` supersedes the older history request, so reload A can only publish if it finished before B began. That means every change A's listing was started for arrived before B's snapshot was requested, and B's snapshot includes it. Changes that arrive during B are recorded in B and settled by B's own refresh. `loadOlder` doesn't advance the generation: it only adds rows that weren't in the cache, so it can't make an in-flight listing stale.

**Regression tests** (`history_requests_test.dart`):
- Your repro. Reload A with another user's add has snapshot count 1 and its list is held. Reload B publishes count 5. A's list then completes, and the count stays at **5**. Without the fix this fails with 1, as you reported.
- Same-emoji supersession. B also has another user's change, so it starts its own list for the same emoji (2 list requests). B's list finishes first (6 users), then A's older list (`['u1']`), and the count stays at 6.

The test mock now answers each reactor request with the list as it was when the request was made, and each hold applies to one request only, so out-of-order completion can be exercised.

**Validation:** `flutter test test/features/messaging test/features/events` passes (480, 1 skipped). `scripts/codegen.sh --check` passes. `flutter analyze --no-fatal-infos` reports only the same 17 unrelated infos. Not merged.

krazyjakee (commented):
Re-review of 4a984836: no blocking findings. The stale reactor-list race from my prior review is fixed.

The snapshot generation check prevents an older listing from writing after a newer newest-page snapshot publishes. A newer same-(message, emoji) refresh replaces the older entry, and identity-based cleanup cannot remove that replacement. Operations are journaled only into the refresh currently in the map. I also traced the failure and supersession cases: if a newer reload fails, the older refresh remains eligible and has continued journaling events; if a newer reload publishes without needing a refresh for that emoji, its snapshot is already authoritative; if it does need one, the replacement refresh carries the newer request journal. Clear-one and clear-all orderings do not leave an emoji unsettled. Not advancing the generation for loadOlder is correct because it only prepends deduplicated rows and does not replace existing snapshot rows. Provider rebuild, disposal, and client replacement clear the map and advance the reactor session, while the client/session checks reject late completions.

Validation on the exact head:
- flutter test --no-pub test/features/messaging test/features/events: 480 passed, 1 skipped.
- The new stale-list regression passes. In a disposable worktree with only the production fix reverted to 08649d1b, the same test fails exactly as expected: wanted 5, got 1.
- Regenerating through the build-runner entrypoint leaves no generated Dart diff. The workspace prevents Dart telemetry from touching the read-only home directory, so the scripts/codegen.sh wrapper itself cannot complete locally; both CI Release tooling jobs pass on this SHA.
- Direct Dart analysis reports only the same 17 unrelated info-level notices. Both stable and beta Analyze & Test CI jobs pass.
- Current CI: 20 passed, 4 intentionally skipped, 0 failed or pending across the push and PR runs.

GitHub does not allow the authenticated PR author account to approve its own PR, so this is a COMMENT review with an approval recommendation. No merge performed.

PR #364, opened 2026-09-15
Improve markdown code readability

krazyjakee:
## Summary

Inline code inherited a dark red foreground, and several syntax highlighting colors were difficult to read against code block backgrounds. Use explicit theme text colors for inline and unhighlighted code, and a syntax palette adapted to light and dark themes.

Increase code text from 14 to 14.5 px with medium weight, and set code block line height to 1.5 for easier scanning.

## Validation

- Static analysis passed for both changed files.
- All six existing theme contrast tests passed.
- Checked the new syntax colors and code text colors against all six preset code surfaces: minimum contrast is 4.75:1.
- `git diff --check` passed.


krazyjakee:
## Review

Reviewed against code quality, file size, and test coverage. Ran `flutter analyze` and `flutter test` against a freshly installed Flutter `beta` SDK (matching CI's `FLUTTER_CHANNEL`) to verify rather than just reading.

**Verified correct:**
- Independently recomputed WCAG 2.1 contrast for all 5 new syntax colors × all 6 theme presets against `BonfireThemeExtension.foreground` (the actual code-surface background via `codeblockDecoration`/`codeSpan.backgroundColor`). All pass ≥4.5:1; the worst case is Solarized `gray` (comment color) at 4.75:1, matching the PR description's claim exactly.
- `flutter analyze` on both changed files: no issues.
- Existing `test/theme/theme_contrast_test.dart` (6 tests) still passes unmodified.
- No dead code left behind — the old `PrismStyle()`/`PrismStyle.dark()` construction is fully replaced, no other callers reference it.
- File sizes are well under the 1000-line threshold (123 and 75 lines); no split needed.

**Finding — missing test coverage (fixed, pushed as `d286ed5`):**
This PR reworks `getMarkdownPrismStyle` (a new 30-token syntax palette) and the code text color in `getMarkdownStyleSheet`, both keyed off theme brightness/palette, but added no test. The existing `theme_contrast_test.dart` only guards `dirtyWhite`/`gray` against `background`/`foreground` — it doesn't touch the new Prism tokens or the code-specific text color at all, so a future palette edit (e.g. tweaking one theme's `foreground`, or adding a 7th preset) could silently reintroduce unreadable code with nothing to catch it — the exact class of bug this PR fixes and the reference `theme_contrast_test.dart` exists specifically to prevent (per its own docstring, citing an App Store rejection for "hard to read type").

Added `test/shared/markdown_prism_style_test.dart`, mirroring the existing test's structure: for each `AppThemePreset`, it renders the real theme, pulls the actual `getMarkdownPrismStyle`/`getMarkdownStyleSheet` output via a `BuildContext`, and asserts every syntax + code-text color clears 4.5:1 against `BonfireThemeExtension.foreground`. All 6 presets pass. This exercises the real functions rather than hand-copied hex values, so it stays meaningful if the palette changes later.

**Not fixed — pre-existing, out of scope:** `getMarkdownStyleSheet`'s `paragraph`/`list`/`listItem`/`headline1-6` still use a hardcoded `Color.fromARGB(255, 235, 235, 235)` instead of `palette.dirtyWhite`, unchanged since before this PR (confirmed via `git show` on the base commit). That's a near-white color applied unconditionally — likely unreadable on the `Light` preset's near-white background — but it predates this change, is unrelated to code readability, and touches a much broader surface (all message body/heading text) than this PR's scope. Flagging for a separate fix rather than expanding this PR.

No 3D model/rendering code is touched by this change, so no visual/PNG review was needed.

---

Issue #378, opened 2026-09-19
iOS release blocked: provisioning profile lacks Associated Domains


krazyjakee:
## What happened

The `iOS App Store` job of the v0.2.21 Release run failed during `build_app`
([run 35433624478](https://github.com/DaccordProject/daccord/actions/runs/35433624478)):

```
ios/Runner.xcodeproj: error: Provisioning profile "***" doesn't include the Associated Domains capability. (in target 'Runner' from project 'Runner')
ios/Runner.xcodeproj: error: Provisioning profile "***" doesn't include the com.apple.developer.associated-domains entitlement. (in target 'Runner' from project 'Runner')
** ARCHIVE FAILED **
```

Every other job in that run succeeded: the GitHub Release is published with all
artifacts, and the Google Play and Mac App Store uploads went through. **iOS is
the only platform missing from v0.2.21.**

## Why

cf81b2de (#342) added `com.apple.developer.associated-domains`
(`applinks:www.daccord.gg`) to `ios/Runner/Release.entitlements`, but the
`IOS_PROFILE_BASE64` secret still holds a profile generated before the App ID
had that capability, so Xcode refuses to sign. This is the follow-up #342's own
docs call for:

> The iOS App ID must enable **Associated Domains**. The app claims
> `applinks:www.daccord.gg`; regenerate `IOS_PROFILE_BASE64` after enabling the
> capability. — `docs/app-store-deploy.md`

Nothing in the repo needs changing, and the server side is already correct:
`https://www.daccord.gg/.well-known/apple-app-site-association` returns 200 as
`application/json` with no redirects, associating `8MVM4FMRRC.com.cattrall.daccord`
with `/open/*`.

## Fix

- [ ] Enable **Associated Domains** on App ID `com.cattrall.daccord` in the Apple Developer portal
- [ ] Regenerate the App Store provisioning profile and update `IOS_PROFILE_BASE64`.
      `scripts/bootstrap-signing.sh` does both and verifies the decoded profile with
      `scripts/verify-associated-domains.py`, failing if the capability is still missing.
- [ ] Ship iOS 0.2.21 without re-tagging — the verify/build/publish jobs are gated to
      tag pushes, so a dispatch run only performs the store upload:

      gh workflow run Release --ref v0.2.21 \
        -f deploy_ios=true -f deploy_mac=false -f deploy_android=false

- [ ] Confirm on a signed install that an `https://www.daccord.gg/open/...` link opens the app

## Note

[...]

Issue #250, opened 2026-08-24
Prevent stale dist/release-notes.txt from overriding future release notes


krazyjakee:
## Problem

The committed release-notes override still names v0.2.13 and instructs maintainers to clear it, while release automation blindly prefers any nonempty file. The next release can publish stale notes.

**Audit severity:** Medium

## Evidence

`dist/release-notes.txt`; release-note selection in the release scripts/workflow.

## Proposed remediation / acceptance criteria

Clear the consumed override, validate any override against the current tag/version, and update the release runbook so stale committed content fails early.

---
Found during the repository anti-slop audit on 2026-08-24.

krazyjakee:
Resolved in 5a54a9cb. The consumed v0.2.13 override is cleared, handwritten App Store notes now require an explicit version marker matching both pubspec and the release tag, and stale/missing markers fail before upload. Four behavioral script tests and the full suite pass.

PR #380, opened 2026-09-19
fix(release): reset dist/release-notes.txt to comment-only template

### LINKED ISSUE #377: Release: dist/release-notes.txt still holds the 0.2.20 notes and will fail the next store release
## Summary

`dist/release-notes.txt` on `master` still begins with `# Release version: 0.2.20`, followed by the 0.2.20 release notes.

`dist/app-store-release-notes.sh` says to "keep the file comment-only between releases so derived notes remain the safe default". When the file has notes in it, the script compares the `# Release version:` line with the version in `pubspec.yaml` and the `v*` tag. It stops before uploading if they don't match:

```
::error::dist/release-notes.txt targets 0.2.20 but pubspec.yaml is <next>.
```

`docs/app-store-deploy.md` has the same rule under "Release notes (\"What's New\")".

## Impact

After the version bump past 0.2.20, the store release-notes step will fail until someone clears or rewrites this file.

## Fix

- [ ] Put the file back to comments only (the header comments with no release text), or write the next version's notes with a matching `# Release version:` line.
- [ ] Optionally add a post-release step or CI check that flags a stale override, so this doesn't depend on remembering.

krazyjakee:
Fixes #377

## Problem
`dist/release-notes.txt` still holds hand-written store notes from a past release. When #377 was filed that was the 0.2.20 copy. Since then the 0.2.21 bump (ebd0e14) replaced it with the 0.2.21 copy. `dist/app-store-release-notes.sh` checks the `# Release version:` marker against `pubspec.yaml` and the tag, so the next version bump (0.2.22) would fail before upload:

```
::error::dist/release-notes.txt targets 0.2.21 but pubspec.yaml is 0.2.22.
```

The file has not been reset since the template was added in #250. Every release from 0.2.14 to 0.2.21 overwrote the previous copy instead.

## Change
- **`dist/release-notes.txt`:** restore the comment-only template from #250 (same blob, `95e3eff`), as `docs/app-store-deploy.md` says to do after each release.
- **`test/app_store_release_notes_test.dart`** (8ddf5f3, revised in ceb97a1 after review):
  - Runs the real script against the committed `pubspec.yaml` and `dist/release-notes.txt`, tagged at the pubspec version. The test fails only for a stale or unmarked override. A comment-only file passes, and so do hand-written notes whose marker matches `pubspec.yaml`, so the documented override flow still works. A version-bump PR that forgets to reset the file now fails CI on that PR, not at store upload.
  - Adds an isolated fixture for the comment-only fallback.
- No changes to the script or docs.

## Is 0.2.21 safe to clear from master?
**Yes, as long as iOS 0.2.21 is recovered from the tag as #378 describes. It must not be dispatched from master.**

v0.2.21 has **not** fully shipped yet:
- In [Release run 35433624478](https://github.com/DaccordProject/daccord/actions/runs/35433624478), the GitHub Release, Google Play and Mac App Store jobs succeeded.
- **iOS App Store failed at archive** because the provisioning profile lacks Associated Domains (#378), so it never reached upload or submission.
- No Release run has happened since, so iOS 0.2.21 is still pending.

Clearing the file on master doesn't affect that recovery:
- #378's recovery step is `gh workflow run Release --ref v0.2.21 -f deploy_ios=true -f deploy_mac=false -f deploy_android=false`.
- The iOS job's `actions/checkout` has no `ref:` override, so it checks out the dispatch ref, which is the tag `v0.2.21` (= ebd0e14). The tag still contains the hand-written 0.2.21 notes, whatever master holds.
- I ran the script at `v0.2.21` with `GITHUB_REF_TYPE=tag GITHUB_REF_NAME=v0.2.21`. It printed "Using hand-written notes from dist/release-notes.txt", passed validation, and wrote the 0.2.21 copy.

**Caveat:** a recovery dispatched from `master` would behave differently after this merges. 0.2.20's recovery (run 34709271039) was dispatched from master. From master, pubspec is still 0.2.21 but the file is comment-only, so iOS would get notes derived from commits instead of the 0.2.21 copy. Dispatch with `--ref v0.2.21`. If recovery has to run from master, run it before merging this PR.

## Validation
- Ran `dist/app-store-release-notes.sh` in a worktree with `pubspec.yaml` set to `0.2.22+1`:
  - **Old file:** exits 1 with the mismatch error above.
  - **Reset file:** falls back to notes derived from commits and exits 0.
- `flutter test test/app_store_release_notes_test.dart`: all 6 tests pass. I also varied the committed pair:

| Committed pair | Result |
|---|---|
| reset file, pubspec 0.2.21 | pass |
| 0.2.21 override, pubspec 0.2.21 | pass |
| 0.2.20 override, pubspec 0.2.21 | fail: `targets 0.2.20 but pubspec.yaml is 0.2.21` |
| 0.2.21 override, pubspec 0.2.22 | fail: `targets 0.2.21 but pubspec.yaml is 0.2.22` |
| override without a marker | fail: `no '# Release version: x.y.z' marker` |


krazyjakee:
Reviewed this PR against the code quality, structure, and test coverage criteria.

**Verified the fix itself:** ran `dist/app-store-release-notes.sh` in a scratch git worktree with `pubspec.yaml` pinned to `0.2.22+1`.
- Old (0.2.21) copy of `dist/release-notes.txt` → `::error::dist/release-notes.txt targets 0.2.21 but pubspec.yaml is 0.2.22.`, exit 1 — reproduces the bug from #377.
- Reset (comment-only) copy → `dist/release-notes.txt has no content after comments — deriving from commits`, exits 0. Confirms the PR's own validation notes.

**Findings**

1. *(Low, test coverage — fixed)* `test/app_store_release_notes_test.dart` only ever exercises the script against synthetic overrides written into a temp dir; nothing in the suite reads the actual committed `dist/release-notes.txt`. That's exactly how this regressed for six releases in a row (0.2.14 → 0.2.21) without anyone noticing until a tagged release failed. The script itself only runs at tag time (`release.yml`), so a forgotten reset isn't caught by `ci.yml`, which runs on every PR.

   Pushed a new test, `the committed override stays comment-only between releases`, that reads the real `dist/release-notes.txt` and asserts there's no content left after stripping comment lines — the same check `validate_override`/the "no content after comments" branch effectively perform. This turns the next forgotten reset into a fast, PR-time `flutter test` failure instead of a release-time surprise.

No other issues: the change is a single data file, well under the 1000-line split threshold, the new template wording is accurate to the script's actual behavior (validated above), and there's no 3D/rendering code involved.

Note: I don't have a local Flutter/Dart toolchain in this environment, so I verified the new test's logic against the actual file content with an equivalent script rather than running `flutter test` directly — worth a confirmation run in CI.

---

test/app_store_release_notes_test.dart
```diff
@@ -78,4 +78,29 @@ void main() {
...
+    final contents = File('dist/release-notes.txt').readAsStringSync();
+    final withoutComments = contents
+        .split('\n')
+        .where((line) => !line.trim().startsWith('#'))
+        .join('\n')
+        .trim();
+
+    expect(
+      withoutComments,
+      isEmpty,
```
krazyjakee: [P2] Allow valid hand-written release overrides through CI

This assertion runs unconditionally, including when `release.yml` invokes the full CI gate for a release. Following `docs/app-store-deploy.md` by committing a matching `# Release version: 0.2.22` plus valid user-facing copy therefore fails this test even though `app-store-release-notes.sh` correctly accepts it. Because the build/store jobs require CI success, this disables the documented hand-written override workflow before the release can ship; the test has no way to distinguish "between releases" from a release in progress. Please remove this unconditional repository-content assertion, or validate version consistency while permitting a current override, and cover the comment-only fallback with an isolated script fixture.

krazyjakee (commented):
Reviewed commit 8ddf5f3766df3bc16ac3b097bd81cd3536d2b21c. The template reset itself checks out, but the newly added unconditional test needs revision (inline comment).

Validation:
- `dist/release-notes.txt` is byte-for-byte identical to the template introduced by 5a54a9cb for #250 (same Git blob: 95e3efff58b32ef04699ddd0ae9d6c44f681f73e). All six lines are comments; there is no active version marker or body text.
- Ran the unchanged script in an isolated temporary directory using the PR Git history, pubspec `0.2.22+1`, and tag environment `v0.2.22`: the old override exits 1 with the version mismatch; the reset file logs "no content after comments — deriving from commits", writes identical nonempty iOS/macOS notes, and exits 0.
- `git diff --check` passes. I could not execute the Flutter tests: Flutter fails before test startup because its SDK cache is on a read-only filesystem. The PR now includes a fifth test, so the description's four-test validation predates the latest commit.

The store-shipment prerequisite is not confirmed. The [v0.2.21 GitHub release](https://github.com/DaccordProject/daccord/releases/tag/v0.2.21) exists, and the Mac/Google Play upload jobs succeeded, but the [iOS job](https://github.com/DaccordProject/daccord/actions/runs/35433624478/job/105873116135) failed during archive because the provisioning profile lacks the Associated Domains capability/entitlement. That run did not reach iOS upload or submission. Please provide evidence of a subsequent successful iOS release, or qualify the description and preserve the intended 0.2.21 notes for recovery. Retrying the original tag retains those notes; manually dispatching from master after this reset would derive different copy.

The release script and deployment docs remain unchanged. Resetting the override still requires a human; automating that in the release/bump flow remains a separate follow-up. No repository edits made during this review.

test/app_store_release_notes_test.dart
krazyjakee: Good catch, fixed in ceb97a1. Replaced the unconditional comment-only assertion with a test that runs the actual `app-store-release-notes.sh` against the real committed `pubspec.yaml`/`dist/release-notes.txt` pair, tagged at the current app version — it only fails on a stale or malformed override (the #377 case), and permits a valid current-release override exactly as the script does. Also added an isolated fixture test for the comment-only fallback path, per your second suggestion.

Verified manually (no local Flutter toolchain here) with the actual `dist/app-store-release-notes.sh`:
- Real committed files as-is → exit 0.
- Comment-only fixture in isolation → exit 0, logs "no content after comments — deriving from commits".
- A simulated stale override (old 0.2.21 copy against pubspec 0.2.22) → exit 1 with the mismatch error, confirming the regression guard still catches the original bug.

---

krazyjakee:
Re the store-shipment prerequisite flagged in the review: I don't have access to App Store Connect/TestFlight, so I can't independently confirm whether v0.2.21 has actually reached iOS beyond what's visible in the [failed iOS job](https://github.com/DaccordProject/daccord/actions/runs/35433624478/job/105873116135) (provisioning profile missing the Associated Domains capability). That's a real open question for @krazyjakee to confirm before merging — if iOS 0.2.21 hasn't shipped yet, worth deciding whether to hold this reset until a successful iOS release, or to explicitly preserve the 0.2.21 copy for recovery per the review's suggestion. It doesn't affect the correctness of the code change itself (verified above and by the review), just the timing of merging it.

Test-coverage feedback (the unconditional CI assertion) is fixed in ceb97a1 — replied inline on that thread.

---

test/app_store_release_notes_test.dart
krazyjakee: Checked ceb97a1 independently, so no further change here. The new test runs the real script against the committed `pubspec.yaml` and `dist/release-notes.txt` pair, tagged at the pubspec version. With the committed pair varied locally:

- reset file, or a `# Release version: 0.2.21` override with pubspec 0.2.21: **all 6 tests pass**, so hand-written notes for the current version still get through CI.
- 0.2.20 override with pubspec 0.2.21: fails with `targets 0.2.20 but pubspec.yaml is 0.2.21`.
- 0.2.21 override with pubspec bumped to 0.2.22 (the forgotten reset from #377): fails with `targets 0.2.21 but pubspec.yaml is 0.2.22`.
- copy without a marker: fails with `no '# Release version: x.y.z' marker`.

The comment-only fallback also has its own fixture test. `flutter analyze` on the file is clean.

krazyjakee:
Re the store-shipment point in [the review](https://github.com/DaccordProject/daccord/pull/380#pullrequestreview-5255880904): I checked the workflow runs. **Clearing the 0.2.21 notes on master is safe, provided iOS 0.2.21 is recovered from the tag.**

- **0.2.21 is not fully shipped.** In [run 35433624478](https://github.com/DaccordProject/daccord/actions/runs/35433624478), the GitHub Release, Google Play and Mac App Store jobs succeeded. iOS failed at archive because of the Associated Domains provisioning issue (#378). No Release run has happened since, so iOS 0.2.21 still needs to go out.
- **The recovery in #378 doesn't read master's copy.** It runs `gh workflow run Release --ref v0.2.21 -f deploy_ios=true -f deploy_mac=false -f deploy_android=false`. The iOS job's `actions/checkout` has no `ref:` override, so it builds from the dispatch ref, which is the `v0.2.21` tag (ebd0e14). The tag still contains the hand-written 0.2.21 notes. I ran the script at `v0.2.21` with `GITHUB_REF_TYPE=tag GITHUB_REF_NAME=v0.2.21`. It printed "Using hand-written notes", passed validation, and wrote the 0.2.21 copy.
- **Caveat:** dispatching from `master` would behave differently. 0.2.20 was recovered that way (run 34709271039). After this merges, a dispatch from master would still see pubspec 0.2.21 but a comment-only file, so iOS would get notes derived from commits instead of the 0.2.21 copy. The next iOS release doesn't need master's copy, but the recovery run must use `--ref v0.2.21`. If it has to run from master, run it before merging this PR.

I've updated the PR description with this, the revised test, and the validation table. Point 1 was fixed in ceb97a1 (details in the inline thread). Not merging.

PR #381, opened 2026-09-19
docs: plain HTTP is accepted only on loopback

krazyjakee:
## Problem
Found while reviewing DaccordProject/accordwebsite#18. The self-hosting docs say to connect to a self-hosted server with `http://192.168.1.50:39099` and to port-forward `39099` for friends. The client can't do either.

`packages/accordkit/lib/src/utils/transport_security.dart:6-22,51-71` accepts cleartext `http://`/`ws://` only when the host is loopback (`localhost`, `127.0.0.0/8`, `::1`). Every other host throws `FormatException("Accord server URL must use HTTPS. HTTP is allowed only for loopback development.")`. `AccordServer` calls this in its constructor (`lib/features/server/models/accord_server.dart:23-25,52`). The rejection of `10.0.0.1` / `192.168.1.2` is deliberate and tested (`packages/accordkit/test/transport_security_test.dart:36-56`), so this PR changes the docs and leaves the client alone.

## Changes (docs only)
- **getting-started/adding-a-server.md**: the self-hosted note now says HTTP is loopback-only, gives the exact error for LAN IPs and points to the proxy guidance. It also moves the note out of the middle of the URL-format list.
- **self-hosting/desktop-app.md**: "another device" now means an `https://` address. A new **Reaching the server from other devices** section covers domain + Caddy (with a minimal Caddyfile), Tailscale serve, and `tls internal`/mkcert with a trusted local CA. The internet-invite port table now forwards 80/443 to the proxy, not 39099.
- **self-hosting/deploying-a-server.md**: tells users to connect through the Caddy HTTPS domain and drops the `http://your-ip:39099` advice.
- **troubleshooting/common-issues.md**: the Broken-pipe fix is scoped to localhost. There is a new entry for the loopback-only error.

The companion website PR regenerates `docs.html` from this branch.


krazyjakee:
Reviewed this docs-only PR against `packages/accordkit/lib/src/utils/transport_security.dart` and `lib/features/server/models/accord_server.dart`. Findings, ranked by severity:

**1. Low — loose-list rendering left over from moving the note (fixed, pushed in 81c58a6)**
`docs/getting-started/adding-a-server.md`: removing the blockquote from between the 3rd and 4th bullets left the blank line that used to separate them in place, splitting the 5-item URL-format list into two markdown lists. CommonMark renders that as a "loose" list (extra paragraph spacing on every item) instead of the clean list the PR's own description says it was aiming for ("moves the note out of the middle of the URL-format list"). Removed the stray blank line so all five bullets render as one tight list.

**Everything else checked out:**
- The quoted error string ("Accord server URL must use HTTPS. HTTP is allowed only for loopback development.") is an exact match for `_validateEndpoint`'s thrown `FormatException`.
- The loopback set described (`localhost`, `127.0.0.1`, `::1`) matches `isLoopbackHost`'s doc comment and accepts the full `127.0.0.0/8` range plus IPv6 loopback forms, consistent with the "such as" phrasing used.
- Cross-references to "Reaching the server from other devices" in `desktop-app.md` resolve correctly, and the link text matches the page's frontmatter title and its usage elsewhere in `docs/index.md`.
- No other file in the diff exceeds a few dozen lines (structure/file-size checks don't apply here), no code paths changed (no unit/regression tests needed), and there are no 3D-model/rendering changes in this PR.

No further issues found; this looks correct and safe to merge.

---

docs/self-hosting/desktop-app.md
```diff
@@ -53,24 +53,42 @@ Click the tray icon to open the menu:
...
 ## Inviting people from outside your network
 
-By default the server is reachable on your local network only. To let friends connect over the internet, forward these ports on your router to the computer running Accord:
+To let friends connect over the internet, follow the domain-name option in "Reaching the server from other devices" above. Forward these ports on your router to the computer running Accord:
 
 | Port | Protocol | Purpose |
 |------|----------|---------|
-| 39099 | TCP | Chat (HTTP + WebSocket) |
+| 80, 443 | TCP | HTTPS reverse proxy (chat HTTP + WebSocket to Accord on `39099`) |
 | 7880, 7881 | TCP | LiveKit voice signaling |
```
krazyjakee, line 88: Two things here:

1. **Label**: in LiveKit, `7880` is the HTTP/WebSocket signaling port. `7881` is `rtc.tcp_port`, the ICE-over-TCP media fallback, not signaling. Suggest splitting the row: `7880 | TCP | LiveKit signaling (WebSocket)` and `7881 | TCP | LiveKit media fallback (ICE/TCP)`.
2. **TLS for voice**: now that chat is behind `https://chat.example.com`, signaling on 7880 is still plain `ws://`. The LiveKit-recommended setup is a second TLS hostname on the same Caddy, for example:
   ```
   voice.example.com {
       reverse_proxy localhost:7880
   }
   ```
   Then set the advertised LiveKit URL to `wss://voice.example.com`, forward only 7881/TCP plus the UDP range, and don't forward 7880. If the desktop app can't yet change the external LiveKit URL it advertises, the page should say that voice for remote members isn't supported through this setup yet (and a desktop-app issue should be filed), rather than implying that forwarding 7880 is enough. The same applies to the Tailscale option: `tailscale serve` only publishes 39099, so voice still needs its own path.

krazyjakee (commented):
Checked against the checklist:

1. **Loopback-only claim**: correct. The quoted error matches `_validateEndpoint` in `packages/accordkit/lib/src/utils/transport_security.dart` exactly, and the `localhost` / `127.0.0.1` / `::1` examples match `isLoopbackHost`.
2. **LAN HTTP rejection is intended.** `transport_security_test.dart` rejects `http://192.168.x.x` on purpose, so documenting it is the right fix. It doesn't need a client issue.
3. **Caddyfile / Tailscale**: correct. `chat.example.com { reverse_proxy localhost:39099 }` gets automatic Let's Encrypt and proxies WebSocket with no extra config. `tailscale serve --bg 39099` publishes `https://<machine>.<tailnet>.ts.net` with a trusted cert. `tls internal` / mkcert and the root-trust caveat are also correct.
5. **Links**: the "Reaching the server from other devices" cross-references and the "Running Accord on Your Desktop" title resolve (see the earlier comment; the loose-list fix is in 81c58a6).

**4. Voice: one gap (inline below).** The proxy only covers chat. Voice connects to a separate `livekit_url` returned by the server (`lib/features/voice/controllers/voice.dart` passes `info.livekitUrl` straight to `_session.connect`), and `transport_security` doesn't validate it. So after following this page, a remote member has working HTTPS chat, but voice still goes to whatever URL the desktop app advertises. That is probably `ws://` on a LAN or loopback address: unreachable from outside, and it sends the LiveKit token in cleartext. The web build also blocks `ws://` from an HTTPS page as mixed content. Not blocking for this PR, since the chat-side claims are all correct, but the page now reads as a complete remote-access recipe, so it should at least say so. Suggestions are inline.

docs/self-hosting/desktop-app.md
krazyjakee: Fixed in 30240b5.

1. **Port labels**: split out and corrected. 7880 is signaling (WebSocket), 7881 is the ICE/TCP media fallback, and 7882/UDP (Docker) or 50000–60000/UDP (desktop) carries media.
2. **TLS for voice**: the answer differs between the two setups.
   - **Desktop app: the URL can't be configured.** `desktop/src-tauri/src/sidecar.rs:95-102` starts accordserver with `--livekit-url http://127.0.0.1:7880` (`app_config.rs:114-116`) and never sets `LIVEKIT_EXTERNAL_URL`. accordserver falls back to the internal URL (`src/config.rs:127-129`) and sends it to clients (`routes/voice.rs:127`, `gateway/mod.rs:1377`). `config.toml` has no field for it. So remote members, whether LAN, domain or Tailscale, get a loopback voice URL. The page now says voice works only on the host computer. The port table lists only 80/443, with an explicit "don't forward 7880/7881/UDP" note. Filed DaccordProject/accordserver#84 (configurable wss:// external URL, second Caddy hostname / `tailscale serve --https=8443 7880`, `use_external_ip`).
   - **Server deployment: configurable** via `LIVEKIT_EXTERNAL_URL`. The docs now explain it must be a `wss://` hostname in front of 7880 and give a two-site Caddyfile for non-Docker setups. The Compose stack already does this (`caddy: livekit.example.com`). The ports table now publishes only 443/80, 7881/TCP and 7882/UDP. 39099 and 7880 stay internal, matching the compose files.

Also fixed while I was there: both compose files hardcode `LIVEKIT_EXTERNAL_URL` and the `caddy:` hostnames under `environment:`/`labels:`, so the `.env` values the Quick Start told users to set were ignored. Step 2 now says to edit the domains in the compose file and keep only the API key/secret in `.env`.

krazyjakee:
Voice review addressed in 30240b5 (details in the inline thread). Desktop-app voice is loopback-only and not configurable, so it's now documented as host-only and tracked in DaccordProject/accordserver#84. Server-deployment voice docs now cover `wss://` via `LIVEKIT_EXTERNAL_URL` and have corrected ports. DaccordProject/accordwebsite#19 is regenerated to match. Not merging.

PR #383, opened 2026-09-19
Accept wildcard "*" Associated Domains profile in verify-associated-domains.py

krazyjakee:
## Problem
Found while checking the regenerated iOS profile for #378. Apple writes a wildcard Associated Domains entitlement as a plain **string** `"*"`, but `scripts/verify-associated-domains.py` only accepted a **list**. So it failed on a valid profile. Xcode accepts the profile, so releases aren't affected. But the next run of `scripts/bootstrap-signing.sh` would stop with a false "profile does not authorize applinks:www.daccord.gg" error.

## Fix
- If `com.apple.developer.associated-domains` is a bare string, treat it as a one-entry list. `"*"` (as a string or a list entry) covers every `applinks:` domain. A bare `"applinks:www.daccord.gg"` string is also accepted.
- The script still rejects profiles with no entitlement, an empty list, only other domains, or malformed entitlements.

## Tests
New file `test/packaging/test_verify_associated_domains.py`. CI's existing `unittest discover -s test/packaging` step picks it up. It covers:
- a wildcard string
- a wildcard in a list
- an explicit matching domain
- a missing entitlement
- a non-matching domain
- an empty list, malformed input, and the script's exit codes

`python3 -m unittest discover -s test/packaging -p 'test_*.py'` → 25 tests OK.

## Checked against the real profile
`~/Downloads/Daccord_iOS_App_Store.mobileprovision` has entitlement value `'*'` (a string). Decoded with `openssl smime -verify -noverify` and piped into the script:
- before: exit 1 (FAIL)
- after: exit 0 (PASS)

No GitHub secrets or workflows were touched.


krazyjakee (commented):
Reviewed at `fcc85fc`. LGTM.

**Checklist**
1. **Rejections still hold.** A missing entitlement (`None`), a list with only other domains, an empty list, a non-dict `Entitlements` and non-plist input all return false or exit 1. The tests cover each case, and I also ran extra probes: `b'*'`, `' *'`, `[{}]`, `applinks:*.daccord.gg` and `applinks:www.daccord.gg?mode=developer` are all rejected. Changing the default from `[]` to `None` doesn't change behaviour, because the `isinstance(domains, list)` guard still rejects it.
2. **A bare non-wildcard string is fine.** Apple only writes the bare-string form for `*`, but treating any string as a one-entry list is the conservative reading. It still needs an exact match against `applinks:www.daccord.gg`, so nothing new is let through.
3. **`bootstrap-signing.sh` needs no change.** It pipes `security cms -D` output (an XML plist) to the script over stdin and only checks the exit code. That contract is unchanged, and the error message there is still accurate.
4. **CI runs the tests.** In the "Verify package manifest generation" step of the Release tooling job, the log shows `Ran 25 tests ... OK` (12 existing + 13 new). I got the same result locally.
5. The listed scope exclusions (subdomain wildcards, secrets/workflows/docs) are fine.

Optional nit, not blocking: `REQUIRED_DOMAIN` could replace the hard-coded `applinks:www.daccord.gg` in the `bootstrap-signing.sh` error text. It's a separate file, though, so leaving it as is works too.

PR #384, opened 2026-09-19
fix(composer): Shift+Enter inserts a newline on desktop (#376)

### LINKED ISSUE #376: Composer: verify Shift+Enter inserts a newline on desktop
## Summary

The user docs used to say **Shift+Enter** inserts a new line in the message composer. That claim was removed from `docs/messaging/sending-messages.md` during the documentation cleanup (branch `docs/trim-agent-docs`) because the code doesn't appear to support it. It hasn't been checked in the running app.

## What the code shows

In `lib/features/messaging/views/message_pane/composer.dart`:

- The `TextField` uses `minLines: 1`, `maxLines: 6` and `textInputAction: TextInputAction.send`. Its `onSubmitted` calls `_send()`.
- `_onComposerKey` handles only Ctrl/Cmd+V. It has no Enter or Shift+Enter handling.

So on desktop, Enter and probably Shift+Enter both send. That would leave no keyboard way to write a multi-line message, apart from pasting.

## To do

- [ ] On Linux, Windows and macOS (and on the web), check what Shift+Enter does in the composer.
- [ ] If it sends, add Shift+Enter → newline, the convention in Discord and most chat apps. Keep Enter → send, and don't change mobile, where the keyboard's send action applies.
- [ ] Put the Shift+Enter note back in `docs/messaging/sending-messages.md` and `docs/troubleshooting/keyboard-shortcuts.md`. `test/keyboard_shortcuts_docs_test.dart` guards that page against listing keys that aren't implemented.

krazyjakee:
Closes #376

## Verification: what Shift+Enter did on desktop
The docs claim was never removed: it's been on master since #103 (`docs/messaging/sending-messages.md:12`). **The claim was false on desktop**, though. Shift+Enter **sent** the message.

- The composer `TextField` is multiline (`maxLines: 6`) with `textInputAction: TextInputAction.send` (`lib/features/messaging/views/message_pane/composer.dart`), and before this PR it had no Enter handling. `_onComposerKey` only handled Ctrl/Cmd+V.
- The desktop embedders only insert `\n` on Enter when the action is `newline`, they never check Shift, and they then perform the action (here: send):
  - Linux: `shell/platform/linux/fl_text_input_handler.cc:524-533`
  - Windows: `shell/platform/windows/text_input_plugin.cc:99-100, 483-495`
  - macOS: `shell/platform/darwin/macos/framework/Source/FlutterTextInputPlugin.mm:873-886`
- Key events go to the framework first. The text input plugin only gets unhandled ones (`linux/fl_view.cc:666-673`, `windows/keyboard_manager.cc:367-381`), so the composer can claim Shift+Enter itself.

## Changes
- **Composer**: on native builds, `_onComposerKey` handles Shift+Enter / Shift+numpad Enter (key down and repeat). It inserts a newline at the caret, replacing any selection, and runs `_onChanged` for the mention popup and typing indicator. Plain Enter is still left to the embedder and sends. It skips Enter with Ctrl/Meta/Alt held, and Enter during an IME composition.
- **Docs**: the Shift+Enter sentence now applies only to the desktop apps, and mobile's Send key is described.
- **Tests**: `test/features/messaging/composer_newline_test.dart` has 5 widget tests: a newline at the caret with no send request, a numpad Enter replacing a selection, plain Enter left alone, the IME guard, and Ctrl+Shift+Enter. 2 of them fail without the fix. `flutter test test/features/messaging test/shared` passes (476). `flutter analyze` on the touched dirs shows only the 4 infos that were already there.

## Not changed (known limitation)
- **Web**: Shift+Enter still sends. The web engine performs the input action from its own DOM keydown listener no matter what the framework returns (`lib/web_ui/.../text_editing.dart:1697-1708`), so a framework-only fix would insert a newline *and* send. The docs say desktop only for this reason. A web fix would need a different approach (for example `TextInputAction.newline` on web plus framework-handled plain Enter). That should be a follow-up.
- Not tried on a physical desktop build. The evidence is the embedder source plus widget tests.

## Update after review (dcefe70b)
- **Linux**: the composer no longer claims Shift+Enter. It records the chord at key-down and lets GTK's input method see the key first. If the IM consumes it, the composition commits with no newline. Otherwise the embedder's `send` action becomes a newline. This fixes the case where GTK is composing but Dart has no composing range (Ctrl+Shift+U hex entry, pending IBus keys). Windows and macOS keep the key-handler path.
- The desktop-only wording is now also in `docs/troubleshooting/keyboard-shortcuts.md` and `docs/getting-started/your-first-message.md`.
- 40c26d81: tied the Shift+Enter record to Enter presses (re-review r4054068830). Superseded by e4264a00.
- e4264a00 (third review, r4054340430): **Linux design inverted.** The earlier `send`-action approach restarted the text-input connection (`_finalizeEditing` → `_scheduleRestartConnection`), which dropped queued GTK input. Linux now uses `TextInputAction.newline`: GTK's IM and the embedder insert Shift+Enter newlines natively, and plain Enter (no modifiers, no reported composition) sends from the key handler. The Shift+Enter flag is gone. Windows and macOS are unchanged. Known limitations on Linux: a plain Enter during a composition GTK hasn't reported to Dart sends instead of committing, and Ctrl/Alt+Enter inserts a newline.
- 19 widget tests. The Linux cases use a fake GTK embedder addressed to the original client ID.


lib/features/messaging/views/message_pane/composer.dart
```diff
@@ -104,6 +112,40 @@ class _ComposerState extends ConsumerState<_Composer> {
...
+    }
+    if (!mods.isShiftPressed ||
+        mods.isControlPressed ||
+        mods.isMetaPressed ||
+        mods.isAltPressed) {
+      return false;
+    }
+    // Leave Enter to the IME while it is composing (it commits the candidate).
+    final composing = _controller.value.composing;
+    return !composing.isValid || composing.isCollapsed;
```
krazyjakee: [P2] Preserve GTK compositions that have no Dart composing range

On a native Linux release build (Flutter 3.44.1, GTK simple input context), focus an empty composer, press Ctrl+Shift+U, release the modifiers, type `3042`, then press Shift+Enter. The result is `\nあ`, whereas committing the same input with plain Enter produces just `あ`. I reproduced the extra newline twice. A control build with only Shift+Enter interception bypassed (returning `KeyEventResult.ignored` for that chord) committed just `あ` with no send, confirming the new interception causes the regression. Before the commit, the controller still reports empty text and `composingBase: -1, composingExtent: -1`, even though GTK is composing; this guard consequently consumes the key and inserts a newline before the native input context commits the character.

Please preserve native composition handling when the IME has not exposed a non-empty composing range, and cover this with a native input regression check. The current widget test supplies a composing range directly, so it cannot catch this case.

krazyjakee (commented):
Reviewed b6ae14a8a4f249364890effb03a30d66e59948a8. One correctness issue needs fixing before approval: an active GTK Unicode composition is not always represented by a non-empty TextEditingValue.composing range, so Shift+Enter can insert an unwanted newline before committing the character (inline reproduction below). There is also a small documentation consistency issue.

Validation:
- Flutter 3.44.1 / Dart 3.12.1: `flutter test test/features/messaging test/shared` passed (476 passed, 1 skipped).
- A temporary native Linux release harness using the actual MessagePane/composer and mocked HTTP passed real X11 key-event checks under Xvfb/Mesa: Shift+Enter at the caret inserted a newline with no send; holding Shift+Enter produced repeated newlines with no send; plain Enter made exactly one message POST.
- Six additional temporary widget checks passed: repeat/typing throttling under Linux, Windows, macOS, Android and iOS platform variants, plus mention-popup dismissal and reopening/picking a mention on the next line. These platform variants are not physical mobile-device coverage.
- `flutter analyze` was attempted but its Dart subprocess failed writing telemetry into this environment's read-only home directory. Running `dart --suppress-analytics analyze` on the clean PR checkout completed with 17 existing informational lints, no warnings or errors. CI stable/beta analyze-and-test jobs are green.
- Japanese/Chinese candidate IMEs and physical Android/iOS hardware keyboards remain unverified. The native GTK Unicode composition case was exercised and failed as described inline.

I recommend a separate web follow-up issue, retaining explicit desktop-only documentation until it is fixed. No PR was merged; temporary review source files were removed.

docs/messaging/sending-messages.md
```diff
@@ -9,7 +9,7 @@ section: messaging
 
 ## Sending
 
-Type your message in the composer at the bottom of the message area and press **Enter** to send. Press **Shift+Enter** to insert a new line without sending.
+Type your message in the composer at the bottom of the message area and press **Enter** to send. In the desktop apps (Windows, macOS and Linux), press **Shift+Enter** to insert a new line without sending. On mobile, the keyboard's **Send** key sends the message.
```
krazyjakee, line 12: [P3] Apply the same desktop scope in the shortcut reference

`docs/troubleshooting/keyboard-shortcuts.md:15` still says `Shift+Enter | New line in message` with no platform qualification. Web readers can therefore still follow that instruction and accidentally send their draft. Please qualify that row with the same desktop-only scope as this paragraph; #376 explicitly names both documentation pages.

lib/features/messaging/views/message_pane/composer.dart
krazyjakee: Fixed in dcefe70b. Checking for a composing range can't be made reliable on GTK, so on Linux the composer no longer claims Shift+Enter:
- At key-down it records whether the latest key was Shift+Enter and returns `ignored`. The embedder then gives GTK's IM the first look (`fl_text_input_handler_filter_keypress`, `fl_text_input_handler.cc:500-504`).
- If the IM consumes the key (your Ctrl+Shift+U `3042` case, or a pending IBus key), it commits and nothing else happens, so the result is `あ` with no newline.
- If it doesn't, the embedder performs `send`, and `onSubmitted` inserts the newline instead of sending. The flag is recorded at key-down, so releasing Shift before the action arrives doesn't matter.
- Any other key-down clears the flag, so a Shift+Enter the IM consumed can't turn a later plain Enter into a newline.

Windows and macOS keep the key-handler path.

Tests: new Linux-variant widget tests in `composer_newline_test.dart` cover:
- Shift+Enter with **no** Dart composing range is not handled, and the IM commit gives `あ`.
- A later plain Enter still sends once.
- An unconsumed Shift+Enter (Shift released before the action) becomes a newline.

The first two fail on b6ae14a8. These simulate the IM through the test text input channel. I couldn't run a native GTK/Xvfb check here, so a re-run of your native reproduction would be welcome.

docs/messaging/sending-messages.md
krazyjakee: Fixed in dcefe70b. `docs/troubleshooting/keyboard-shortcuts.md` now reads `Shift+Enter | New line in message (desktop apps: Windows, macOS and Linux)`. I also found the same unqualified claim in `docs/getting-started/your-first-message.md:19` and gave it the same desktop-only scope.

krazyjakee:
Addressed review https://github.com/DaccordProject/daccord/pull/384#pullrequestreview-5256737102 in dcefe70b:

- **P2 (GTK composition):** on Linux, Shift+Enter now always goes to GTK's input method first. The composer turns the embedder's `send` action into a newline only when the IM didn't consume the key, so a composition that has no Dart composing range, or is still pending, commits without an extra newline. Windows and macOS are unchanged. Added Linux-variant regression tests; they fail on the previous commit.
- **P3 (docs):** `keyboard-shortcuts.md` now says Shift+Enter is desktop-only, and so does `getting-started/your-first-message.md`, which had the same claim.
- `flutter test test/features/messaging test/shared`: 484 passed. `flutter analyze` on the touched dirs: only the 4 infos that were already there.
- Not verified here: a native GTK/Xvfb run of the Ctrl+Shift+U reproduction. The tests simulate the IM through the test text input channel.

Agree on a separate web follow-up. Not merged.

lib/features/messaging/views/message_pane/composer.dart
```diff
@@ -93,8 +94,23 @@ class _ComposerState extends ConsumerState<_Composer> {
...
+      return KeyEventResult.ignored;
+    }
     final mods = HardwareKeyboard.instance;
+    final shiftEnter = _isShiftEnterChord(event, mods);
+    if (_inputMethodOwnsEnter) {
+      // Every key press reaches this handler before GTK's input method, so
+      // this records whether the latest one was Shift+Enter. Any other key
+      // clears it, so a Shift+Enter the IM consumed can't turn a later
+      // plain Enter into a newline.
+      _shiftEnterPending = shiftEnter;
```
krazyjakee: [P2] Preserve the chord until its own native action is resolved

This tracks the latest framework key, but GTK's action for an earlier key can arrive after another key-down. If Shift+Enter is followed by `a` before its `send` action returns, `a` sets `_shiftEnterPending` to false here and `_onSubmitted()` calls `_send()` instead of inserting a newline. This can send an unfinished draft when input events are buffered.

Native reproduction on this commit, using the actual composer in a Linux release harness under Xvfb/Mesa: put `hi` in the focused composer, then run `xdotool key --delay 0 shift+Return a`. It posts `{"content":"hi"}` instead of leaving `hi\na` in the editor. I reproduced it four times at delay 0, and also at delays 1 and 2 ms; slower delivery passed.

A deterministic widget reproduction is: dispatch Shift down, Enter down/up, Shift up, then `keyA` down/up, then `testTextInput.receiveAction(TextInputAction.send)`. Expecting no message POST fails: one POST occurs. Please associate submission handling with the originating Enter rather than a mutable latest-key boolean, while retaining the IM-consumed-key protection, and add coverage for this buffered ordering.

krazyjakee (commented):
Re-reviewed dcefe70b92c119b1b8dfca9a3e157d85e90cc0cc. The original GTK composition and documentation findings are addressed, but I cannot approve yet: buffered key events can overwrite the pending Shift+Enter chord before its native send action returns, causing an unintended message send (P2 inline).

Validation:
- Native Linux release harness using the actual MessagePane/composer, mocked HTTP, GTK simple input context, Xvfb and Mesa: Ctrl+Shift+U, release modifiers, type 3042, Shift+Enter commits just `あ`, with no newline or message POST. A later plain Enter posts `あ` exactly once.
- Holding Shift+Enter produces repeated newlines with no message POST. After an IM commit, both a mouse click within the composer and clicking another text field then returning to the composer leave plain Enter sending exactly once.
- `flutter test test/features/messaging test/shared`: 484 passed, 1 skipped, including the 13 PR platform-variant cases. Four additional temporary Linux widget tests passed: click after commit, focus away/back after commit, key repeats, and numpad selection replacement with Shift released before the action. The PR's delayed-action test also covers releasing Shift before submission.
- One additional temporary widget test fails for the buffered-key ordering described inline; the same problem reproduces natively.
- Windows/macOS key handling is behaviorally unchanged by this commit and the corresponding widget variants pass. Physical Windows/macOS and Japanese/Chinese candidate IMEs were not exercised.
- All three docs pages consistently scope Shift+Enter to Windows/macOS/Linux desktop apps. `dart --suppress-analytics analyze lib/features/messaging/views/message_pane test/features/messaging/composer_newline_test.dart` reports no issues. Current CI checks pass (optional jobs skipped).

No PR was merged. Temporary review source files were removed.

lib/features/messaging/views/message_pane/composer.dart
krazyjakee: Fixed in 40c26d81. The mutable latest-key boolean is gone. The record (`_lastEnterWasShiftEnter`) is now set **only by Enter / numpad Enter key-downs**; other keys never touch it.

**Why this pairs each action with its Enter:** the Linux embedder performs `send` only in the `GDK_KEY_Return`/`KP_Enter`/`ISO_Enter` branch of `fl_text_input_handler_filter_keypress`, and only when GTK's IM didn't consume the key. So every action comes from an Enter the composer has already seen, and it is attributed to the most recent one.
- **Buffered keys (your repro):** `shift+Return a` leaves the record at Shift+Enter, so the action becomes a newline and nothing is sent.
- **IM-consumed keys:** a consumed Enter produces no action, and the next Enter press overwrites its record. This works both ways: a consumed Shift+Enter followed by plain Enter sends, and a consumed plain Enter followed by Shift+Enter inserts a newline.
- **No reset on submit:** two quick Shift+Enters whose actions arrive back to back give two newlines.

**Known limitation** (in the doc comment): two *unconsumed* Enters with different Shift state within one embedder round trip (a few ms) both use the later one's state. Without a per-key id from the embedder, I don't see a way to tell a consumed Enter from one that is still in flight. A FIFO queue gets thrown off by IM-consumed Enters, and that can turn a Shift+Enter into a send, which is worse.

**Tests** (Linux variant, `composer_newline_test.dart`):
- Your repro: Shift down, Enter down/up, Shift up, `keyA` down/up, then `receiveAction(send)`. Result: no POST and `hi\n`; after the embedder's `a` update it is `hi\na`. This fails on dcefe70b.
- A consumed plain Enter followed by Shift+Enter gives a newline and no send.
- Two back-to-back Shift+Enter actions give two newlines.

Suite: `flutter test test/features/messaging test/shared` 487 passed. Analyze shows only the 4 infos that were already there. I couldn't run your native Xvfb repro here, so a re-run would be appreciated. Not merged.

lib/features/messaging/views/message_pane/composer.dart
```diff
@@ -104,6 +120,93 @@ class _ComposerState extends ConsumerState<_Composer> {
...
+  /// for both actions.
+  bool _lastEnterWasShiftEnter = false;
+
+  void _onSubmitted() {
+    // EditableText unfocuses on a `send` action just before onSubmitted runs;
+    // the field is enabled, so asking for focus straight back lands in the
+    // same frame.
+    _focusNode.requestFocus();
+    if (_inputMethodOwnsEnter && _lastEnterWasShiftEnter) {
+      _insertNewline();
```
krazyjakee: [P2] Preserve buffered input when converting a native send action to a newline

The original reproduction no longer sends, but still loses user input: with `hi` focused, `xdotool key --delay 0 shift+Return a` leaves `hi\n`, dropping `a` (four native release runs; also reproduced at 1 and 2 ms). `shift+Return a b c` at delay 0 drops all three letters. Two rapid `shift+Return` chords similarly produce only one newline, despite the new same-Shift test.

`TextInputAction.send` runs `EditableText._finalizeEditing(... shouldUnfocus: true)`. Refocusing here causes `_scheduleRestartConnection()` after the callback, replacing the text-input client. GTK updates/actions already queued for the old client ID are then discarded by `TextInput._handleTextInputInvocation`. The new tests miss this because `testTextInput.updateEditingValue` / `receiveAction` address the current client, including after it has been replaced. A temporary test that captures the original `TextInput.setClient` ID, receives the Shift+Enter send action, then delivers the buffered editing update with that original ID deterministically loses the character.

Please keep newline handling from invalidating queued native input, and cover the real client-ID ordering (including two same-Shift actions). Merely changing the boolean or adding an `onEditingComplete` callback will not by itself bypass the unconditional restart for a send action. The native expectation should be `hi\na`, with no POST, and two same-Shift chords should retain both newlines.

krazyjakee (commented):
Re-reviewed 40c26d81fa1f3894e8d89a707b9c0a9a3c4d24c6. The buffered-key boolean fix prevents the unintended send, and the GTK composition / stale-Shift checks pass. I cannot approve yet: the native buffered reproduction now silently drops the following character, and two buffered Shift+Enters lose the second newline (P2 inline).

Validation on Flutter 3.44.1 / Dart 3.12.1:
- Native Linux release harness using the actual MessagePane/composer, mocked HTTP, GTK simple input context, Xvfb and Mesa: `xdotool key --delay 0 shift+Return a` leaves `hi\n` with no message POST, but loses `a`. Reproduced four times at 0 ms and also at 1 and 2 ms; at 5 and 12 ms it leaves `hi\na`.
- Ctrl+Shift+U, release modifiers, type 3042, Shift+Enter commits exactly `あ`, with no newline or POST. Subsequent plain Enter sends `あ` once. Clicking inside the composer, or focusing another text field and returning, still lets plain Enter send exactly once. Those click/focus checks also pass after an ordinary Shift+Enter newline: no stale Shift state leaks.
- Two Shift+Enters at 0 and 2 ms produce only one newline; at 12 ms they produce two. Shift+Enter followed by `a b c` at 0 ms loses all three letters. This exceeds the documented different-Shift limitation.
- The narrower mixed-Shift attribution limitation is acceptable to me as an explicitly documented follow-up: an unconsumed plain Enter is present in those sequences, so they do not recreate the original send-without-a-send-chord bug. I reproduced both orders. This does not excuse the separate buffered edit/action loss above.
- `flutter test test/features/messaging test/shared`: 487 passed, 1 skipped, including all 16 composer platform-variant cases. One additional temporary widget regression using the original native text-input client ID fails because the queued character is discarded.
- `dart --suppress-analytics analyze lib/features/messaging/views/message_pane test/features/messaging/composer_newline_test.dart`: no issues. Current CI checks pass; optional jobs are skipped.

Physical Windows/macOS and Japanese/Chinese candidate IMEs were not exercised. Temporary review source files were removed. No PR was merged.

lib/features/messaging/views/message_pane/composer.dart
krazyjakee: Fixed in e4264a00. Thanks for pinning down the mechanism: I confirmed `_finalizeEditing` always calls `_scheduleRestartConnection()` for `send` (`editable_text.dart:3910-3911`), so any design built on the send action races. I took the inversion you suggested, **on Linux only**:

- **The field uses `TextInputAction.newline`.** Shift+Enter is never handled in Dart. GTK's IM sees it first, so a composition commits even when Dart has no composing range. Otherwise the embedder inserts `\n` into its own model, in order with any keys that follow. A `newline` action on a multiline field is a no-op in Dart: no unfocus, no `onSubmitted`, no connection restart. The Shift+Enter flag is gone.
- **Plain Enter** (no modifiers, no *reported* composition) sends from the key handler. `_send` clears the field synchronously, so `setEditingState('')` reaches the embedder before it processes any key typed after Enter. Held-Enter repeats are swallowed.
- Windows and macOS are unchanged.

**Tests.** The Linux cases now use a fake GTK embedder that keeps its own text model, follows the field's `inputAction`, and addresses every message to the client ID registered **before** the key press. A restart would therefore show up as a dropped update, and each test also checks that no `TextInput.setClient` happened. Cases:
- Your repro (`hi`, Shift+Enter, `a` buffered): gives `hi\na` with no POST.
- Rapid double Shift+Enter: gives `hi\n\n` with no POST.
- Hidden composition (no Dart composing range) plus Shift+Enter: the IM commits `あ` and no newline is added.
- Plain Enter: sends once and keeps an `a` typed right after it.
- Held Enter: sends once and adds no newlines.
- Plain Enter during a reported composition: goes to the IM.

The first two fail on 40c26d81 (`hi\n`). A temporary test modelled on yours (send action on the original client ID) reproduced the dropped `a` on 40c26d81 before the change. Windows/macOS tests are unchanged. `flutter test test/features/messaging test/shared`: 490 passed. Analyze shows only the 4 infos that were already there.

**Known limitations** (documented on `_nativeNewlines`). Please check them in your native harness:
1. A **plain** Enter during a composition that GTK hasn't reported to Dart (e.g. Ctrl+Shift+U `3042` then plain Enter) now sends the draft instead of committing. Shift+Enter in that state is still safe.
2. Ctrl/Alt+Enter inserts a newline on Linux (it sends on Windows/macOS).
3. Not verified natively: a key typed within one embedder round trip *before* a plain Enter (`xdotool key --delay 0 x Return`). The framework could handle the Enter before that key's GTK update arrives. The same ordering applies to the Windows/macOS Shift+Enter key-handler path.

I couldn't run Xvfb here. Not merged.

PR #385, opened 2026-09-19
Take markdown body text colour from the theme (Light theme readability)

krazyjakee:
Found while reviewing #364.

## Problem
`getMarkdownStyleSheet` hard-coded the paragraph, list and headline colour to `#EBEBEB`. That's fine on dark presets, but on **Light** it measures **1.01:1** against the `background` surface, so markdown message text was effectively invisible. Confirmed with the new test run against `main` (only Light fails; Solarized has no light variant and passes).

## Fix
Use `BonfireThemeExtension.dirtyWhite` (the high-emphasis token every other text style in the app uses). On dark presets this is a small shift from `#EBEBEB` to the palette's body-text colour (e.g. `#D5D9DF` on Dark).

## Test
`test/shared/markdown_text_contrast_test.dart` checks that every coloured text slot (paragraph, list, listItem, headline1–6) reaches 4.5:1 on both `background` and `foreground` for each preset. It uses the same approach as `theme_contrast_test.dart` and #364's code contrast test. The full suite passes (1563) and `flutter analyze` is clean.

**Note:** this changes the same file as #364 (the top of `getMarkdownStyleSheet`), so whichever PR merges second will need a small rebase. The resolution is to keep `final palette = …` and set `textColor = palette.dirtyWhite`.


krazyjakee (commented):
Reviewed the implementation and tests; no blocking findings. I verified the actual markdown widget in Light and Dark with paragraphs, lists, headings, links, inline/fenced code, quotes, and an @everyone mention. The Dark shift to #D5D9DF remains comfortably readable (10.29:1 on the message background). The new test is a meaningful regression guard: restoring #EBEBEB reproduces the Light failure at 1.011:1, while the PR passes all six presets. Local full suite: 1563 passed; changed-file analysis and diff checks are clean; CI is green. The #364 merge-order note is correct: retain #364's palette/codeStyle setup and use final textColor = palette.dirtyWhite. GitHub does not allow the PR author account to approve its own PR, so I am recording the successful review as a comment instead.

